PR #3, opened 2016-08-09
Remove unnecessary explicit dependency on glog.

wohe:
It is not needed (anymore?) and leads to linking errors
(does not find libglog) when glog_catkin is used and
libgoogle-glog-dev is not installed.


damonkohler:
LGTM

PR #8, opened 2016-08-24
Adds rate timer.

damonkohler:


cartographer/common/time.h
```diff
@@ -40,6 +46,32 @@ struct UniversalTimeScaleClock {
 using Duration = UniversalTimeScaleClock::duration;
 using Time = UniversalTimeScaleClock::time_point;
 
+// Visible for testing.
+class Clock {
+ public:
+  virtual ~Clock(){};
```
wohe: virtual ~Clock() {}
i.e. add space and drop ';'

cartographer/common/time.h
```diff
@@ -32,6 +32,12 @@ struct UniversalTimeScaleClock {
   using duration = std::chrono::duration<rep, period>;
   using time_point = std::chrono::time_point<UniversalTimeScaleClock>;
   static constexpr bool is_steady = true;
+
+  static time_point now() noexcept {
+    using namespace std::chrono;
+    return time_point(
+        duration_cast<duration>(system_clock::now().time_since_epoch()));
```
wohe: This needs a correction for the different epochs. We should probably pull the kUtsEpochOffsetFromUnixEpochInSeconds constant from cartographer_ros's time_conversion.h over here.

damonkohler:
PTAL

cartographer/common/rate_timer.h
```diff
@@ -0,0 +1,134 @@
...
+
+    std::vector<double> diff(deltas.size());
+    std::transform(deltas.begin(), deltas.end(), diff.begin(),
+                   [mean](const double x) { return x - mean; });
+    const double squared_sum =
+        std::inner_product(diff.begin(), diff.end(), diff.begin(), 0.);
+    const double sigma = std::sqrt(squared_sum / deltas.size());
+
+    std::ostringstream out;
+    out << std::setprecision(2) << mean << "s +/- " << sigma << "s";
```
wohe: Add a space in front of the unit "s" in both places? Should we enforce scientific notation?

cartographer/common/rate_timer.h
```diff
@@ -0,0 +1,134 @@
...
+    result.reserve(count);
+    for (size_t i = 0; i != count; ++i) {
+      result.push_back(
+          common::ToSeconds(events_[i + 1].time - events_[i].time));
+    }
+    return result;
+  }
+
+  // Returns the average and standard deviation of the deltas.
+  string DeltasDebugString() const {
```
wohe: Is this used externally? Consider making it private.

cartographer/common/rate_timer.h
```diff
@@ -0,0 +1,134 @@
...
+    events_.push_back(Event{time, ClockType::now()});
+    while (events_.size() > 2 &&
+           (events_.back().wall_time - events_.front().wall_time) >
+               window_duration_) {
+      events_.pop_front();
+    }
+  }
+
+  // Computes all differences in seconds between consecutive pulses.
+  std::vector<double> ComputeDeltasInSeconds() const {
```
wohe: private?

cartographer/common/rate_timer.h
```diff
@@ -0,0 +1,134 @@
...
+    if (events_.size() < 2) {
```
wohe: Should we move this check to DebugString() to avoid printing NaNs and make it a CHECK here?

cartographer/common/rate_timer.h
```diff
@@ -0,0 +1,134 @@
...
+    return result;
+  }
+
+  // Returns the average and standard deviation of the deltas.
+  string DeltasDebugString() const {
+    const auto deltas = ComputeDeltasInSeconds();
+    const double sum = std::accumulate(deltas.begin(), deltas.end(), 0.);
+    const double mean = sum / deltas.size();
+
+    std::vector<double> diff(deltas.size());
```
wohe: This could be rewritten similar to:
double squared_sum = 0.;
for (const double x : deltas) {
  squared_sum += common::Pow2(x - mean);
}

This would not only be more concise and readable, but also save memory.

cartographer/common/rate_timer.h
```diff
@@ -0,0 +1,134 @@
...
+    std::transform(deltas.begin(), deltas.end(), diff.begin(),
+                   [mean](const double x) { return x - mean; });
+    const double squared_sum =
+        std::inner_product(diff.begin(), diff.end(), diff.begin(), 0.);
+    const double sigma = std::sqrt(squared_sum / deltas.size());
```
wohe: Should we use std::sqrt(squared_sum / (deltas.size() - 1))? This would be the corrected sample standard deviation like we had before.

cartographer/common/rate_timer.h
damonkohler: Done.

wohe:
LGTM

PR #14, opened 2016-08-26
Sets up Travis CI.

damonkohler:


docs/source/index.rst
```diff
@@ -45,33 +45,17 @@ Getting Started without ROS
 
 On Ubuntu 14.04 (Trusty):
```
wohe: Should we add "and 16.04 (Xenial)? We have tested that it works before.

scripts/install_debs.sh
```diff
@@ -0,0 +1,37 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -o errexit
+set -o verbose
+
+# Install the required libraries that are available as debs.
+apt-get update
```
wohe: sudo apt-get update?

wohe:
LGTM

docs/source/index.rst
damonkohler: I think we should wait to do so until we're testing Xenial in CI.

PR #22, opened 2016-09-08
Optimize only once per submap in 2D.

wohe:
The setting was unintentionally set to optimize twice per submap which
harms parallelism.


SirVer:
lgtm.

PR #25, opened 2016-09-21
Removes max_covariance_trace option.

wohe:
This option is no longer really useful to predict outliers.
It is removed and other parameters scaled to minimize the change
in behavior. And related cleanup.

### DIFF
--- a/cartographer/common/math.h
+++ b/cartographer/common/math.h
@@ -82,21 +82,20 @@ T atan2(const Eigen::Matrix<T, 2, 1>& vector) {
   return ceres::atan2(vector.y(), vector.x());
 }
 
-// Computes (A/'scale')^{-1/2} for A being symmetric, positive-semidefinite.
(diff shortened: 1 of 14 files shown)


cartographer/common/math.h
```diff
@@ -82,21 +82,20 @@ T atan2(const Eigen::Matrix<T, 2, 1>& vector) {
   return ceres::atan2(vector.y(), vector.x());
 }
 
-// Computes (A/'scale')^{-1/2} for A being symmetric, positive-semidefinite.
+// Computes 'A'^{-1/2} for A being symmetric, positive-semidefinite.
```
damonkohler: Now that ComputeSqrtLambda is gone, could you add some documentation here referring to the SPA paper?

wohe: As discussed, the constraint struct member is called 'sqrt_Lambda_ij' which should serve as this reference.

PR #40, opened 2016-10-07
commit .gitignore with build and .idea for convenience

catskul:


damonkohler:
Thanks for contributing, @catskul. However, this seems like something that belongs in your own personal .gitignore file as it's specific to your development setup.

catskul:
I've seen both strategies, private and shared. It's certainly up to you guys of course.

The official build instructions suggest a build dir in the cartographer dir so perhaps a `.gitignore` with just that might make sense?

Feel free to close otherwise.

catskul:
CLA now signed for Carnegie Robotics

damonkohler:
Good point about ignoring the build directory. We do most of our development by building the library in conjunction with ROS support which doesn't create a build directory in the source tree.

If you would please remove the .idea line, I'll merge this.

catskul:
Done.

PR #41, opened 2016-10-11
Improves logging, fixes #94.

damonkohler:


cartographer/mapping_3d/motion_filter.cc
```diff
@@ -39,8 +39,9 @@ MotionFilter::MotionFilter(const proto::MotionFilterOptions& options)
 
 bool MotionFilter::IsSimilar(const common::Time time,
                              const transform::Rigid3d& pose) {
-  LOG_EVERY_N(INFO, 500) << "Motion filter reduced the number of scans to "
-                         << 100. * num_different_ / num_total_ << "%.";
+  LOG_IF_EVERY_N(INFO, num_total_ > 0, 500)
```
wohe: This logs after the first call which is still not very useful information. `LOG_IF_EVERY_N(INFO, num_total_ >= 500, 500)`? Then we would skip the NaN we log now, but leaves the logging unchanged otherwise.

cartographer/common/ordered_multi_queue.h
```diff
@@ -66,8 +66,8 @@ class OrderedMultiQueue {
     auto* queue = FindOrNull(queue_key);
     if (queue == nullptr) {
       // TODO(damonkohler): This will not work for every value of "queue_key".
-      LOG_EVERY_N(WARNING, 60) << "Ignored value for queue: '" << queue_key
-                               << "'";
+      LOG_FIRST_N(WARNING, 1) << "Ignoring values for queue: '" << queue_key
```
wohe: `LOG_EVERY_N(WARNING, 1000)` for now?

cartographer/mapping_3d/motion_filter.cc
damonkohler: Done.

cartographer/common/ordered_multi_queue.h
damonkohler: Done.

PR #44, opened 2016-10-11
Improves documentation.

damonkohler:
* Adds system requirements.
* References our mailing list.
* Nudges folks toward using ROS.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -30,13 +30,14 @@ configurations.
 .. _Cartographer: https://github.com/googlecartographer/cartographer
 .. _SLAM: https://en.wikipedia.org/wiki/Simultaneous_localization_and_mapping
 
-Documentation
-=============
+Getting started
+===============
 
-You will find complete documentation for using Cartographer at `our Read the
-Docs site`_.
+* Learn to use Cartographer at `our Read the Docs site`_.
+* Please join the `mailing list`_ and ask questions.
 
 .. _our Read the Docs site: https://google-cartographer.readthedocs.io
+.. _mailing list: https://groups.google.com/forum/#!forum/google-cartographer
 
 Contributing
 ============
@@ -50,12 +51,10 @@ page`_.
     :alt: Build Status
     :scale: 100%
     :target: https://travis-ci.org/googlecartographer/cartographer
-
 .. |docs| image:: https://readthedocs.org/projects/google-cartographer/badge/?version=latest
     :alt: Documentation Status
     :scale: 100%
     :target: https://google-cartographer.readthedocs.io/en/latest/?badge=latest
-
 .. |video| image:: https://j.gifs.com/wp3BJM.gif
     :alt: Cartographer 3D SLAM Demo
     :scale: 100%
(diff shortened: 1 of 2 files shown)


README.rst
```diff
@@ -38,6 +38,13 @@ Docs site`_.
 
 .. _our Read the Docs site: https://google-cartographer.readthedocs.io
 
+Community
+=========
+
+Please join the `mailing list`_ and ask questions.
```
SirVer: Please direct any questions to our mailinglist?

docs/source/index.rst
```diff
@@ -27,8 +27,15 @@ configurations.
 .. _Cartographer: https://github.com/googlecartographer/cartographer
 .. _SLAM: https://en.wikipedia.org/wiki/Simultaneous_localization_and_mapping
 
+Getting started
+===============
+
+The easiest way to get started with Cartographer is to use `ROS
```
SirVer: I would switch this around: Cartographer is a standalone, framework agnostice c++ library. To see it in action quickly ...

docs/source/index.rst
```diff
@@ -38,7 +45,10 @@ complete documentation for using Cartographer with ROS at the
 .. _Cartographer ROS Read the Docs site: https://google-cartographer-ros.readthedocs.io
 
 Getting started without ROS
-===========================
+---------------------------
+
+Please see our ROS integration as a starting point for integrating your system
+with the standalone library. Currently, it is the best available reference.
```
SirVer: mention that it is kept up to date with development through CI and is therefore always correct.

docs/source/index.rst
```diff
@@ -57,6 +67,17 @@ On Ubuntu 14.04 (Trusty):
   :linenos:
   :lines: 20-
 
+System Requirements
+===================
+
+Although Cartographer may run on other systems, it is confirmed to be working
```
SirVer: do we plan to ever support 32 bit? I do not think so and I would therefore mention this here.

README.rst
damonkohler: This is the same call to action that Ceres uses. I prefer it since it encourages folks to join even if they don't currently have any questions.

docs/source/index.rst
damonkohler: Done.

docs/source/index.rst
damonkohler: Would rather not speculate. It's possible that contributions will come in to support arbitrary architectures. Nothing fundamentally prevents us from working on 32-bit.

catskul: Perhaps we should make note of that as a known limitation. 

Proposed wording:
https://github.com/catskul/cartographer/commit/9561cc6ad8f1f883ce6fbbb8c573d12609ac3957

If we wanted to be thorough, we could add in a check for 32bit in CMakeLists.txt with an error message to the same effect.

damonkohler: Briefly looked into adding a CMake check. Doesn't look like it's straight forward or reliable. #46 calls it out as a known issue.

PR #47, opened 2016-10-12
Prepare CMake files for testing support.

SirVer:
- Only build gmock if we really require testing.
- Adds a wrapper for generating catkin aware tests. This works around
  ros/catkin:830.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -22,7 +22,8 @@ set(CARTOGRAPHER_VERSION ${CARTOGRAPHER_MAJOR_VERSION}.${CARTOGRAPHER_MINOR_VERS
 set(CARTOGRAPHER_SOVERSION ${CARTOGRAPHER_MAJOR_VERSION}.${CARTOGRAPHER_MINOR_VERSION})
 
 include("${CMAKE_SOURCE_DIR}/cmake/functions.cmake")
-google_initialize_cartographer_project(True)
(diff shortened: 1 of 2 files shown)


cmake/functions.cmake
```diff
@@ -306,7 +325,7 @@ function(google_proto_library NAME)
   target_link_libraries("${NAME}" ${PROTOBUF_LIBRARY} pthread)
 endfunction()
 
-macro(google_initialize_cartographer_project)
+macro(google_initialize_cartographer_project ENABLE_TESTING)
```
damonkohler: Pull out a different macro to enable testing.

SirVer:
all done.

CMakeLists.txt
```diff
@@ -23,6 +23,7 @@ set(CARTOGRAPHER_SOVERSION ${CARTOGRAPHER_MAJOR_VERSION}.${CARTOGRAPHER_MINOR_VE
 
 include("${CMAKE_SOURCE_DIR}/cmake/functions.cmake")
 google_initialize_cartographer_project()
```
damonkohler: This should probably be renamed to google_initialize_project() since it applies to more projects now. Can wait for another PR, though.

damonkohler:
@tfoote or @dirk-thomas could you have a look at ros/catkin:#830? See the change to functions.cmake in this PR.

PR #51, opened 2016-10-12
Change contribution link to local link

zachpanz88:
Set the link to be a local link rather than a full URL


zachpanz88:
This really isn't a needed change, closing

PR #59, opened 2016-10-13
Show the unprojected point cloud in 2D SLAM.

wohe:
We now project to 2D later in 2D SLAM, so that the roll and pitch
applied to the laser fan in visible in the visualization.

### DIFF
--- a/cartographer/sensor/laser.h
+++ b/cartographer/sensor/laser.h
@@ -49,8 +49,8 @@ struct LaserFan3D {
 };
 
 // Builds a LaserFan from 'proto' and separates any beams with ranges outside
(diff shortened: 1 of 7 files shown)


cartographer/sensor/laser.h
```diff
@@ -52,6 +48,10 @@ struct LaserFan3D {
   std::vector<uint8> reflectivities;
 };
 
+// Builds a LaserFan from 'proto' and separates any beams with ranges outside
+// the range ['min_range', 'max_range']. Beams beyond 'max_range' inserted into
```
SirVer: Beams beyond... *are*

PR #79, opened 2016-10-19
Add a PCD file creator.

at-wat:
This commit adds PCD file (PCD_v7) creator based on PointsProcessor class.

### DIFF
--- a/cartographer/io/pcd_points_processor.cc
+++ b/cartographer/io/pcd_points_processor.cc
@@ -1,0 +1,78 @@
+/*
+ * Copyright 2016 The Cartographer Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


SirVer:
Stuff changed a bit in the repo, so this needs updating. Sorry for that :/

Your implementation throws away a lot of points, was that intentional? I think this should look similar to  [the PLY writer](https://github.com/googlecartographer/cartographer/blob/master/cartographer/io/ply_writing_points_processor.cc) and retain all points in the output and also stream the points to disk.

cartographer/io/pcd_points_processor.cc
```diff
@@ -0,0 +1,78 @@
...
+  fs << "VERSION .7" << std::endl;
+  fs << "FIELDS x y z" << std::endl;
+  fs << "SIZE 4 4 4" << std::endl;
+  fs << "TYPE F F F" << std::endl;
+  fs << "COUNT 1 1 1" << std::endl;
+  fs << "WIDTH " << num << std::endl;
+  fs << "HEIGHT 1" << std::endl;
+  fs << "VIEWPOINT 0 0 0 1 0 0 0" << std::endl;
+  fs << "POINTS " << num << std::endl;
+  fs << "DATA ascii" << std::endl;
```
SirVer: this should probably write out a binary file, otherwise size explodes really quickly.

at-wat:
My first opinion was to use the PCD output directly as a map for localization.
However, considering other usages, full points in binary format would be better.

I'll take a look around the changes.

at-wat:
I'll open a new PR.

PR #80, opened 2016-10-19
Adds a PLY writing points processor.

SirVer:


cartographer/io/ply_writing_points_processor.cc
```diff
@@ -0,0 +1,111 @@
...
+
+  WriteBinaryPlyHeader(has_colors_, num_points_, &file_);
+  file_.flush();
+
+  switch (next_->Flush()) {
+    case FlushResult::kFinished:
+      return FlushResult::kFinished;
+
+    case FlushResult::kRestartStream:
+      LOG(FATAL) << "Not supported";
```
damonkohler: Same as with X-ray.

PR #81, opened 2016-10-19
Unify to AddLaserFan for both 2D and 3D.

wohe:


cartographer/mapping_2d/global_trajectory_builder.h
```diff
@@ -40,16 +40,14 @@ class GlobalTrajectoryBuilder
   const mapping::GlobalTrajectoryBuilderInterface::PoseEstimate& pose_estimate()
       const override;
 
-  void AddHorizontalLaserFan(common::Time time,
-                             const sensor::LaserFan& laser_fan) override;
+  // Handles approximately horizontal laser fans.
```
SirVer: approximately? Could use a bit more explanation.

wohe: Will fix it in the follow up PR.

PR #86, opened 2016-10-19
Adds a PointProcessor that writes ASCII files.

SirVer:
The files will just contain XYZ points on a file of their own. This is
useful for example as interop with AutoCAD.

### DIFF
--- a/cartographer/io/xyz_writing_points_processor.cc
+++ b/cartographer/io/xyz_writing_points_processor.cc
@@ -6,6 +6,7 @@
 
 namespace cartographer {
 namespace io {
+
 namespace {
 
 void WriteXyzPoint(const Eigen::Vector3f& point, std::ofstream* output) {
@@ -16,7 +17,7 @@ void WriteXyzPoint(const Eigen::Vector3f& point, std::ofstream* output) {
 
 XyzWriterPointsProcessor::XyzWriterPointsProcessor(const string& filename,
                                                    PointsProcessor* next)
-    : next_(next), file_(filename, std::ios_base::out) {
+    : next_(next), file_(filename, std::ios_base::out | std::ios_base::binary) {
   file_ << std::setprecision(6);
 }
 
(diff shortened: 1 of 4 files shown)


cartographer/io/xyz_writing_points_processor.cc
```diff
@@ -0,0 +1,45 @@
...
+
+    case FlushResult::kRestartStream:
+      LOG(FATAL) << "XYZ generation must be configured to occur after any "
+                    "stages that require multiple passes.";
+  }
+  LOG(FATAL);
+}
+
+void XyzWriterPointsProcessor::Process(std::unique_ptr<PointsBatch> batch) {
+  for (size_t i = 0; i < batch->points.size(); ++i) {
```
wohe: Range-based for loop?

cartographer/io/xyz_writing_points_processor.cc
```diff
@@ -0,0 +1,45 @@
...
+
+void WriteXyzPoint(const Eigen::Vector3f& point, std::ofstream* output) {
+  (*output) << point.x() << " " << point.y() << " " << point.z() << "\n";
+}
+
+}  // namespace
+
+XyzWriterPointsProcessor::XyzWriterPointsProcessor(const string& filename,
+                                                   PointsProcessor* next)
+    : next_(next), file_(filename, std::ios_base::out) {
```
wohe: Make it binary, so that it would give similar output under Windows?

cartographer/io/xyz_writing_points_processor.cc
```diff
@@ -0,0 +1,45 @@
+#include "cartographer/io/xyz_writing_points_processor.h"
+
+#include <iomanip>
+
+#include "glog/logging.h"
+
+namespace cartographer {
+namespace io {
+namespace {
```
damonkohler: Blank line before.

PR #95, opened 2016-10-25
Make all 'PointProcessor's registerable with the PipelineBuilder.

SirVer:
Also the default instance has now knowledge of all of Cartographer's
'PointsProcessor's by default.

### DIFF
--- a/cartographer/io/pcd_writing_points_processor.cc
+++ b/cartographer/io/pcd_writing_points_processor.cc
@@ -18,6 +18,9 @@
 
 #include <iomanip>
 
+#include "cartographer/common/lua_parameter_dictionary.h"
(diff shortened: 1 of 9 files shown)


cartographer/io/pcd_writing_points_processor.cc
```diff
@@ -28,7 +31,7 @@ namespace {
 // Writes the PCD header claiming 'num_points' will follow it into
 // 'output_file'.
 void WriteBinaryPcdHeader(const bool has_color, const int64 num_points,
-                          std::ofstream* stream) {
+                          std::ofstream* const stream) {
   string color_header_field = !has_color ? "" : " rgb";
```
damonkohler: const

PR #97, opened 2016-10-25
Add serialization of the sparse pose graph.

wohe:


damonkohler:
FYI @macmason @athackst

PR #99, opened 2016-10-25
Adds a voxel filtering and moving objects removing points processor.

SirVer:


cartographer/io/outlier_removing_points_processor.cc
```diff
@@ -0,0 +1,118 @@
...
+      return FlushResult::kRestartStream;
+
+    case State::kPhase2:
+      LOG(INFO) << "Filtering outliers...";
+      state_ = State::kPhase3;
+      return FlushResult::kRestartStream;
+
+    case State::kPhase3:
+      CHECK(next_->Flush() == FlushResult::kFinished)
+          << "Restarting not implemented.";
```
damonkohler: Same comment as in the other processors?

PR #100, opened 2016-10-25
Make RegisterType public.

SirVer:
This is required for more complex points processor that need to sideload
data.


SirVer:
As discussed, keeping this internal as it waters down the public API.

PR #106, opened 2016-10-27
CHECKs that all queues are finished on destruction.

damonkohler:


cartographer/sensor/ordered_multi_queue.h
```diff
@@ -50,15 +50,20 @@ inline std::ostream& operator<<(std::ostream& out, const QueueKey& key) {
...
 class OrderedMultiQueue {
  public:
   using Callback = std::function<void(std::unique_ptr<Data>)>;
 
-  // Will wait to see at least one value for each unfinished queue before
-  // dispatching the next smallest value across all queues.
   OrderedMultiQueue() {}
-  ~OrderedMultiQueue() {}
+
+  // CHECKs that all queues are finished.
```
wohe: Should this comment be there? The code is saying this clearly, which is preferable.

PR #107, opened 2016-10-27
Add num_available_packets() to MapBuilder.

damonkohler:


damonkohler:
This API isn't necessary. Dropping it.

PR #124, opened 2016-11-15
Drop sensor data until progress can be made.

damonkohler:
The sensor collator now keeps only the most recent packet for each queue
until data is available for all queues. For example, this prevents us
from processing arbitrarily many IMU packets before our first scan
match.

### DIFF
--- a/cartographer/sensor/ordered_multi_queue.h
+++ b/cartographer/sensor/ordered_multi_queue.h
@@ -159,49 +159,23 @@ class OrderedMultiQueue {
 
       // If we haven't dispatched any data yet, fast forward all queues until a
       // common start time has been reached.
-      if (last_dispatched_time_ == common::Time::min() &&
-          !FastForwardToCommonStartTime()) {
-        return;
+      if (common_start_time_ == common::Time::min()) {
+        for (auto& entry : queues_) {
+          common_start_time_ =
+              std::max(common_start_time_, entry.second.queue.Peek<Data>()->time);
+        }
+        LOG(INFO) << "All sensor data is available starting at '"
+                  << common_start_time_ << "'.";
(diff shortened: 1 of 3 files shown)


cartographer/sensor/ordered_multi_queue.h
```diff
@@ -151,19 +148,62 @@ class OrderedMultiQueue {
...
+  // Drops data until all queues have reached a common start time and are ready
+  // for dispatching to begin. Returns true if dispatching is ready to begin.
+  bool FastForwardToCommonStartTime() {
+    if (common_start_time_ == common::Time::min()) {
+      common_start_time_ =
+          std::max_element(queues_.begin(), queues_.end(),
+                           [](decltype(queues_)::value_type& lhs,
+                              decltype(queues_)::value_type& rhs) {
+                             const auto* const lhs_data =
+                                 CHECK_NOTNULL(lhs.second.queue.Peek<Data>());
```
wohe: Remove duplication by turning this into a for loop?

cartographer/sensor/ordered_multi_queue.h
```diff
@@ -151,19 +148,62 @@ class OrderedMultiQueue {
...
+                                 CHECK_NOTNULL(rhs.second.queue.Peek<Data>());
+                             return lhs_data->time < rhs_data->time;
+                           })
+              ->second.queue.Peek<Data>()
+              ->time;
+      LOG(INFO) << "All sensor data is available starting at '"
+                << common_start_time_ << "'.";
+    }
+    bool ready = true;
+    for (auto& entry : queues_) {
```
wohe: Could this be replaced by pop & continue in line 166?

PR #133, opened 2016-11-18
Change dropping data before start to be idempotent.

wohe:
We drop as much data from each queue as possible without changing the
start time that would be determined if we only saw the dispatched
data, i.e. we retain one piece of data not beyond the common start time.
This makes the process idempotent.

### DIFF
--- a/cartographer/sensor/ordered_multi_queue.h
+++ b/cartographer/sensor/ordered_multi_queue.h
@@ -168,13 +168,27 @@ class OrderedMultiQueue {
                   << common_start_time_ << "'.";
       }
 
-      if (next_data->time < common_start_time_) {
-        next_queue->queue.Pop();
-        continue;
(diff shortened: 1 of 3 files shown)


cartographer/sensor/ordered_multi_queue.h
```diff
@@ -168,13 +168,27 @@ class OrderedMultiQueue {
...
+      } else if (next_queue->queue.Size() < 2) {
+        if (!next_queue->finished) {
+          // We cannot decide whether to drop or dispatch this yet.
+          return;
+        }
+        last_dispatched_time_ = next_data->time;
+        next_queue->callback(next_queue->queue.Pop());
+      } else {
+        // We take a peek at the time after next data. If it also is before the
+        // start time we drop 'next_data', otherwise we just found the first
```
damonkohler: before 'common_start_time_'

PR #160, opened 2016-12-13
Revert to workaround #158.

wohe:
Currently, on xenial/kinetic the transitive dependencies on
libcartographer and its dependencies are not duplicated after
each library that depends on it in the linker command.


SirVer:
lgtm. Oh man, cmake can really ruin a great day.

PR #161, opened 2016-12-13
Support fixing z during 3D sparse pose adjustment.

wohe:
This is in preparation of using the 3D sparse pose graph
optimization for 2D SLAM.


cartographer/mapping_3d/sparse_pose_graph/optimization_problem.cc
```diff
@@ -72,8 +72,9 @@ struct ConstantYawQuaternionPlus {
 
 OptimizationProblem::OptimizationProblem(
     const mapping::sparse_pose_graph::proto::OptimizationProblemOptions&
-        options)
-    : options_(options) {}
+        options,
+    bool fix_z)
```
SirVer: enum instead of bool?

PR #170, opened 2016-12-21
Rerun CMake when files are added/removed.

damonkohler:


cmake/functions.cmake
```diff
@@ -106,6 +106,18 @@ macro(google_initialize_cartographer_project)
 
   message(STATUS "Build type: ${CMAKE_BUILD_TYPE}")
 
+  # Add a hook that reruns CMake when source files are added or removed.
+  set(LIST_FILES_CMD "find ${CMAKE_SOURCE_DIR}/ -not -iwholename '*.git*' | sort | sed 's/^/#/'")
+  set(FILES_LIST_PATH "${CMAKE_BINARY_DIR}/AllFiles.cmake")
+  set(DETECT_CHANGES_CMD "bash" "-c" "${LIST_FILES_CMD} | diff -N -q ${FILES_LIST_PATH} - || ${LIST_FILES_CMD} > ${FILES_LIST_PATH}")
```
SirVer: break line

cmake/functions.cmake
```diff
@@ -106,6 +106,18 @@ macro(google_initialize_cartographer_project)
...
+  add_custom_target(regen ALL
```
SirVer: use a name that is never conflicting with some cmake target.

cmake/functions.cmake
damonkohler: Not possible. Requires CMake > 3.0

https://cmake.org/cmake/help/v3.0/manual/cmake-language.7.html#quoted-argument

cmake/functions.cmake
damonkohler: Done.

PR #187, opened 2017-01-24
Make intensities a separate channel in PointsBatch.

### LINKED ISSUE #171: Change points processor pipeline to have an intensity channel
and add a points processor that can convert from intensities to RGB values. This should replace the awkward intensity command line flags for the asset writer in cartographer_ros.

SirVer:
And add a PointsProcessor that translates from intensities to RGB.

Fixes #171.

### DIFF
--- a/cartographer/io/coloring_points_processor.cc
+++ b/cartographer/io/coloring_points_processor.cc
@@ -30,7 +30,9 @@ ColoringPointsProcessor::FromDictionary(
   const string frame_id = dictionary->GetString("frame_id");
   const std::vector<double> color_values =
       dictionary->GetDictionary("color")->GetArrayValuesAsDoubles();
-  Color color = {{color_values[0], color_values[1], color_values[2]}};
+  Color color = {{static_cast<uint8_t>(color_values[0]),
+                  static_cast<uint8_t>(color_values[1]),
+                  static_cast<uint8_t>(color_values[2])}};
   return common::make_unique<ColoringPointsProcessor>(color, frame_id, next);
 }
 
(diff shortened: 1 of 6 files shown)


cartographer/io/coloring_points_processor.cc
```diff
@@ -30,7 +30,9 @@ ColoringPointsProcessor::FromDictionary(
   const string frame_id = dictionary->GetString("frame_id");
   const std::vector<double> color_values =
       dictionary->GetDictionary("color")->GetArrayValuesAsDoubles();
-  Color color = {{color_values[0], color_values[1], color_values[2]}};
+  Color color = {{static_cast<uint8_t>(color_values[0]),
```
damonkohler: const

cartographer/io/points_batch.h
```diff
@@ -51,7 +51,17 @@ struct PointsBatch {
   // Trajectory index that produced this point.
   int trajectory_index;
 
+  // The geometry of the points in a metric frame.
```
damonkohler: Remove "The"

cartographer/io/points_batch.h
```diff
@@ -51,7 +51,17 @@ struct PointsBatch {
...
   std::vector<Eigen::Vector3f> points;
+
+  // Intensities are optional (might be empty). They represent the by device
```
damonkohler: Intensities are optional and may be unspecified. The meaning of these intensity values varies by device. For example, the VLP16 provides values in the range [0, 100] for non-specular return values and values up to 255 for specular returns. On the other hand, Hokuyo lasers provide a 16-bit value that rarely peaks above 4096.

PR #192, opened 2017-01-26
Actually check that all points are colored or none.

SirVer:


cartographer/io/ply_writing_points_processor.cc
```diff
@@ -109,9 +109,15 @@ void PlyWritingPointsProcessor::Process(std::unique_ptr<PointsBatch> batch) {
     has_colors_ = !batch->colors.empty();
     WriteBinaryPlyHeader(has_colors_, 0, file_.get());
   }
+  if (has_colors_) {
+    CHECK(!batch->colors.empty()) << "First PointsBatch had colors, but "
```
wohe: `CHECK_EQ(batch->colors.size(), batch->points.size())`?

PR #196, opened 2017-01-31
math.h NormalizeAngleDifference infinite loop

ahundt:
Consider the case of:
```C++
NormalizeAngleDifference(std::numeric_limits<double>::max());
```

The previous version of this function will be stuck in an infinite loop because: 

```c++
std::numeric_limits<double>::max() == std::numeric_limits<double>::max() - M_PI
```

In addition to eliminating the infinite loop, the updated version is frequently noticeably faster when the the value is well outside the -pi to pi range.

This file change is based on the following stackoverflow post of my own creation:
http://stackoverflow.com/a/11126083/99379

Also, please consider switching to this function name as it may be more clear:
```C++
template<typename T>
inline T NormalizeRadiansPiToMinusPi(T rad);
```

### DIFF
--- a/cartographer/common/math.h
+++ b/cartographer/common/math.h
@@ -71,7 +71,7 @@ T NormalizeAngleDifference(T difference) {
   // Copy the sign of the value in radians to the value of pi.
   T signed_pi = std::copysign(M_PI,difference);
   // Set the value of difference to the appropriate signed value between pi and -pi.
-  difference = std::fmod(difference + signed_pi,(2 * M_PI)) - signedPI;
+  difference = std::fmod(difference + signed_pi,(2 * M_PI)) - signed_pi;
   return difference;
 }
 
--- a/cartographer/common/math.h
+++ b/cartographer/common/math.h
@@ -68,12 +68,10 @@ constexpr double RadToDeg(double rad) { return 180. * rad / M_PI; }
 // Bring the 'difference' between two angles into [-pi; pi].
 template <typename T>
 T NormalizeAngleDifference(T difference) {
-  while (difference > M_PI) {
-    difference -= T(2. * M_PI);
-  }
-  while (difference < -M_PI) {
-    difference += T(2. * M_PI);
-  }
+  // Copy the sign of the value in radians to the value of pi.
+  T signed_pi = std::copysign(M_PI,difference);
+  // Set the value of difference to the appropriate signed value between pi and -pi.
+  difference = std::fmod(difference + signed_pi,(2 * M_PI)) - signedPI;
   return difference;
 }
 


ahundt:
Going to need help with that compiler error as I'm not familiar with how to deal with a Ceres::Jet. I enabled edits from maintainers.

wohe:
Hi, I don't think this is a good change for several reasons:

1. Normalizing `std::numeric_limits<double>::max()` does not make sense in the first place, and changing it from an infinite loop does not improve the situation. If you wanted to document more clearly that this is only reasonable up to some magnitude, rather add a `CHECK` to the function. Note that this is a theoretical discussion and large values should not appear. In practice, you'd expect this function to take at most one iteration for each loop.

2. The current code is more readable.

3. I'm pretty sure performance of this function does not matter. If you think it does: Did you profile?

4. The reason this does not compile is because of automatic differentiation. The functions copysign and fmod are not defined for jets to my knowledge. If they were, you'd still have to remove the `std::`. Fixing this would be trivial though, since you can just take the scalar value of the jet, do the computation, and convert back to a jet. The reason why this is possible is because this function is piecewise constant.

Please let me know if you have further questions.

ahundt:
I was just giving an example failure case, it fails on many values smaller than max which was just convenient for describing the problem. Yes, I profiled but it was years ago in 2012 so I no longer have the data. I previously worked on autonomous vehicles and a similar implementation with while loops was a difficult to detect cause of our system stalling out. It was triggered by some bad data from a sensor getting through. The worst case behavior of the pre-pull code, O(n) or even an infinite loop, is much worse than the overhead of a call to fmod.

There are several [other implementations on the linked page](http://stackoverflow.com/a/11126083/99379) if you find one of them mathematically cleaner.

ahundt:
Is there any example code for converting jets to/from scalars you can point me to? I assume that would be a template instantiation.

Also, if your typical case is already between -pi and pi it is easy enough to check if the value is already in the correct range and skip the fmod call altogether at the cost of an extra branch. I doubt it matters either way unless profiling determines this call is a bottleneck.

Honestly, I was just looking for examples of integrating protobufs with eigen matrices and saw a exact replica of a serious bug I've encountered in the past, fixed, and ran on a robot and in testing for many many hours. I actually won't have a use case for googlecartographer for a while, as I'm currently working on stationary arms with an RGBD camera. Or perhaps if the time comes when 3D slam + RGBD support with colorized pixels is implemented. :-)

wohe:
Thanks for your comments. We agree that this function will not work well for large values, no need to convince me. The point is that this function serves the purpose of normalizing angle differences that arise during SLAM, i.e. they will almost always be small. From this point of view adding support for large input values does not make sense since this only happens when the algorithm is not working as intended. For the same reason, one could add a `CHECK`, since this would make it obvious.

Regarding the performance: Of course I meant profiling the Cartographer code. I'm pretty sure this function is not contributing significantly to the running time. I prefer readability in this case.

Regarding jets: You just access the scalar member as `.a`, and specialize in the case of `double` where this is neither needed nor possible. After the computation you construct a new jet, again only in the templated version intended for handling jets. But if this complicates the code for no measurable gain, I would suggest not doing it.

ahundt:
Sure, feel free to close this pull request if desired. This conversation will still be here in the history if someone encounters the problem.

wohe:
For future reference: Of course the function is not piecewise constant, but the
difference between function argument and result. An implementation that supports
automatic differentiation would be along these lines:

```
inline double NormalizeAngleDifference(double difference) {
  // Copy the sign of the value in radians to the value of pi.
  double signed_pi = std::copysign(M_PI,difference);
  // Set the value of difference to the appropriate signed value between pi and -pi.
  difference = std::fmod(difference + signed_pi,(2 * M_PI)) - signed_pi;
  return difference;
}

// For jets:
template <typename T>
T NormalizeAngleDifference(T difference) {
  const double piecewise_constant =
      NormalizeAngleDifference(difference.a) - difference.a;
  return difference + T(piecewise_constant);
}
```

PR #203, opened 2017-02-09
Enable adding as a CMake subproject

ojura:
As described [here](https://groups.google.com/forum/#!topic/google-cartographer/DmA-rCdUZgA).


SirVer:
Nice change! Thank you.

Issue #197, opened 2017-02-01
Support for loading saved submaps and pose graph for long-term SLAM.


ajakhotia:
Hi Authors and other Contributors
    Let me start by thanking you all for writing Cartographer and open-sourcing it. It adds immense value to the robotics society as it is probably the only SLAM algorithm as of now that is field-ready. Great job!

I am using Cartographer with our in-house framework to create a solution for a long-term 2D mapping and localization. For the implementation to be truly long-term, I need to add the ability of being able to save the pose of the robot, sub-maps and the pose-graph(SPA) on shutdown and load them back into cartographer on subsequent startup of the software. I will be beginning my efforts on this very soon and was wondering if you guys have any thoughts, pointers to references or inputs before I begin. I would be more than happy to collaborate as suitable.

Here's a quick brain dump of what I think needs to be done to accomplish this:

- Serialization, saving, loading and construction of sub-maps in local trajectory builder(front-end) and sparse pose graph in global trajectory builder.

- Saving and loading of robot pose(with covariance?) and instantiation of map builder with this pose as the starting point.

- A global matching scheme to match the new scans to the loaded sub-maps in the vicinity(scaled based on covariance?) of the seed pose before cartographer resurrects and operates as normal. May be?

- A strategy to appropriately kick-start the internal kalman filters. Not sure how the time-jump will affect those.

- Swapping in new laser frames to replace stale frames in the pose-graph that are no longer representative of the environment. Also limiting the insertion of laser scans to situations when new area is being explored or when swapping out an stale scan . This will
    1. Avoid growth of pose-graph over course of time to an intractable number of constraints. 
    2. Make the pose-graph adaptive to the long-term, slow changes in the environment.

Please do share your thoughts that you think might help. I will be updating this post as I think of more items.

Thanks

-- Anurag

SirVer:
@wohe @macmason @yutakaoka  We had several discussions with TRI about this very idea, but afaik no real work has started on this. @TRI: Guys, what is your status on this?

haidai:
I'm also looking a similar case but we're using cartographer components and data types as a starting point and not worrying as much about the submaps for now since it's hard enough to do good global localization and figuring out how to edit graphs without having to do recursively also. Might be a helpful direction in your case too?

yutakaoka:
[...]

PR #220, opened 2017-03-09
HybridGrid proto.

macmason:


macmason:
Related to issue #197

cartographer/mapping_3d/hybrid_grid_test.cc
```diff
@@ -177,6 +178,120 @@ TEST(HybridGridTest, TestIteration) {
...
+  }
+  // 105 is the hand-verified number of times we added something.
+  ASSERT_EQ(105, grid_size);
+
+  proto = ToProto(hybrid_grid);
+  ASSERT_EQ(grid_size, proto.indices_size());
+  ASSERT_EQ(grid_size, proto.values_size());
+
+  int proto_index = 0;
+  for (const auto& entry : hybrid_grid) {
```
wohe: This is very close to repeating the implementation and thus does not provide a lot of confidence in the correctness. Did you consider verifying that `HybridGrid(ToProto(original)) == original` as a test case? Maybe with (pseudo) randomly generated content.

Similar for the other test case. Combining the two into one then probably makes sense.

cartographer/mapping_3d/hybrid_grid_test.cc
```diff
@@ -177,6 +178,120 @@ TEST(HybridGridTest, TestIteration) {
...
+    ASSERT_EQ(entry.second, proto.values(proto_index));
+    proto_index++;
+  }
+}
+
+namespace {
+
+struct EigenComparator {
+  bool operator()(const Eigen::Vector3i lhs, const Eigen::Vector3i& rhs) {
+    return (lhs.x() < rhs.x()) || (lhs.x() == rhs.x() && lhs.y() < rhs.y()) ||
```
wohe: Use `<tuple>` and do
```
return std::forward_as_tuple(lhs.x(), lhs.y(), lhs.z()) <
    std::forward_as_tuple(rhs.x(), rhs.y(), rhs.z());
```
or so?

cartographer/transform/transform.cc
```diff
@@ -99,6 +103,14 @@ proto::Vector3d ToProto(const Eigen::Vector3d& vector) {
   return proto;
 }
 
+proto::Vector3i ToProto(const Eigen::Vector3i& vector) {
+  proto::Vector3i result;
```
macmason: Done.

cartographer/mapping_3d/hybrid_grid_test.cc
macmason: Major rewrite here; PTAL.

macmason:
PTAL.

PR #222, opened 2017-03-18
add #include <array> in files which use std::array

wjwwood:
This pull request adds some missing `#include <array>` in files which directly use `std::array`. Without these includes, this fails to compile with clang on macOS, for example:

```
[  1%] Building CXX object CMakeFiles/cartographer.dir/cartographer/io/coloring_points_processor.cc.o
In file included from /Users/william/ros2_ws/src/vendor/cartographer/cartographer/io/coloring_points_processor.cc:17:
/Users/william/ros2_ws/src/vendor/cartographer/cartographer/io/coloring_points_processor.h:49:15: error: implicit instantiation of undefined template 'std::__1::array<unsigned char, 3>'
  const Color color_;
              ^
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/../include/c++/v1/__tuple:116:92: note: template is declared here
template <class _Tp, size_t _Size> struct __attribute__ ((__type_visibility__("default"))) array;
                                                                                           ^
```

### DIFF
--- a/cartographer/io/points_batch.h
+++ b/cartographer/io/points_batch.h
@@ -17,6 +17,7 @@
 #ifndef CARTOGRAPHER_IO_POINTS_BATCH_H_
 #define CARTOGRAPHER_IO_POINTS_BATCH_H_
 
+#include <array>
 #include <cstdint>
 #include <vector>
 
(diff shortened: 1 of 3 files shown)


wjwwood:
I signed it!

Issue #193, opened 2017-01-26
Velocity speed estimate is noisy and unstable with a higher frequency laser (previously: Cartographer crashes with a Pioneer 3DX bag)


ojura:
I am getting a crash when mapping with Cartographer and a Pioneer robot. I am using Ubuntu 16.04, ROS Kinetic, cartographer @ddb3c890a6 and googlecartographer/cartographer_ros@f7057fc5a35aed4. A sample crash log can be found [here](https://gist.github.com/ojura/c171a38030b68ef32faf7960de301175). The launch file and the appropriate bag can be found [here](http://larics.fer.hr/owncloud/index.php/s/COT7XqqmFOIdKuu). To run, extract the archive, change into the `launch` directory and call `roslaunch pioneer_cartographer.launch` from there.

The crash usually occurs somewhere around 60-70 seconds into the bag (counting from the 25th second, where it starts in the launch file) , after the robot makes a right turn to the hallway looking up. The pose estimate makes wild jumps just before it crashes, with a message like the following one: `[FATAL]: ... probability_grid.h:161] Check failed: limits_.Contains(xy_index) -215`

The crash seems to be related to odometry - if `use_odometry` is set to `false` in the `.lua` file, Cartographer doesn't crash. However, scan matching in that case doesn't work very well; it gets stuck in the same hallway where it crashes when using odometry.

wohe:
Hi, this kind of crash happens when the laser scan that is about to be inserted is no longer contained by the current submap. The size of the submap can be configured in `trajectory_builder_2d.lua` and defaults to 400 meters. I'm not suggesting to change it though, it seems appropriate for your setup. The underlying problem is that the SLAM is not tracking well, i.e. a tuning issue.

I took a quick look at your bag and it seems neither the laser nor the odometry is providing the quality expected by the default settings. If you set

```
TRAJECTORY_BUILDER_2D.use_online_correlative_scan_matching = true
TRAJECTORY_BUILDER_2D.real_time_correlative_scan_matcher.linear_search_window = 0.
TRAJECTORY_BUILDER_2D.submaps.resolution = 0.1
```

you can generate a map from your bag. The coarser resolution I chose because of the quality of the laser scan, the use of the correlative scan matcher to figure out the rotation with neither IMU nor odometry. At this point you can even disable the use of odometry and still get a good map.

Please let me know if you have further questions.

ojura:
Hi @wohe,

[...]

PR #225, opened 2017-03-23
Add laser decimation

ojura:
This turned out to be the best mitigation of #193 for me.

I attempted to do the discarding of laser messages as early in the call stack of processing them as I could, while being able to access the trajectory builder options, where I think this option should be located. This turned out to be in the global trajectory builder class for the 2D case. I expect that you might have a better idea where this should go.

Also included are two completely miscellaneous things that I decided to add in the pull request so I don't have to keep them stashed the whole time. :)


cartographer/mapping_2d/global_trajectory_builder.cc
```diff
@@ -35,6 +36,17 @@ const Submaps* GlobalTrajectoryBuilder::submaps() const {
 void GlobalTrajectoryBuilder::AddRangefinderData(
     const common::Time time, const Eigen::Vector3f& origin,
     const sensor::PointCloud& ranges) {
+  // TODO average laser scans instead of discarding them?
```
wohe: Not sure what you mean by averaging, but would it also help in your case if you combine multiple scans using the predicted poses in the `LocalTrajectoryBuilder` introducing a parameter `scans_per_accumulation`? This would be similar to the 3D case if you look at the `mapping_3d/kalman_local_trajectory_builder.cc` for example.

Otherwise we now add a stop gap solution that we will remove again in the future, instead of moving a smaller step in the direction we would like to go.

ojura: I agree, this is indeed a stop gap solution, I was not sure if I should submit a pull request for this at all. I decided to go for it because it really made a big improvement on my bag from #193. 

I will take a look at the 3D case and try to do it that way. If it also gives good results, I will update the pull request and replace scan dropping with scan accumulation.

The TODO was just an idea I thought would make sense - by "averaging" I meant averaging the rangefinder data. E.g. the distance in some direction theta is the average of the distances in the same direction in these sequential scans which would otherwise be dropped. Perhaps it's the same thing as the scan accumulation from 3D that you have mentioned - at this point I can't say because I haven't looked into it yet.

ojura:
I will close the pull request for now, until I take a look at the 3D local trajectory builders. I will make separate pull requests for those two misc commits.

PR #227, opened 2017-03-23
Multi trajectory toproto

macmason:
(Note that this branch is based on the proto_fixup branch in #226)

@wohe I _think_ I'm on the right track here, but very close scrutiny is appreciated. There are also some cleanups (in particular, now-obsolete functions) that will be included here, assuming it passes your sanity-check.

(And some tests...)

Related to #197 .


cartographer/mapping/proto/sparse_pose_graph.proto
```diff
@@ -44,8 +44,9 @@ message SparsePoseGraph {
     // Pose of the scan relative to submap, i.e. taking data from the scan frame
     // into the submap frame.
     optional transform.proto.Rigid3d relative_pose = 3;
-    // 6x6 square root inverse of the covariance matrix.
-    repeated double sqrt_Lambda = 4;  // NOLINT
+    // 6x6 square root inverse of the covariance matrix. This is stored in
+    // column-major order for ease of Eigen::Map.
```
wohe: Shouldn't the covariance be a symmetric matrix, hence the square root inverse as well, i.e. be in column-major and row-major order at the same time?

cartographer/mapping/sparse_pose_graph.cc
```diff
@@ -74,6 +74,17 @@ proto::SparsePoseGraphOptions CreateSparsePoseGraphOptions(
 
 proto::SparsePoseGraph SparsePoseGraph::ToProto() {
   proto::SparsePoseGraph proto;
+
+  // The proto uses integer 'trajectory_id's, but the SPG itself doesn't track
```
wohe: Nit: Write out 'sparse pose graph'.

cartographer/mapping/sparse_pose_graph.cc
```diff
@@ -84,22 +95,39 @@ proto::SparsePoseGraph SparsePoseGraph::ToProto() {
...
-  const std::vector<std::vector<const Submaps*>> components =
-      GetConnectedTrajectories();
-  CHECK_EQ(components.size(), 1);
-  CHECK_EQ(components[0].size(), 1);
-  const Submaps* const submaps = components[0][0];
-  for (const auto& transform : GetSubmapTransforms(*submaps)) {
-    *trajectory->add_submap()->mutable_pose() = transform::ToProto(transform);
+  for (const auto node : trajectory_nodes) {
+    const int trajectory_index =
+        trajectory_indices[node.constant_data->trajectory];
```
wohe: Can we use `.at()` here and in other places to make it obvious we are not adding entries to the map?

cartographer/mapping/sparse_pose_graph.cc
```diff
@@ -84,22 +95,39 @@ proto::SparsePoseGraph SparsePoseGraph::ToProto() {
...
+    // Make sure proto.trajectories(trajectory_index) is valid.
+    while (proto.trajectory_size() <= trajectory_index) {
```
wohe: This should by the way the indices are chosen add at most one trajectory, right?

cartographer/mapping/sparse_pose_graph.cc
```diff
@@ -84,22 +95,39 @@ proto::SparsePoseGraph SparsePoseGraph::ToProto() {
...
     constraint_proto->mutable_scan_id()->set_scan_index(constraint.j);
+    constraint_proto->mutable_scan_id()->set_trajectory_id(
+        trajectory_indices[trajectory_nodes[constraint.j]
+                               .constant_data->trajectory]);
     constraint_proto->set_tag(mapping::ToProto(constraint.tag));
   }
 
-  // TODO(whess): Support multi-trajectory.
-  proto::Trajectory* const trajectory = proto.add_trajectory();
-  *trajectory = mapping::ToProto(GetTrajectoryNodes());
```
wohe: I think this `mapping::ToProto()` function is not used anywhere else. Should it be removed since it only supports a single trajectory and should not be used in the future?

cartographer/mapping/sparse_pose_graph.cc
```diff
@@ -84,22 +95,39 @@ proto::SparsePoseGraph SparsePoseGraph::ToProto() {
...
+    while (proto.trajectory_size() <= trajectory_index) {
+      proto.add_trajectory();
+    }
+
+    auto* node_proto = proto.mutable_trajectory(trajectory_index)->add_node();
+    node_proto->set_timestamp(common::ToUniversal(node.constant_data->time));
+    *node_proto->mutable_pose() =
+        transform::ToProto(node.pose * node.constant_data->tracking_to_pose);
+
+    const Submaps* const submaps =
```
wohe: This should only be done once for each trajectory, not for each node of the trajectory, right? Move it after line 116 into the while loop?

cartographer/mapping/sparse_pose_graph.cc
```diff
@@ -84,22 +95,39 @@ proto::SparsePoseGraph SparsePoseGraph::ToProto() {
     Eigen::Map<Eigen::Matrix<double, 6, 6>>(
         constraint_proto->mutable_sqrt_lambda()->mutable_data()) =
         constraint.pose.sqrt_Lambda_ij;
-    // TODO(whess): Support multi-trajectory.
+
     constraint_proto->mutable_submap_id()->set_submap_index(constraint.i);
```
wohe: These indexes are still into the global list of nodes/submaps. I think you need two counters for each trajectory to remap the indexes to 0, 1, ... for each trajectory.

macmason: Ah, yes indeed.

Major change just landed; PTAL.

cartographer/mapping/sparse_pose_graph.cc
macmason: Done.

cartographer/mapping/sparse_pose_graph.cc
macmason: The new version loses this code.

cartographer/mapping/sparse_pose_graph.cc
macmason: Yes, but the code changes in the new version.

cartographer/mapping/sparse_pose_graph.cc
macmason: PTAL at where this ended up.

cartographer/mapping/sparse_pose_graph.cc
macmason: Obsoleted by later changes.

cartographer/mapping/sparse_pose_graph.cc
```diff
@@ -74,32 +80,52 @@ proto::SparsePoseGraphOptions CreateSparsePoseGraphOptions(
...
-  proto::Trajectory* const trajectory = proto.add_trajectory();
-  *trajectory = mapping::ToProto(GetTrajectoryNodes());
-  const std::vector<std::vector<const Submaps*>> components =
-      GetConnectedTrajectories();
-  CHECK_EQ(components.size(), 1);
-  CHECK_EQ(components[0].size(), 1);
-  const Submaps* const submaps = components[0][0];
-  for (const auto& transform : GetSubmapTransforms(*submaps)) {
-    *trajectory->add_submap()->mutable_pose() = transform::ToProto(transform);
+  for (size_t i = 0; i < grouped_nodes.size(); ++i) {
```
wohe: Range-based for loop? Same for the inner loop.

cartographer/mapping/sparse_pose_graph.h
```diff
@@ -33,9 +34,16 @@ namespace mapping {
 proto::SparsePoseGraphOptions CreateSparsePoseGraphOptions(
     common::LuaParameterDictionary* const parameter_dictionary);
 
-// Splits TrajectoryNodes by ID.
-std::vector<std::vector<TrajectoryNode>> SplitTrajectoryNodes(
-    const std::vector<TrajectoryNode>& trajectory_nodes);
+// TrajectoryNodes and Constraits are provided in a flat vector, but
```
wohe: typo: `Constraints`. Or better: Remove `and Constraints` since they remain to be in a flat vector.

cartographer/mapping/sparse_pose_graph.h
```diff
@@ -33,9 +34,16 @@ namespace mapping {
...
+// serialization requires that we group them by trajectory. This groups the
+// elements of 'trajectory_nodes' into 'grouped_nodes' (so that
+// *grouped_nodes[i] contains a complete single trajectory). The re-indexing
```
wohe: nit: `(*grouped_nodes)[i]` and similar in other places.

cartographer/mapping/sparse_pose_graph.cc
```diff
@@ -74,32 +80,52 @@ proto::SparsePoseGraphOptions CreateSparsePoseGraphOptions(
...
-    constraint_proto->mutable_submap_id()->set_submap_index(constraint.i);
-    constraint_proto->mutable_scan_id()->set_scan_index(constraint.j);
+
+    constraint_proto->mutable_submap_id()->set_trajectory_id(
+        new_indices[constraint.i].first);
+    constraint_proto->mutable_submap_id()->set_submap_index(
+        new_indices[constraint.i].second);
+
+    constraint_proto->mutable_scan_id()->set_trajectory_id(
+        new_indices[constraint.i].first);
```
macmason: Done.

cartographer/mapping/sparse_pose_graph.cc
macmason: Done.

cartographer/mapping/sparse_pose_graph.h
macmason: Done.

cartographer/mapping/sparse_pose_graph.h
macmason: Done.

cartographer/mapping/sparse_pose_graph.cc
```diff
@@ -74,32 +80,52 @@ proto::SparsePoseGraphOptions CreateSparsePoseGraphOptions(
...
     }
     Eigen::Map<Eigen::Matrix<double, 6, 6>>(
         constraint_proto->mutable_sqrt_lambda()->mutable_data()) =
         constraint.pose.sqrt_Lambda_ij;
-    // TODO(whess): Support multi-trajectory.
-    constraint_proto->mutable_submap_id()->set_submap_index(constraint.i);
-    constraint_proto->mutable_scan_id()->set_scan_index(constraint.j);
+
+    constraint_proto->mutable_submap_id()->set_trajectory_id(
+        new_indices[constraint.i].first);
```
macmason: If I understand correctly, my `j` values (the scan indices) are correct; each `TrajectoryNode` owns a single scan, and therefore the `new_indices` mapping gets us there. Right?

The submap indices are quite a bit trickier, because each `TrajectoryNode` (and therefore each trajectory_id) has a `Submaps*`, but doesn't know which `Submap` it belongs to.

I think. 

Right?

wohe: Correct regarding the `j` values and `TrajectoryNode`.

There is not a single submap for each node, but each submap belongs to a single trajectory, like the nodes. This information is currently not directly exposed, but `mapping_2d::SparsePoseGraph` (similar for 3D) has a `std::vector<SubmapState>` which is indexed by the flat submap indices and the `SubmapState` contains the trajectory the submap belongs to. This should allow us to compute a mapping to a `trajectory_id`/`submap_index` pair similar to what you did for the trajectory nodes.

cartographer/mapping/trajectory_node.h
```diff
@@ -67,8 +66,6 @@ struct TrajectoryNodes {
   std::vector<mapping::TrajectoryNode> trajectory_nodes;
 };
 
-proto::Trajectory ToProto(const std::vector<TrajectoryNode>& nodes);
```
Mofef: `cartographer_ros::WriteAssets` still calls this function. How will this be resolved?

SirVer: It will break when this PR gets merged, then @macmason will sent a PR to cartographer_ros to unbreak it. They will be merged very shortly after each other, so it should not be broken for long.

macmason:
Ok, this got pretty heavy-handed, but PTAL.

cartographer/mapping/trajectory_node.h
macmason: As @SirVer said, once this PR is LGTM, I'll follow up with one for `cartographer_ros`, and then somebody will hit the merge buttons very quickly.

jihoonl: @macmason Is it possible tocreate a PR for `cartographer_ros` in advance? I would like to see how you handled `ToProto` and `FromProto` in `cartographer_ros`. And if possible, I want to contribute on top of your changes.

cartographer/mapping_2d/sparse_pose_graph.cc
```diff
@@ -294,30 +294,25 @@ void SparsePoseGraph::WaitForAllComputations() {
   common::MutexLocker locker(&mutex_);
   const int num_finished_scans_at_start =
       constraint_builder_.GetNumFinishedScans();
-  while (!locker.AwaitWithTimeout(
-      [this]() REQUIRES(mutex_) {
-        return static_cast<size_t>(constraint_builder_.GetNumFinishedScans()) ==
-               trajectory_nodes_.size();
-      },
-      common::FromSeconds(1.))) {
+  while (!locker.AwaitWithTimeout([this]() REQUIRES(mutex_) {
```
jihoonl: This code style change seems like irrelevant to this commit message. Is it accidental?

SirVer: Probably clang-format. That is fine.

macmason: Correct, I have a particularly aggressive editor configuration.

@jihoonl I'll have the `cartographer_ros` PR ready to go before they merge this one, but I haven't started on it yet.

jihoonl:
@macmason @wohe it looks like submap serialization is missing. Will submap come along with this PR? Or is it something waiting for contribution?

macmason:
@jihoonl my plan was to leave that for a followup; this PR is already more intrusive than I'd like, so I'd prefer to break it up.

macmason:
@wohe I think I've answered all of your questions and this is ready for a second look. Thanks!

cartographer/mapping/sparse_pose_graph.cc
```diff
@@ -38,24 +36,52 @@ proto::SparsePoseGraph::Constraint::Tag ToProto(
...
-    const auto* trajectory = node.constant_data->trajectory;
-    if (trajectory_ids.emplace(trajectory, grouped_nodes->size()).second) {
-      new_indices->emplace_back(grouped_nodes->size(), 0);
-      grouped_nodes->push_back({node});
-    } else {
-      const int id = trajectory_ids[trajectory];
-      new_indices->emplace_back(id, (*grouped_nodes)[id].size());
-      (*grouped_nodes)[id].push_back(node);
-    }
+  grouped_nodes->resize(trajectory_indices.size());
```
wohe: Revert to `trajectory_ids`? We should consistently name them IDs. The rationale was as follows:
- trajectories are uniquely identified by these numbers, so `trajectory_ids`
- submaps and nodes are not uniquely identified by their index inside of their trajectory, so `submap_index` etc
- submap IDs are then pairs of trajectory ID and submap index in that trajectory

Similarly: Rename `IndexTrajectories()` to `ComputeTrajectoryIds()`?

cartographer/mapping/sparse_pose_graph.cc
```diff
@@ -38,24 +36,52 @@ proto::SparsePoseGraph::Constraint::Tag ToProto(
...
+
+  for (const auto& node : trajectory_nodes) {
+    auto index_iterator =
+        trajectory_indices.find(node.constant_data->trajectory);
+    CHECK(index_iterator != trajectory_indices.end());
+    const int index = index_iterator->second;
```
wohe: `unordered_map::at()` throws if the key is not found, so simplify to:
```
const int id = trajectory_indices.at(node.constant_data->trajectory);
```

Similar in other places.

cartographer/mapping/sparse_pose_graph.cc
```diff
@@ -38,24 +36,52 @@ proto::SparsePoseGraph::Constraint::Tag ToProto(
...
+    (*grouped_nodes)[index].push_back(node);
+  }
+}
+
+// TODO(macmason): This function is very nearly a copy of GroupTrajectoryNodes.
+// Consider refactoring them to share an implementation.
+void GroupSubmapStates(
+    const std::vector<SparsePoseGraph::SubmapState>& submap_states,
+    const std::unordered_map<const Submaps*, int>& trajectory_indices,
+    std::vector<std::vector<SparsePoseGraph::SubmapState>>* grouped_submaps,
```
wohe: I think `grouped_submaps` is unused and should thus be removed. Replace by a local variable `std::vector<int>` to just count how many submaps have been added to a particular trajectory, i.e. what the next index would be?

This change could also be done for the nodes variant of the function, but I'm not feeling strongly there since it is used right now, although that use could be removed if we add all trajectories to the proto first and rearrange the nested loops to a single loop using `new_indices` to find the place where to add each node.

cartographer/mapping/sparse_pose_graph.cc
```diff
@@ -81,9 +107,23 @@ proto::SparsePoseGraphOptions CreateSparsePoseGraphOptions(
...
   std::vector<std::vector<TrajectoryNode>> grouped_nodes;
-  std::vector<std::pair<int, int>> new_indices;
-  GroupTrajectoryNodes(GetTrajectoryNodes(), &grouped_nodes, &new_indices);
+  std::vector<std::pair<int, int>> grouped_node_indices;
+  GroupTrajectoryNodes(trajectory_nodes, trajectory_indices, &grouped_nodes,
+                       &grouped_node_indices);
+
+  std::vector<std::vector<SparsePoseGraph::SubmapState>> grouped_submaps;
+  std::vector<std::pair<int, int>> grouped_submap_indices;
+  GroupSubmapStates(GetSubmapStates(), trajectory_indices, &grouped_submaps,
```
wohe: Did you consider improving this by making `GroupSubmapStates()` part of the interface instead of `GetSubmapStates()`? I think that would make sense. While it would mean that there are then three very similar implementations (the nodes variant, and a submap variant for each of 2D and 3D), it also removes quite a bit of code and would not expose internals as much.

cartographer/mapping/sparse_pose_graph.cc
macmason: Done.

cartographer/mapping/sparse_pose_graph.cc
macmason: I hadn't considered it, but I like the idea. I'd also like to land this PR; I've added a TODO in the header file, and I can file an issue if you'd like.

cartographer/mapping/sparse_pose_graph.cc
macmason: Oh, much nicer. Done.

cartographer/mapping/sparse_pose_graph.cc
macmason: Done.

macmason:
Updates pushed. (Sorry for the spam; I responded in batch using the "review" concept, and I can't tell if that messaged you or not.)

macmason:
See also: `cartographer_ros` PR 303.

PR #234, opened 2017-04-20
Submaps To/Fromproto

jihoonl:
Related to #197 

This PR is to support serialization and deserialization of submaps. It is to show the progress and to discuss in open space. @macmason and @wohe. Could you guys check if it is the right way to do it? CLA will be signed during the review.

### DIFF
--- a/cartographer/mapping/proto/submaps.proto
+++ b/cartographer/mapping/proto/submaps.proto
@@ -1,0 +1,23 @@
+// Copyright 2016 The Cartographer Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+syntax = "proto2";
+
+import "cartographer/transform/proto/transform.proto";
+
+package cartographer.mapping.proto;
+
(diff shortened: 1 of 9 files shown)


wohe:
Doing it this way will not scale unfortunately. The reason is that if all data gets put into a single protobuf, we will need to keep this protobuf in memory. This increases peak memory consumption a lot, I would assume by 2x or worse. A better way would be to split the data into many smaller protobufs.

I would suggest the following approach. A protobuf message gets defined that has a `oneof` of all the different types we need to store, i.e. the SparsePoseGraph proto and separately each grid and sensor data. We then stream this data to a file which contains a sequence of these messages. At every point in time we would only have a single protobuf in memory as it is encoded, compressed, and written out to the file.

I could not find any implementation we could simply depend on, but we can follow the idea from https://github.com/google/or-tools/blob/master/src/base/recordio.h. Only supporting compressed data, e.g. using the FastGzipString() function from `port.h` should be sufficient. If it turns out we're unsatisfied with performance or compression ratio, we can investigate more. We'd also only need support for our proto message, and can directly use std::ostream/std::istream.

I would suggest adding code to read/write a sequence of protos to a file in a separate PR. @macmason: Any additional thoughts? @jihoonl: Do you want to prepare this PR?

macmason:
I agree with @wohe with one caveat:

The or-tools "RecordIO" isn't the "real" RecordIO because it doesn't do any error checking. 

[TensorFlow has something better defined](https://www.tensorflow.org/api_guides/python/python_io#tfrecords_format_details), although porting it may be far more effort than it's worth; depending on TF would be crazy, so you'd probably have to basically reimplement the thing.

jihoonl:
Creating a sequence of protobuf message sounds better and scalable approach.

I won't be able to work on this immediately because I am going for long holidays this week. So, it will actually be nice if someone else can develop this feature. I will see what I can do as a next work when I am back from holidays.

Closing this PR.

Issue #177, opened 2017-01-06
Build failure on Debian testing


nxdefiant:
Building cartographer fails on Debian testing. Googletest version is 1.8.0-6 (Debian)

Build output is as follows:

```
==> make cmake_check_build_system in '/opt/ros/ros_catkin_ws/build_isolated/cartographer/install'
-- Build type: Release
-- Found PythonInterp: /usr/bin/python (found version "2.7.13") 
CMake Error at /usr/src/googletest/googletest/cmake/internal_utils.cmake:149 (add_library):
  add_library cannot create target "gmock" because another target with the
  same name already exists.  The existing target is a static library created
  in source directory "/usr/src/googletest/googlemock".  See documentation
  for policy CMP0002 for more details.
Call Stack (most recent call first):
  /usr/src/googletest/googletest/cmake/internal_utils.cmake:172 (cxx_library_with_type)
  /usr/src/gmock/CMakeLists.txt:84 (cxx_library)


CMake Error at /usr/src/googletest/googletest/cmake/internal_utils.cmake:149 (add_library):
  add_library cannot create target "gmock_main" because another target with
  the same name already exists.  The existing target is a static library
  created in source directory "/usr/src/googletest/googlemock".  See
  documentation for policy CMP0002 for more details.
Call Stack (most recent call first):
  /usr/src/googletest/googletest/cmake/internal_utils.cmake:172 (cxx_library_with_type)
  /usr/src/gmock/CMakeLists.txt:89 (cxx_library)


CMake Error at /usr/src/gmock/CMakeLists.txt:106 (install):
  install TARGETS given target "gmock" which does not exist in this
  directory.


-- Found GMock: gmock_main;-lpthread  
-- Boost version: 1.62.0
-- Found the following Boost libraries:
--   iostreams
--   regex
-- Found required Ceres dependency: Eigen version 3.3.0 in /usr/include/eigen3
-- Found required Ceres dependency: glog
-- Found Ceres version: 1.12.0 installed in: /usr/local with components: [LAPACK, SchurSpecializations, OpenMP]
-- Configuring incomplete, errors occurred!
See also "/opt/ros/ros_catkin_ws/build_isolated/cartographer/install/CMakeFiles/CMakeOutput.log".
See also "/opt/ros/ros_catkin_ws/build_isolated/cartographer/install/CMakeFiles/CMakeError.log".
Makefile:3786: die Regel f?r Ziel ?cmake_check_build_system? scheiterte
[...]

PR #236, opened 2017-04-24
Support new Debian version of gmock

jspricke:
Debian moved gmock to /usr/src/googletest/googlemock, according to the
upstream move into the googletest project. This patch adds explicit
search patches for the old and new environment.

Also, it searches for gmock.cc instead of CMakeLists.txt, as the later
is too generic.


SirVer:
@jspricke This breaks building on trusty - see travis. Could you have another look?

jspricke:
Looks like you don't have libgtest-dev intalled on trusty.

damonkohler:
In that case, you should be able to modify the Dockerfiles to fix the Travis build.

jspricke:
Looks like the gtest in google-mock is newer then the one in libgtest-dev. Will try to find a new solution later.

SirVer:
@jspricke I rebased your work on top of master and added travis configuration so that debian is not regressing in the future. Could you please sign the CLA so that we can get this merged?

This fixes https://github.com/googlecartographer/cartographer/issues/177 as well.

jspricke:
I think this should be signed already by Magazino, or do I need to do anything @SirVer?

SirVer:
dunno. @damonkohler probably knows.

jihoonl:
I have added @jspricke's email(github@jochen.sprickerhof.de) in Magazino CLA about a week ago. Does it only work with `magazino.eu` email?

SirVer:
Probably. This is copied from an internal help text about company CLAs:

> As part of signing the CLA, the company designates a Google Group that identifies who at their company is authorized to contribute to Google projects. If someone is trying to contribute to your project, but they are not in their company's authorized contributor group, they'll need to contact whoever administers the group for their company.

SirVer:
@jihoonl Actually, now rereading our internals maybe I misunderstood CLABot. Probably it was happy because it saw the correct email address already - I only looked into the database for individual contributor. 

So in fact this can be merged now. Sorry for the noise.

.travis.yml
```diff
@@ -22,12 +22,13 @@ cache:
     - /home/travis/docker/
 
 env:
-  - UBUNTU_RELEASE=trusty DOCKER_CACHE_FILE=/home/travis/docker/trusty-cache.tar.gz
-  - UBUNTU_RELEASE=xenial DOCKER_CACHE_FILE=/home/travis/docker/xenial-cache.tar.gz
+  - LSB_RELEASE=trusty DOCKER_CACHE_FILE=/home/travis/docker/trusty-cache.tar.gz
+  - LSB_RELEASE=xenial DOCKER_CACHE_FILE=/home/travis/docker/xenial-cache.tar.gz
+  - LSB_RELEASE=stretch DOCKER_CACHE_FILE=/home/travis/docker/stretch-cache.tar.gz
```
wohe: Should we also add jessie? Assuming that it works of course.

PR #241, opened 2017-04-28
Add support for files containing multiple protos.

wohe:
This adds code for reading and writing files that contain
sequences of compressed protocol buffer messages.

### DIFF
--- a/cartographer/io/recordio.cc
+++ b/cartographer/io/recordio.cc
@@ -1,0 +1,85 @@
+/*
+ * Copyright 2016 The Cartographer Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "cartographer/io/recordio.h"
+
+namespace cartographer {
+namespace io {
+
+namespace {
+
+const size_t kMagic = 0x7b1d1f7b5bf501db;
+
+void WriteSizeAsLittleEndian(size_t size, std::ostream* out) {
+  for (int i = 0; i != 8; ++i) {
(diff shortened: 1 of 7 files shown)


wohe:
This is related to #197 and #234.

cartographer/io/recordio.cc
```diff
@@ -0,0 +1,85 @@
...
+ */
+
+#include "cartographer/io/recordio.h"
+
+namespace cartographer {
+namespace io {
+
+namespace {
+
+const size_t kMagic = 0x7b1d1f7b5bf501db;
```
SirVer: comment

cartographer/io/recordio.cc
```diff
@@ -0,0 +1,85 @@
...
+
+}  // namespace
+
+RecordWriter::RecordWriter(const string& filename)
+    : out_(filename, std::ios::out | std::ios::binary) {}
+
+RecordWriter::~RecordWriter() {}
+
+void RecordWriter::Write(const string& uncompressed_data) {
+  WriteSizeAsLittleEndian(kMagic, &out_);
```
SirVer: As discussed, probably enough to have this once at the beginning of the file.

cartographer/io/recordio.cc
```diff
@@ -0,0 +1,85 @@
...
+  }
+  size_t compressed_size;
+  if (!ReadSizeAsLittleEndian(&in_, &compressed_size)) {
+    return false;
+  }
+  string compressed_data(compressed_size, '\0');
+  if (!in_.read(&compressed_data.front(), compressed_size)) {
+    return false;
+  }
+  common::FastGunzipString(compressed_data, decompressed_data);
```
SirVer: What happens if the data in the file is bogus? This should return false then, not crash.

cartographer/io/recordio.h
```diff
@@ -0,0 +1,78 @@
...
+  RecordWriter& operator=(const RecordWriter&) = delete;
+
+  template <typename MessageType>
+  void WriteProto(const MessageType& proto) {
+    string uncompressed_data;
+    proto.SerializeToString(&uncompressed_data);
+    Write(uncompressed_data);
+  }
+
+  bool Close();
```
SirVer: Is this called on destruction?

cartographer/io/recordio.cc
wohe: Unfortunately it right now silently produces wrong output. I think this should be fixed in the future by introducing a class to properly de/compress a stream to/from LZ4 or so, instead of gzipping individual messages.

PR #243, opened 2017-05-02
Some code documentation

Mofef:
When reading through the code a few weeks ago I added comments to parts that took me a long time to understand. I hope these will help others as well.

Obviously I also hope to learn from the reviewer if my assumptions are right! :)


cartographer/mapping_2d/local_trajectory_builder.cc
```diff
@@ -228,15 +228,18 @@ LocalTrajectoryBuilder::AddHorizontalRangeData(
     return nullptr;
   }
 
+  // indices of submaps_ where new scans are inserted are decided here...
```
wohe: Style: Use full sentences, e.g. combine the three comments to one comment here:
```
  // Querying matching and insertion indices must be done here before calling
  // InsertRangeData() since the queried values are valid for next insertion.
```

cartographer/mapping_2d/sparse_pose_graph/constraint_builder.cc
```diff
@@ -60,6 +60,16 @@ ConstraintBuilder::~ConstraintBuilder() {
   CHECK(when_done_ == nullptr);
 }
 
+/**
```
wohe: Function comments belong to the header file. Add additional information there?

cartographer/mapping_2d/sparse_pose_graph/constraint_builder.cc
```diff
@@ -60,6 +60,16 @@ ConstraintBuilder::~ConstraintBuilder() {
...
+ * In the submap search for the best fit of the scans point_cloud in the vicinity of initial_relative_pose.
+ * If the match score is better than the threshold defined in the options add a constraint.
+ *
+ * @param submap_index
```
wohe: Please remove redundant comments.

cartographer/mapping_2d/sparse_pose_graph/constraint_builder.cc
```diff
@@ -184,14 +225,16 @@ void ConstraintBuilder::ComputeConstraint(
   const sensor::PointCloud filtered_point_cloud =
       adaptive_voxel_filter_.Filter(*point_cloud);
 
-  // The 'constraint_transform' (i <- j) is computed from:
-  // - a 'filtered_point_cloud' in j,
-  // - the initial guess 'initial_pose' for (map <- j),
-  // - the result 'pose_estimate' of Match() (map <- j).
-  // - the ComputeSubmapPose() (map <- i)
+  // The 'constraint_transform' (submap <- scan) is computed from:
```
wohe: These were named `i` and `j` to follow `mapping::SparsePoseGraph::Constraint`. Maybe leave both, i.e. `submap i <- scan j`?

cartographer/mapping_2d/sparse_pose_graph/constraint_builder.cc
```diff
@@ -184,14 +225,16 @@ void ConstraintBuilder::ComputeConstraint(
...
+  // - a 'filtered_point_cloud' in scan,
+  // - the initial guess 'initial_pose' for (map <- scan),
+  // - the result 'pose_estimate' of Match() (map <- scan).
+  // - the ComputeSubmapPose() (map <- submap)
   float score = 0.;
   transform::Rigid2d pose_estimate = transform::Rigid2d::Identity();
 
+  // Compute pose_estimate in three stages: Fast estimate, pruning, refinement
```
wohe: Style: Variable names in single quotes, i.e. `'pose_estimate'`

wohe: nit: lower-case `fast` and period after `refinement`.

cartographer/mapping_2d/sparse_pose_graph/constraint_builder.cc
```diff
@@ -184,14 +225,16 @@ void ConstraintBuilder::ComputeConstraint(
...
+  // First estimate using fast correlative scan matcher; return if score is too low.
```
wohe: nit: `;` -> `and`
Also: Did you run clang-format? Lines should be up to 80 characters usually.

cartographer/mapping_2d/sparse_pose_graph.cc
```diff
@@ -243,6 +251,7 @@ void SparsePoseGraph::ComputeConstraintsForScan(
     }
   }
 
+
```
wohe: Please remove extra empty line.

cartographer/mapping_2d/sparse_pose_graph.cc
```diff
@@ -216,6 +222,8 @@ void SparsePoseGraph::ComputeConstraintsForScan(
   for (const mapping::Submap* submap : insertion_submaps) {
     const int submap_index = GetSubmapIndex(submap);
     CHECK(!submap_states_[submap_index].finished);
+    // however probability_grid might be final, so the last ray cast is drawn,
```
wohe: I'm not sure I understand this comment. This code adds constraints between a submap and each of scans that were added to it.

cartographer/mapping_2d/sparse_pose_graph.cc
```diff
@@ -106,6 +106,8 @@ void SparsePoseGraph::AddScan(
   });
   trajectory_connectivity_.Add(submaps);
 
+  // if the last (i.e. latest) insertion_submap is not in the index
+  // it means we have started a new submap...
```
wohe: Rather than repeating the condition, move this comment inside the if. This way, comments and code cannot diverge as easily. For example:
```
    // We have started a new submap.
```

cartographer/mapping_2d/sparse_pose_graph.cc
```diff
@@ -114,10 +116,14 @@ void SparsePoseGraph::AddScan(
     submap_states_.back().trajectory = submaps;
     CHECK_EQ(submap_states_.size(), submap_indices_.size());
   }
+  // ... One call before the first of insertion_submaps was finished
```
wohe: I'm not sure I understand this comment. We check here, whether this was the last insertion into the submap. This is the case when `finished_probability_grid` is non-null.

cartographer/mapping_2d/sparse_pose_graph.cc
```diff
@@ -114,10 +116,14 @@ void SparsePoseGraph::AddScan(
...
   const mapping::Submap* const finished_submap =
       insertion_submaps.front()->finished_probability_grid != nullptr
           ? insertion_submaps.front()
           : nullptr;
+  // by providing finished_submap (!=nullptr) to ComputeConstraintsForScan
```
wohe: Seems this information should be moved to the comment for `ComputeConstraintsForScan()` in the header?

cartographer/mapping_2d/sparse_pose_graph/constraint_builder.cc
Mofef: Oh, good to know. :) to be honest I'm pretty new to C/C++ so i didn't know about this convention. Now that I see the comments there mine seam to be mostly redundant.
But maybe we can convert the existing ones to doxygen, so an IDE will see them?

cartographer/mapping_2d/local_trajectory_builder.cc
Mofef: iirc the fact I wanted to point out, was that `InsertRangeData` might lead to another `submap` being created. This leads to the fact that inside `InsertRangeData` the insertion indices are not necessarily "the last" and "the one before the last". At some point that was surprising for me. 
I can still rephrase it if you prefer.

cartographer/mapping_2d/sparse_pose_graph.cc
Mofef: To me it was weird, that there can be a submap, that scans are being inserted to, which is not even in the index yet.
How about 
```
// SPG learns about new submaps from unseen indices in 'insertion_submaps'
```

cartographer/mapping_2d/sparse_pose_graph.cc
Mofef: horrible sentence, I aggree. Rephrase:
"The last time `AddScan` was called [related to the last comment] this happend:
The first of `insertion_submaps` was finished, so [continues in the next comment]"

Do you think these continued comments are bad in general? I don't have a strong opinion on this. It just seamed handy to visually encode information about the number of lines this comment refers to.

cartographer/mapping_2d/sparse_pose_graph.cc
Mofef: Actually I think `ComputeConstraintsForScan` should not call `ComputeConstraintsForOldScans`. (To make my point  more clear: Otherwise the first one should be called `ComputeConstraintsForThisScanAndOldScansIfSubmapFinished`. This might sound sarcastic, but I honestly just meant to describe the problem I had with understanding the code :) ) I'm not sure if there is a more elegant solution to this, so for now I figured pointing this out shortly before `ComputeConstraintsForScan` is used might be sufficient.

cartographer/mapping_2d/sparse_pose_graph.cc
Mofef: Again just a remark, where my expectation from reading the code turned out wrong after a closer look.
For some reason I don't remember I wanted to point out, that the `CHECK` above does not imply that `submap.probability_grid` is still `Null`. So a not yet finished Submap might be in the process of finishing, where just the constraints to this last scan are still missing.

cartographer/mapping_2d/sparse_pose_graph.cc
Mofef: Oh, sure :)

Mofef:
Thanks for the review! It was quite hard to remember why things where confusing for me when I started to read through the code... :)

cartographer/mapping_2d/sparse_pose_graph/constraint_builder.cc
SirVer: This is not generallicaly a c++ convention (though it always makes sense IMHO). Cartographer follows the [Google c++ style guide](https://google.github.io/styleguide/cppguide.html) which is well written and contains good explanations for the conventions.

cartographer/mapping_2d/sparse_pose_graph.cc
wohe: Here we figure out whether a submap was finished. In practice this might happen at a similar time like growing `submap_states_` (the code above), but is conceptually two separate concerns. I would suggest not adding comments that tie these concerns together. This would encode the current implementation details in comments.

cartographer/mapping_2d/sparse_pose_graph.cc
Mofef: I thought about it again.
Maybe we can remove the conditional assignment to `finished_submap` and instead conditionally add `ComputeConstraintsForOldScans` as a work item based on the same condition...?

So I suggest to just revert this comment for now and open a new PR with those changes. (Exceptyou see a problem with this upfront)

Mofef:
Hey, sorry for the delay I was on vacation the last 3 weeks.
I hope I didn't overlook anything

Mofef:
I just noticed that i messed sth. up when resolving merge conflicts :/
I will take care of it on monday

cartographer/mapping_2d/local_trajectory_builder.cc
```diff
@@ -176,6 +176,8 @@ LocalTrajectoryBuilder::AddHorizontalRangeData(
     return nullptr;
   }
 
+  // Querying matching and insertion indices must be done here before calling
```
wohe: Add this comment also in the 3D case, i.e. `mapping_3d/*_local_trajectory_builder.cc`? Similar in other places: keep 3D in sync.

cartographer/mapping_2d/sparse_pose_graph.cc
```diff
@@ -233,6 +235,8 @@ void SparsePoseGraph::ComputeConstraintsForScan(
       matching_id.trajectory_id, scan_data->time, pose, optimized_pose);
   for (const mapping::Submap* submap : insertion_submaps) {
     const mapping::SubmapId submap_id = GetSubmapId(submap);
+    // If this was the last scan added to the 'submap' it will be marked as
```
wohe:
```
    // Even if this was the last scan added to 'submap', the submap will only be
    // marked as finished in 'submap_data_' further below.
```
?

cartographer/mapping_2d/sparse_pose_graph/constraint_builder.cc
wohe: Not fully clear to me, so maybe it is better to write out the stages. Like this:
```
  // Compute 'pose_estimate' in three stages:
  // 1. Fast estimate using the fast correlative scan matcher.
  // 2. Prune if the score is too low.
  // 3. Refine.
```
?

Mofef: very nice! 
done.

cartographer/mapping_3d/sparse_pose_graph/constraint_builder.cc
```diff
@@ -179,12 +179,16 @@ void ConstraintBuilder::ComputeConstraint(
   const sensor::PointCloud filtered_point_cloud =
       adaptive_voxel_filter_.Filter(point_cloud);
 
-  // The 'constraint_transform' (submap 'i' <- scan 'j') is computed from the
-  // initial guess 'initial_pose' for (submap 'i' <- scan 'j') and a
-  // 'filtered_point_cloud' in 'j'.
+  // The 'constraint_transform' (submap i <- scan j) is computed from:
```
Mofef: I made this also more similar to the formulation in 2D. However please double check since i'm not very familiar with the 3D case.

Issue #244, opened 2017-05-02
Support tracking external observations (Landmarks) as constraints


SirVer:
Users expressed interest to add constraints to visual features, GPS tracks, or bluetooth beacons as constraint for the SLAM. This is the tracking issue for discussion and implementation of this feature.

ahundt:
Here is an example of a mobile base with an AR tag on the gripper, as I had mentioned on today's google hangout:

<img src="https://user-images.githubusercontent.com/55744/26938959-ccfc8ede-4c43-11e7-836b-a4a97167e1c4.jpg" height="256">

In this case the landmark may move relative to both the sensor and world. Both sensors and grippers could be on the robot, the gripper, or the wall. Furthermore, with an AR tag on the base in addition to the gripper poses based on individual joint angles could be estimated and incorporated to refine the estimate. 

I understand this case may be out of scope but even if it cannot be incorporated, it could still useful to consider that landmarks such as AR tags may move in known manner relative to both the world and the robot. Perhaps the API design might then be set up such that it can be more easily extended for this sort of data later on.

jihoonl:
@SirVer mentioned that the currently proposed API in #246 is not sufficient in some use cases during the open house meeting. Could you explain what is missing?

PR #246, opened 2017-05-02
Adds a high level API for pose-relative landmarks

SirVer:
This is an compiling, but non-functional API proposal for #244.

### DIFF
--- a/cartographer/mapping/collated_trajectory_builder.cc
+++ b/cartographer/mapping/collated_trajectory_builder.cc
@@ -97,6 +97,12 @@ void CollatedTrajectoryBuilder::HandleCollatedSensorData(
       wrapped_trajectory_builder_->AddOdometerData(data->time,
                                                    data->odometer_pose);
       return;
+
+    case sensor::Data::Type::kLandmark:
+      wrapped_trajectory_builder_->AddLandmarkData(
+          data->time, data->landmarks.landmarks,
+          data->landmarks.oriented_landmarks);
+      return;
   }
   LOG(FATAL);
 }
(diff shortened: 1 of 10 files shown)


SirVer:
/cc @mofef, @jihoonl This is our suggestion for a public API for pose-relative constraints. This is just the API and no implementation. Do you think you can take this from here or do you require some pointers for the implementation?

Also, this is only tracking -> landmark. Landmark -> landmark would need to go into the SPA interface, but at our meeting it seemed sensible to start with this and see if you can model your problem in this frame of reference. Should this be infeasible, we can revisit the landmark -> landmark constraints again. Agreed?

/cc @macmason Has TRI an opinion on this?

macmason:
LGTM.

Mofef:
Thanks! To us this looks sufficient for now. I didn't find time to think about the detailed implementation yet, but it doesn't seam to be a very complex task. I will get back to you if I run into troubles.

jihoonl:
Looks good to me too.

SirVer:
This has bit rotted quite a bit and there is some agreement that this design is not sufficient. Closing for now as we have no immediate plans of implementing this.

If somebody wants to see this happen, please ping here or on the mailinglist again and we can start putting together a better design.

timethy:
I'm very interested in tracking external landmarks (say, AR tags) to enhance submap construction but also for loop-closure purposes and think this would be of use to others, as well.
I'm aware of the work on fixed-frame constraints (e.g. GPS, VICON), however they do not help in tracking (initially unknown) identificable landmarks.
Maybe @SirVer could briefly explain me his ideas on how to implement it - or provide pointers.
I'm happy to contribute.

SirVer:
Depending on what you want to achieve in detail, you might be good to go. Some tracking of map-fixed constraints has recently be added by @zjwoody. See the commits here: https://github.com/googlecartographer/cartographer/commits?author=zjwoody

wohe:
FYI @pifon2a is looking into implementing this already.

timethy:
Thanks for the pointers!

PR #248, opened 2017-05-03
Enable multiple configuration for multi-trajectory

yutakaoka:
Hi,
This is related to the PR I sent in https://github.com/googlecartographer/cartographer_ros/pull/322.
Thanks,


cartographer/mapping/map_builder.cc
```diff
@@ -99,6 +99,30 @@ int MapBuilder::AddTrajectoryBuilder(
   return trajectory_id;
 }
 
+int MapBuilder::AddTrajectoryBuilder(
```
SirVer: Remove the original AddTrajectoryBuilder and take the (to be created) TrajectoryBuilderOptions proto here. `CHECK` that the correct oneof is set here (the one that matches the settings in the original `MapBuilderOptions` and therefore the sparse pose graph).

SirVer:
Let's get this design right before we look at the cartographer_ros pull request. 

Right now the API is confusing - MapBuilder takes a trajectory option in the constructor and in AddTrajectoryBuilder - but there most of the struct is ignored.

My suggestion is to remove `trajectory_builder_{2,3}d_options` from the `MapBuilderOptions` proto and create a new proto `TrajectoryBuilderOptions` that consists of a oneof `{ mapping_2d.LocalTrajectoryBuilderOptions, mapping_3d.LocalTrajectoryBuilderOptions }` and take this in `AddTrajectoryBuilder`.

yutakaoka:
Thank you for your advice. I am just trying the `oneof` feature, but it is failing to compile. The `oneof` seems to be available after protobuf-2.6. We can apt-get only protobuf-2.5 with Ubuntu14.04. Is there any advice?

SirVer:
Unfortunate. I think using a proto with 2 fields, but only one will ever be set is the closest approximation.

cartographer/mapping/map_builder.cc
yutakaoka: I created TrajectoryBuilderOptions proto.

SirVer:
lgtm. Thanks for the contribution!

Issue #250, opened 2017-05-03
Rename LaserFans into PointClouds


SirVer:
Historically, Cartographer dealt mostly with data from lasers, but internally we only deal with 3D point clouds with an origin. The naming in the repositories often still mention LaserFans.

PR #257, opened 2017-05-03
Remove the LaserScan proto.

wohe:


SirVer:
Related to https://github.com/googlecartographer/cartographer/issues/250.

PR #261, opened 2017-05-04
Add interface to get all submap transforms from SPG

Mofef:
This is needed by googlecartographer/cartographer_ros#325.


ojura:
This is now unnecessary since the Cartographer API changed in #269, it is enough to call GetSubmapTransforms(trajectory) for all trajectories, see magazino/cartographer_ros#1.

ojura:
The only thing that needs to be changed in this repo is that SparsePoseGraph::constraints() should be public, i.e. https://github.com/googlecartographer/cartographer/compare/master...ojura:public-constraints. Let me know if you want me to open a pull request for this.

Mofef:
cool thank you :)

PR #263, opened 2017-05-05
Move ConstantData into the SPG

SirVer:
I did not remember the reason why mapping::TrajectoryNode::ConstantData was owned outside of Cartographer. I think it had to do with us creating multiple MapBuilders in the past. However, this use case seems gone, so moving ConstantData into the SparsePoseGraph now actually works and seems save.

This is based on https://github.com/googlecartographer/cartographer/pull/262.


cartographer/mapping/map_builder.h
```diff
@@ -49,8 +48,7 @@ proto::MapBuilderOptions CreateMapBuilderOptions(
 // and a SparsePoseGraph for loop closure.
 class MapBuilder {
  public:
-  MapBuilder(const proto::MapBuilderOptions& options,
-             std::deque<mapping::TrajectoryNode::ConstantData>* constant_data);
+  MapBuilder(const proto::MapBuilderOptions& options);
```
wohe: Remove `#include <deque>`?

Issue #256, opened 2017-05-03
Refactor multi-trajectory support to keep data separate for each trajectory


wohe:
Currently many parts of the code, e.g. `SparsePoseGraph::GetTrajectoryNodes()`, use a vector that contains nodes from all trajectories interleaved. Each node knows its trajectory. `SparsePoseGraph::Constraint` contains a scan index into this flat list of nodes.

What should be done: Instead, nodes should be kept per trajectory. Constraints would then identify a node as a pair of trajectory ID (an integer 0, 1, ...) and a scan index (the index of the scan in its trajectory).

Similarly for submaps: Change the submap index in `SparsePoseGraph::Constraint` to a pair of trajectory ID and submap index inside this trajectory.

SirVer:
I started looking into this - this will be pretty epic...

PR #271, opened 2017-05-10
Use  (trajectory_id, submap_index) as submap IDs in places.

SirVer:
Related to #256.

PAIR=wohe


cartographer/mapping/submaps.h
```diff
@@ -63,6 +63,21 @@ inline uint8 ProbabilityToLogOddsInteger(const float probability) {
   return value;
 }
 
+// NOCOM(#hrapp): document and rename
```
wohe: Add a small comment to `submap_index` and remove this commet? E.g.
```
  int submap_index;  // Zero-based index of the submap per trajectory.
```

PR #281, opened 2017-05-12
Turn submap state into a per trajectory list.

SirVer:
Related to #256.

### DIFF
--- a/cartographer/mapping_3d/sparse_pose_graph.cc
+++ b/cartographer/mapping_3d/sparse_pose_graph.cc
@@ -115,12 +115,12 @@ void SparsePoseGraph::AddScan(
   if (submap_ids_.count(insertion_submaps.back()) == 0) {
     submap_states_.resize(
         std::max<size_t>(submap_states_.size(), trajectory_id + 1));
+    auto& trajectory_submap_states = submap_states_.at(trajectory_id);
     submap_ids_.emplace(
         insertion_submaps.back(),
         mapping::SubmapId{
             trajectory_id,
-            static_cast<int>(submap_states_.at(trajectory_id).size())});
-    auto& trajectory_submap_states = submap_states_.at(trajectory_id);
+            static_cast<int>(trajectory_submap_states.size())});
     trajectory_submap_states.emplace_back();
     trajectory_submap_states.back().submap = insertion_submaps.back();
   }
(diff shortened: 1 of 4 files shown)


cartographer/mapping_3d/sparse_pose_graph.cc
```diff
@@ -112,15 +112,17 @@ void SparsePoseGraph::AddScan(
...
-    CHECK_EQ(submap_states_.size(), submap_indices_.size());
+  if (submap_ids_.count(insertion_submaps.back()) == 0) {
+    submap_states_.resize(
+        std::max<size_t>(submap_states_.size(), trajectory_id + 1));
+    submap_ids_.emplace(
+        insertion_submaps.back(),
+        mapping::SubmapId{
+            trajectory_id,
+            static_cast<int>(submap_states_.at(trajectory_id).size())});
+    auto& trajectory_submap_states = submap_states_.at(trajectory_id);
```
wohe: Move one up, so that it can already be used there?

cartographer/mapping_3d/sparse_pose_graph.cc
```diff
@@ -433,44 +455,35 @@ std::vector<transform::Rigid3d> SparsePoseGraph::ExtrapolateSubmapTransforms(
...
-  // Extrapolate to the remaining submaps.
-  for (; flat_index != submap_states_.size(); ++flat_index) {
-    const auto& state = submap_states_[flat_index];
-    if (state.id.trajectory_id != trajectory_id) {
-      continue;
-    }
-    if (result.empty()) {
-      result.push_back(transform::Rigid3d::Identity());
+  std::vector<transform::Rigid3d> result;
+  while (result.size() != submap_states_.at(trajectory_id).size()) {
```
wohe:
```
for (const auto& state : submap_states_.at(trajectory_id)) {
```
?

cartographer/mapping_3d/sparse_pose_graph.cc
```diff
@@ -433,44 +455,35 @@ std::vector<transform::Rigid3d> SparsePoseGraph::ExtrapolateSubmapTransforms(
...
+          submap_transforms.at(trajectory_id).at(result.size()).pose);
     } else {
-      // Accessing local_pose() in Submaps is okay, since the member is const.
-      result.push_back(result.back() *
-                       submap_states_[flat_index_of_result_back]
-                           .submap->local_pose()
-                           .inverse() *
-                       state.submap->local_pose());
+      // Extrapolate to the remaining submaps.
+      if (result.empty()) {
```
wohe: `else if`?

cartographer/mapping_3d/sparse_pose_graph.cc
```diff
@@ -433,44 +455,35 @@ std::vector<transform::Rigid3d> SparsePoseGraph::ExtrapolateSubmapTransforms(
...
-    }
-    if (static_cast<size_t>(trajectory_id) >= submap_transforms.size() ||
-        result.size() >= submap_transforms.at(trajectory_id).size()) {
-      break;
-    }
-    result.push_back(
-        submap_transforms.at(trajectory_id).at(result.size()).pose);
-    flat_index_of_result_back = flat_index;
+  const size_t trajectory_id = trajectory_ids_.at(trajectory);
+  if (trajectory_id >= submap_states_.size()) {
```
wohe: Shouldn't this be impossible given that we extend `trajectory_ids_` and `submap_states_` at the same time?

SirVer: unfortunately, this is possible since trajecotry_ids_ grows also when sensor data is added (for example IMU), but submap_states_ not.

PR #285, opened 2017-05-15
Change progress reporting to not use flat indices.

SirVer:
Related to #256.

PAIR=wohe


SirVer:
As discussed, this works under a flawed assumption and is buggy in its current form.

PR #290, opened 2017-05-16
Sparse pose graph no longer uses Submaps* for trajectory identification.

SirVer:
Related to #256.

PAIR=wohe

### DIFF
--- a/cartographer/mapping_2d/global_trajectory_builder.h
+++ b/cartographer/mapping_2d/global_trajectory_builder.h
@@ -28,6 +28,7 @@ class GlobalTrajectoryBuilder
     : public mapping::GlobalTrajectoryBuilderInterface {
  public:
   GlobalTrajectoryBuilder(const proto::LocalTrajectoryBuilderOptions& options,
+                          int trajectory_id,
                           SparsePoseGraph* sparse_pose_graph);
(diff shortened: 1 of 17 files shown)


cartographer/mapping_2d/global_trajectory_builder.h
```diff
@@ -49,6 +50,7 @@ class GlobalTrajectoryBuilder
 
  private:
   const proto::LocalTrajectoryBuilderOptions options_;
```
wohe: Is `options_` unused? Remove!

cartographer/mapping_3d/global_trajectory_builder.cc
```diff
@@ -55,7 +56,7 @@ void GlobalTrajectoryBuilder::AddRangefinderData(
   sparse_pose_graph_->AddScan(
       insertion_result->time, insertion_result->range_data_in_tracking,
       insertion_result->pose_observation, insertion_result->covariance_estimate,
-      insertion_result->submaps, insertion_result->matching_submap,
+      trajectory_id_, insertion_result->matching_submap,
```
wohe: Same here: Remove `submaps` from the InsertionResult?

PR #296, opened 2017-05-18
Fix 3D SLAM loop closure issues.

wohe:
3D submaps are now oriented approximately gravity aligned.
This is so that accumulating error in the local SLAM frame is
no longer a problem for finding loop closures. It also ensures
that the z search window size is approximately in the gravity
direction.

We now also pass an estimate of gravity orientation when doing
multi-trajectory matches. Otherwise trajectories starting with
different orientation relative to gravity could not be connected.

The gravity alignment is currently derived from the ImuTracker.
It might be possible to further improve on this by using the
latest gravity direction from the optimized poses.

### DIFF
--- a/cartographer/mapping_3d/sparse_pose_graph.cc
+++ b/cartographer/mapping_3d/sparse_pose_graph.cc
@@ -187,6 +187,18 @@ void SparsePoseGraph::ComputeConstraint(const mapping::NodeId& node_id,
   // Only globally match against submaps not in this trajectory.
   if (node_id.trajectory_id != submap_id.trajectory_id &&
       global_localization_samplers_[node_id.trajectory_id]->Pulse()) {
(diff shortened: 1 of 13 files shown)


cartographer/mapping_3d/sparse_pose_graph.cc
```diff
@@ -190,7 +196,8 @@ void SparsePoseGraph::ComputeConstraint(const mapping::NodeId& node_id,
         &trajectory_nodes_.at(node_id.trajectory_id)
              .at(node_id.node_index)
              .constant_data->range_data_3d.returns,
-        submap_nodes, &trajectory_connectivity_);
+        submap_nodes, initial_relative_pose.rotation(),
```
SirVer: I don't get this... Could you add a comment why `initial_relative_pose` is the gravity orientation.

PR #297, opened 2017-05-18
Check Travis cache size

### LINKED ISSUE #292: Travis issues
@wohe told me to get in touch if I run into Travis cache issues again. Have a look here: https://travis-ci.org/larics/cartographer_ros

Let me know if I can do something to help debugging this.

Here's an excerpt from the log:
```
if [ -f ${DOCKER_CACHE_FILE} ]; then
  gunzip -c ${DOCKER_CACHE_FILE} | docker load;
fi
invalid argument
The command "scripts/load_docker_cache.sh" failed and exited with 1 during .
```

ojura:
@damonkohler please have a look if this would be acceptable. If yes, let me know so I can make a similar pull request in cartographer_ros as well.

'docker load' errors out if is called to load an empty file. When
pushing branches rapidly, an empty Travis cache sometimes gets saved,
which subsequently breaks Travis builds.

This should fix #292.

### DIFF
--- a/scripts/load_docker_cache.sh
+++ b/scripts/load_docker_cache.sh
@@ -22,5 +22,10 @@ set -o verbose
 set -o pipefail
 
 if [ -f ${DOCKER_CACHE_FILE} ]; then
-  gunzip -c ${DOCKER_CACHE_FILE} | docker load;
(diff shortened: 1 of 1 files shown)


damonkohler:
I would prefer to fix this in the saving script. That way, we can avoid destroying the cache in the first place. I'll have a PR for that in a moment.

PR #307, opened 2017-05-30
Remove 2D scan matcher covariances.

wohe:
This replaces the scaled covariances derived from the Ceres
scan matcher by directly configurable weights. Using covariances
did not provide the expected benefit, and replacing the scaling
matrix by two values will allow a faster evaluation of the cost
function in the future.

### DIFF
--- a/cartographer/mapping/proto/sparse_pose_graph_options.proto
+++ b/cartographer/mapping/proto/sparse_pose_graph_options.proto
@@ -28,6 +28,11 @@ message SparsePoseGraphOptions {
   optional mapping.sparse_pose_graph.proto.ConstraintBuilderOptions
       constraint_builder_options = 3;
 
+  // Weights used in the optimization problem for non-loop-closure scan matcher
+  // constraints.
+  optional double matcher_translation_weight = 7;
+  optional double matcher_rotation_weight = 8;
+
   // Options for the optimization problem.
   optional mapping.sparse_pose_graph.proto.OptimizationProblemOptions
       optimization_problem_options = 4;
--- a/cartographer/mapping/sparse_pose_graph.cc
+++ b/cartographer/mapping/sparse_pose_graph.cc
@@ -16,7 +16,6 @@
 
 #include "cartographer/mapping/sparse_pose_graph.h"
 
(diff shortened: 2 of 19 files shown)


xlicz:
Hi, I'm just wondering what's the "expected benefit" truly mean? As far as I know, the pose graph optimization problem need covariances and giving a precise covariance always make the optimization problem get a precise result.

wohe:
1. Expected benefit would be increased map quality.
2. Precise covariances are not known, so I'm not sure what you are arguing for here.
3. What do you mean with precise result?

xlicz:
Thanks for your nice repof and sorry a little bit abuse of the word "precise ".
For 2:
We may get a close estimation for the covariance, like Andrea Censi do for ICP covariance estimation in his work "An accurate closed-form estimate of ICP’s covariance".
For 3:
For graph optimization problem, covariance matrixs means how much we trust the constraints. If we can get an closer approximation for covariance, we may get better optimization result.

wohe:
For 2: In realistic situations you might not be able to get a good estimation of covariances. Important situations like a broken 2D assumption (laser scanner is not level and the world does not look similar at different heights, e.g. slanted walls) and noise in the environment (e.g. people walking through the scene) are not taken into account.
For 3: You assume that you have better matrices available to use as weights and that the quality benefit outweighs the negative impact on performance. In my experience with real world data, both assumptions are wrong. If you disagree, please provide convincing real world data.

xlicz:
These points make sense. After some experiments, I found that in most cases, the improvement of covariance estimation is minor, as indeed we may or may not get a good covariance estimation. And even when we do have good covariances, we would not get a level better result over fixed covariances because the fixed weights are more or less right. An exception might be the corridor case, at which covariance estimation can at least tells us we don't have information in corridor direction. The

SirVer:
@xlicz Even in the corridor case we did not see the expected elongation of the covariances. Our experiments have been done years ago though, so if you investigate again, please let us know what you find.

wohe:
To be a bit more precise and if I remember correctly, we did see an elongation in the expected direction, but it was not sufficiently pronounced to be useful.

PR #324, opened 2017-06-09
Add 2D trajectory builder option to accept initial pose estimation

aochiai:
This PR add an option to `mapping_2d::LocalTrajectoryBuilder` to accept initial pose estimation.
User can provide initial pose estimation when they start a new trajectory.

### DIFF
--- a/cartographer/mapping_2d/local_trajectory_builder.cc
+++ b/cartographer/mapping_2d/local_trajectory_builder.cc
@@ -32,7 +32,12 @@ LocalTrajectoryBuilder::LocalTrajectoryBuilder(
       real_time_correlative_scan_matcher_(
           options_.real_time_correlative_scan_matcher_options()),
       ceres_scan_matcher_(options_.ceres_scan_matcher_options()),
-      odometry_state_tracker_(options_.num_odometry_states()) {}
+      odometry_state_tracker_(options_.num_odometry_states()) {
+  if (options_.has_initial_pose_estimate()) {
+    pose_estimate_ = transform::ToRigid3(options_.initial_pose_estimate());
+    last_pose_estimate_.pose = pose_estimate_;
+  }
+}
 
 LocalTrajectoryBuilder::~LocalTrajectoryBuilder() {}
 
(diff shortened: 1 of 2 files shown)


wohe:
Can you elaborate a bit on the aim of this PR?

As it is right now, it seems there is no need to add this to Cartographer. In the single trajectory case, you could just transform the poses outside of Cartographer itself. If there are multiple trajectories, you'd check the connectivity to determine whether you'd want to transform them.

Most importantly, the proposed change does not properly influence results, e.g. like one would want if there is already knowledge about where to start related to an already existing trajectory.

aochiai:
This PR aims multi trajectory cases.
The motivation of this PR is:
1. I'd like to use a map before it is well optimized to navigate robots if I can provide a appropriate initial estimation.
2. Sometimes it's useful to adjust initial coordinate of a entire map by giving initial estimation to the first trajectory. (e.g. To build a map with the coordinate same as former maps but need to change start point)

> Most importantly, the proposed change does not properly influence results, e.g. like one would want if there is already knowledge about where to start related to an already existing trajectory.

I expected this scenario and tested in such situation. It seems to work well in that case.
Can you tell me a little more detail why do you think this doesn't properly work?

wohe:
Can you elaborate more about your use cases.

1. If you are only interested in changing the coordinate system, this could obviously be done on the outside, i.e. why is just transforming the output not sufficient for your use case?

2. This change does not influence loop closures because it does not change `trajectory_connectivity_` in SparsePoseGraph. See also #329.

3. It seems to me the feature you want should be added on the pose graph level, not the local trajectory builder. The reason for that is that conceptually the poses in the local trajectory builder track the relative changes without loop closure. The optimized poses in the pose graph can differ a lot. Your change sets the local pose to an initial estimate in the map frame of the pose graph. Since the pose graph does not know about this it will still treat the initial pose as arbitrary and do global loop closure matches against all other submaps.

SirVer:
@aochiai What is the status of this PR?

wohe:
Closing this for now, please re-open if you disagree.

aochiai:
Thank you for reviewing and comments.
I will revise my PR.

Issue #253, opened 2017-05-03
Add full serialization of Cartographer's state


SirVer:
This can't be a single protobuf, since this would require a ton of memory.

Instead the approach should be to invent a container proto, with a set of [`oneof`](https://developers.google.com/protocol-buffers/docs/proto3#oneof)s for each individual piece of state and streaming it through a [RecordWriter](https://github.com/wohe/cartographer/blob/98fffc0e3b15f7b7669917e10be16d54c1a49d1d/cartographer/io/recordio.h).

brandon-northcutt:
@SirVer Yes. I'm working on it now. I'll make a PR once I have something I think is going in the right direction. Speaking of, what is the preferred workflow for a PR?

SirVer:
> Speaking of, what is the preferred workflow for a PR?

We are not particularly picky. Just open one whenever you want a first look or feel it is ready for merging, then rebase or merge master regularly. Once we merge, we will flatten the whole PR into a single commit on master.

wohe:
@brandon-northcutt Any update? Please let me know if there is anything blocking you I can help with.

wohe:
Friendly ping. @macmason What's the state of this?

macmason:
I believe this is still in @brandon-northcutt 's queue.

Brandon, let me know if that's wrong.

brandon-northcutt:
@wohe @macmason : Sorry for the delay. I'm hoping to get this PR in before the round-table Thursday so we can discuss it.

brandon-northcutt:
Looks like trajectory.proto doesn't include range data. Is it true that each trajectory node has either a 2D range scan or a 3D compressed range scan?

SirVer:
@brandon-northcutt Yes.

jihoonl:
@wohe you mentioned that you want to serialize odometry in [map_builder.cc:157](https://github.com/googlecartographer/cartographer/blob/master/cartographer/mapping/map_builder.cc#L157). Is it necessary? Isn't odometry used only for local trajectory building?

wohe:
Yes to both. Right now it is not included in the pose graph optimization, but it should be.

SirVer:
We serialize submaps and the sparse pose graph now. Still missing is range data, IMU and odometry.

brandon-northcutt:
[...]

PR #334, opened 2017-06-13
Trajectory serialization

brandon-northcutt:
This PR follows a conversation with @macmason about where to serialize range scan data. This location seems appropriate to me given the data resides in the trajectory nodes. We are looking for feedback if this seems appropriate. This is working towards the full serialization #253 issue we have been discussing with @SirVer, @wohe, and @damonkohler.

---
This change is [Reviewable](https://reviewable.io/reviews/googlecartographer/cartographer/334)


brandon-northcutt:
I signed it!

brandon-northcutt:
I signed it!

cartographer/mapping/proto/trajectory.proto
```diff
@@ -17,14 +17,17 @@ syntax = "proto2";
...
 
 option java_outer_classname = "TrajectoryOuterClass";
 
 message Trajectory {
-  // NEXT_ID: 7
+  // NEXT_ID: 9
   message Node {
     optional int64 timestamp = 1;
     // Transform from tracking to map frame.
+    optional sensor.proto.RangeData range_data = 7;//2D Range data
```
wohe: I suggest not to add this data here. The problem is that the `Trajectory` proto is used in `SparsePoseGraph` which contains and would thus need to contain all range data in a single proto.

Rather, we should have a separate proto message that contains a node ID and the range data, and we would store many instances of this message in the proto stream. But let's do this in a follow up PR.

cartographer/sensor/proto/sensor.proto
```diff
@@ -40,3 +40,10 @@ message RangeData {
   optional PointCloud point_cloud = 2;
   optional PointCloud missing_echo_point_cloud = 3;
 }
+
+// Proto representation of ::cartographer::sensor::CompressedRangeData
+message CompressedRangeData {
+  optional transform.proto.Vector3f origin = 1;
+  optional CompressedPointCloud point_cloud = 2;
```
wohe: Rename `point_cloud` -> `returns` and `missing_echo_point_cloud` -> `misses`? You could do the same for `RangeData` to follow the naming of the C++ structs.

cartographer/sensor/proto/sensor.proto
```diff
@@ -40,3 +40,10 @@ message RangeData {
   optional PointCloud point_cloud = 2;
   optional PointCloud missing_echo_point_cloud = 3;
 }
+
+// Proto representation of ::cartographer::sensor::CompressedRangeData
+message CompressedRangeData {
```
wohe: Add corresponding code to `range_data.cc` to convert to and from this proto?

brandon-northcutt:
Review status: 0 of 6 files reviewed at latest revision, 3 unresolved discussions.

---

*[cartographer/sensor/proto/sensor.proto, line 45 at r2](https://reviewable.io:443/reviews/googlecartographer/cartographer/334#-KmcaInieQxb0K1yGlKN-r2-45:-KmwryLb_z7XWHge9NuK:b-896fix) ([raw file](https://github.com/googlecartographer/cartographer/blob/ddffef071e738d4c0fe3ab4493021edf03419193/cartographer/sensor/proto/sensor.proto#L45)):*
<details><summary><i>Previously, wohe (Wolfgang Hess) wrote…</i></summary><blockquote>

Add corresponding code to `range_data.cc` to convert to and from this proto?
</blockquote></details>

Done.

---

*[cartographer/sensor/proto/sensor.proto, line 47 at r2](https://reviewable.io:443/reviews/googlecartographer/cartographer/334#-KmcaInieQxb0K1yGlKN-r2-47:-Kmwrz2Dmcxl2aQyTaMb:b-896fix) ([raw file](https://github.com/googlecartographer/cartographer/blob/ddffef071e738d4c0fe3ab4493021edf03419193/cartographer/sensor/proto/sensor.proto#L47)):*
<details><summary><i>Previously, wohe (Wolfgang Hess) wrote…</i></summary><blockquote>

Rename `point_cloud` -> `returns` and `missing_echo_point_cloud` -> `misses`? You could do the same for `RangeData` to follow the naming of the C++ structs.
</blockquote></details>

Done.

---

*[cartographer/mapping/proto/trajectory.proto, line 29 at r2](https://reviewable.io:443/reviews/googlecartographer/cartographer/334#-KmcaInieQxb0K1yGlKM-r2-29:-Kmm5YKwvy-QZEUHQKZG:b-yjuiuj) ([raw file](https://github.com/googlecartographer/cartographer/blob/ddffef071e738d4c0fe3ab4493021edf03419193/cartographer/mapping/proto/trajectory.proto#L29)):*
<details><summary><i>Previously, wohe (Wolfgang Hess) wrote…</i></summary><blockquote>

I suggest not to add this data here. The problem is that the `Trajectory` proto is used in `SparsePoseGraph` which contains and would thus need to contain all range data in a single proto.

Rather, we should have a separate proto message that contains a node ID and the range data, and we would store many instances of this message in the proto stream. But let's do this in a follow up PR.
</blockquote></details>

Removed the range data from trajectory.proto. I'll continue this discussion in the serialization issue. #253

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/googlecartographer/cartographer/334)*

brandon-northcutt:
Review status: 0 of 6 files reviewed at latest revision, 4 unresolved discussions.

---

*[cartographer/sensor/compressed_point_cloud.h, line 53 at r3](https://reviewable.io:443/reviews/googlecartographer/cartographer/334#-KmxWcEPs5D7cXu35B0n:-KmxWcEQr9w741AYKtNU:b-y2uhel) ([raw file](https://github.com/googlecartographer/cartographer/blob/0fc139fba1bdbc4565720cc7874bf965ba502028/cartographer/sensor/compressed_point_cloud.h#L53)):*
> ```objc
>   proto::CompressedPointCloud ToProto() const;
> 
>   CompressedPointCloud(const std::vector<int32>& point_data, size_t num_points);
> ```

I made this constructor public so that I can use it in compressed_point_cloud.cc's ToCompressedPointCloud() method. Let me know if that is acceptable, or if another approach is better.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/googlecartographer/cartographer/334)*

cartographer/sensor/compressed_point_cloud.cc
```diff
@@ -178,5 +178,17 @@ proto::CompressedPointCloud CompressedPointCloud::ToProto() const {
...
+// Proto to CompressedPointCloud
+CompressedPointCloud ToCompressedPointCloud( const proto::CompressedPointCloud&
+proto) {
+  CompressedPointCloud compressed_point_colud;
+  const size_t num_points = proto.point_data_size();
+  std::vector<int32> point_data;
+  for (int i = 0; i != num_points; ++i) {
+    point_data.emplace_back(proto.point_data(i));
+  }
+  return CompressedPointCloud(point_data, num_points);
```
wohe: Is this constructor unused otherwise? Remove and make this a constructor taking the proto?

cartographer/sensor/compressed_point_cloud.cc
```diff
@@ -178,5 +178,17 @@ proto::CompressedPointCloud CompressedPointCloud::ToProto() const {
   return result;
 }
 
+// Proto to CompressedPointCloud
+CompressedPointCloud ToCompressedPointCloud( const proto::CompressedPointCloud&
+proto) {
+  CompressedPointCloud compressed_point_colud;
+  const size_t num_points = proto.point_data_size();
```
brandon-northcutt: Tests added. I went ahead and decompressed the points to make sure what went in is what came out. Does that seem acceptable?

cartographer/sensor/compressed_point_cloud.cc
```diff
@@ -149,6 +149,17 @@ CompressedPointCloud::CompressedPointCloud(const std::vector<int32>& point_data,
                                            size_t num_points)
```
wohe: I think this constructor is not used anywhere. Remove here and in the header?

cartographer/sensor/compressed_point_cloud.h
```diff
@@ -50,13 +50,18 @@ class CompressedPointCloud {
 
   proto::CompressedPointCloud ToProto() const;
 
+  CompressedPointCloud(const proto::CompressedPointCloud& proto);
```
wohe: Style: Constructors come first, i.e. move write below the `explicit CompressedPointCloud(`. Also: Make `explicit` since it takes a single argument.

cartographer/sensor/compressed_point_cloud.h
```diff
@@ -50,13 +50,18 @@ class CompressedPointCloud {
...
+
  private:
   CompressedPointCloud(const std::vector<int32>& point_data, size_t num_points);
 
   std::vector<int32> point_data_;
   size_t num_points_;
 };
 
+CompressedPointCloud ToCompressedPointCloud(
```
wohe: Remove, since it is no longer needed? Call sites can use the constructor directly.

cartographer/sensor/range_data.cc
```diff
@@ -25,17 +25,33 @@ namespace sensor {
...
-  auto range_data = RangeData{
-      transform::ToEigen(proto.origin()), ToPointCloud(proto.point_cloud()),
-      ToPointCloud(proto.missing_echo_point_cloud()),
+  return RangeData{
+      transform::ToEigen(proto.origin()), ToPointCloud(proto.returns()),
+      ToPointCloud(proto.misses()),
+  };
+}
+
+proto::CompressedRangeData ToProto(
```
wohe: Move next to `Compress()` to match the ordering in the header?

cartographer/sensor/point_cloud.cc
```diff
@@ -63,5 +63,6 @@ PointCloud ToPointCloud(const proto::PointCloud& proto) {
   return point_cloud;
 }
 
+
```
wohe: Remove extra empty line?

cartographer/sensor/compressed_range_data_proto_test.cc
```diff
@@ -0,0 +1,71 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "cartographer/sensor/compressed_point_cloud.h"
+#include "cartographer/sensor/point_cloud.h"
+#include "cartographer/sensor/range_data.h"
+
+#include <errno.h>
```
wohe: Remove unused headers?

cartographer/sensor/compressed_range_data_proto_test.cc
```diff
@@ -0,0 +1,71 @@
...
+
+#include "cartographer/common/port.h"
+#include "cartographer/sensor/proto/sensor.pb.h"
+#include "gtest/gtest.h"
+
+namespace cartographer {
+namespace sensor {
+namespace {
+
+TEST(CompressedRangeDataProtoTest, ToProto) {
```
wohe: If this tests code in `range_data.cc` it should be in the file `range_data_test.cc`. We already test compression there. I'm absolutely not opposed to more tests, but I'd be fine with having a unit test for the proto conversion of CompressedPointCloud for this PR.

cartographer/sensor/compressed_point_cloud.cc
```diff
@@ -149,6 +149,17 @@ CompressedPointCloud::CompressedPointCloud(const std::vector<int32>& point_data,
...
     : point_data_(point_data), num_points_(num_points) {}
 
+CompressedPointCloud::CompressedPointCloud(
+    const proto::CompressedPointCloud& proto) {
+  const size_t data_size = proto.point_data_size();
```
wohe: Move down where it is used?

cartographer/sensor/compressed_point_cloud.cc
```diff
@@ -149,6 +149,17 @@ CompressedPointCloud::CompressedPointCloud(const std::vector<int32>& point_data,
...
+
+  num_points_ = proto.num_points();
+  point_data_.reserve(proto.point_data_size());
```
wohe: Reuse `data_size` here?

cartographer/sensor/compressed_point_cloud.cc
```diff
@@ -149,6 +149,17 @@ CompressedPointCloud::CompressedPointCloud(const std::vector<int32>& point_data,
...
+  for (int i = 0; i != data_size; ++i) {
+    point_data_.emplace_back(proto.point_data(i));
+  }
```
wohe: As I said it would be nice to add a `CHECK()` that verifies that the proto contained valid data. Otherwise code might crash/misbehave at hard to debug places. At least we should have a `TODO` here. The point is not so much to check that we correctly serialized, but that we, e.g. did not read malformed data from disk.

cartographer/sensor/compressed_point_cloud.cc
brandon-northcutt: Done.

cartographer/sensor/compressed_point_cloud.cc
brandon-northcutt: Done.

cartographer/sensor/compressed_point_cloud.cc
brandon-northcutt: Done.

cartographer/sensor/compressed_point_cloud.h
brandon-northcutt: Done.

cartographer/sensor/compressed_point_cloud.h
brandon-northcutt: Good point. Done.

cartographer/sensor/range_data.cc
brandon-northcutt: Done.

cartographer/sensor/point_cloud.cc
brandon-northcutt: Done.

cartographer/sensor/compressed_range_data_proto_test.cc
brandon-northcutt: Done.

cartographer/sensor/compressed_range_data_proto_test.cc
brandon-northcutt: Tests added and moved into range_data_test.cc.

cartographer/sensor/compressed_point_cloud.cc
brandon-northcutt: I added a check to at least make sure a required proto member exists.

cartographer/sensor/range_data_test.cc
```diff
@@ -15,45 +15,137 @@
...
-  const std::vector<Eigen::Vector3f> returns = {Eigen::Vector3f(0, 1, 2),
-                                                Eigen::Vector3f(4, 5, 6),
-                                                Eigen::Vector3f(0, 1, 2)};
-  const RangeData range_data = {
-      Eigen::Vector3f(1, 1, 1), returns, {Eigen::Vector3f(7, 8, 9)}};
+class RangeDataTest : public ::testing::Test {
+ protected:
+  RangeDataTest() {
+    origin_ = Eigen::Vector3f(1, 1, 1);
+    returns_.emplace_back(Eigen::Vector3f(0, 1, 2));
```
macmason-tri: `emplace_back` should let you do:

```
returns_.emplace_back(0, 1, 2);
```

cartographer/sensor/range_data_test.cc
```diff
@@ -15,45 +15,137 @@
...
+    misses_.emplace_back(Eigen::Vector3f(7, 8, 9));
+  }
+  void RangeDataToProtoTest() {}
+
+  void CompressedRangeDataToProtoTest() {
+    const CompressedPointCloud compressed_returns(returns_);
+    const CompressedPointCloud compressed_misses(misses_);
+    const CompressedRangeData compressed_range_data = {
+        origin_, compressed_returns, compressed_misses};
```
macmason: This `}` is in a weird place.

cartographer/sensor/range_data_test.cc
brandon-northcutt: Nice! Done.

cartographer/sensor/range_data_test.cc
brandon-northcutt: Done.

cartographer/sensor/compressed_point_cloud.cc
```diff
@@ -145,9 +145,16 @@ CompressedPointCloud::CompressedPointCloud(const PointCloud& point_cloud)
...
-    : point_data_(point_data), num_points_(num_points) {}
+CompressedPointCloud::CompressedPointCloud(
+    const proto::CompressedPointCloud& proto) {
+  CHECK(proto.has_num_points()) << "The compressed point cloud is malformed.";
+  num_points_ = proto.num_points();
+  const size_t data_size = proto.point_data_size();
+  point_data_.reserve(data_size);
+  for (int i = 0; i != data_size; ++i) {
+    point_data_.emplace_back(proto.point_data(i));
+  }
```
wohe: Please add:
```
// TODO(wohe): Verify that 'point_data_' does not contain malformed data.
```

cartographer/sensor/compressed_point_cloud.cc
```diff
@@ -145,9 +145,16 @@ CompressedPointCloud::CompressedPointCloud(const PointCloud& point_cloud)
   CHECK_EQ(num_blocks, 0);
 }
 
-CompressedPointCloud::CompressedPointCloud(const std::vector<int32>& point_data,
-                                           size_t num_points)
-    : point_data_(point_data), num_points_(num_points) {}
+CompressedPointCloud::CompressedPointCloud(
+    const proto::CompressedPointCloud& proto) {
+  CHECK(proto.has_num_points()) << "The compressed point cloud is malformed.";
```
wohe: This is not needed. It is quite common to interpret missing fields in a proto as their default value, i.e. no points in this case.

cartographer/sensor/compressed_point_cloud.cc
```diff
@@ -145,9 +145,16 @@ CompressedPointCloud::CompressedPointCloud(const PointCloud& point_cloud)
...
+  num_points_ = proto.num_points();
+  const size_t data_size = proto.point_data_size();
```
wohe: `size_t` -> `int`, to match the generated protobuf code and the comparison below.

cartographer/sensor/range_data_test.cc
```diff
@@ -15,45 +15,139 @@
  */
 
 #include "cartographer/sensor/range_data.h"
-
-#include <utility>
 #include <vector>
-
+#include <typeinfo>
```
wohe: Unused? Please remove.

cartographer/sensor/range_data_test.cc
```diff
@@ -15,45 +15,139 @@
  */
 
 #include "cartographer/sensor/range_data.h"
```
wohe: Style: Empty line between the related include and C/C++ includes.

cartographer/sensor/range_data_test.cc
```diff
@@ -15,45 +15,139 @@
...
-
-#include <utility>
 #include <vector>
-
+#include <typeinfo>
 #include "gmock/gmock.h"
```
wohe: Style: Empty line between C/C++ include and other includes.

cartographer/sensor/range_data_test.cc
```diff
@@ -15,45 +15,139 @@
...
 
 namespace cartographer {
 namespace sensor {
 namespace {
 
 using ::testing::Contains;
+using ::testing::FloatNear;
```
wohe: Unused? Remove.

cartographer/sensor/range_data_test.cc
```diff
@@ -15,45 +15,139 @@
...
+class RangeDataTest : public ::testing::Test {
+ protected:
+  RangeDataTest() {
+    origin_ = Eigen::Vector3f(1, 1, 1);
+    returns_.emplace_back(0, 1, 2);
+    returns_.emplace_back(4, 5, 6);
+    returns_.emplace_back(0, 1, 2);
+    misses_.emplace_back(7, 8, 9);
+  }
+  void RangeDataToProtoTest() {}
```
wohe: Unused? Remove. Same with `CompressedRangeDataToProtoTest`.

cartographer/sensor/range_data_test.cc
```diff
@@ -15,45 +15,139 @@
...
+  EXPECT_TRUE(actual.misses[0].isApprox(misses_[0], 0.001f));
 
-  // Returns will be reordered, so we compare in an unordered manner.
-  EXPECT_EQ(3, actual.returns.size());
-  EXPECT_THAT(actual.returns,
-              Contains(ApproximatelyEquals(Eigen::Vector3f(0, 1, 2))));
-  EXPECT_THAT(actual.returns,
-              Contains(ApproximatelyEquals(Eigen::Vector3f(4, 5, 6))));
+  // Returns may be reordered, so we compare in an unordered manner.
+  for (uint i = 0; i < returns_.size(); ++i) {
```
wohe: Use range-based for loops?

cartographer/sensor/range_data_test.cc
```diff
@@ -15,45 +15,139 @@
...
+    EXPECT_THAT(actual.returns, Contains(ApproximatelyEquals(returns_[i])));
+  }
+  for (uint i = 0; i < misses_.size(); ++i) {
+    EXPECT_THAT(actual.misses, Contains(ApproximatelyEquals(misses_[i])));
+  }
+}
+
+TEST_F(RangeDataTest, RangeDataToProto) {
+  const auto proto = ToProto(RangeData{origin_, returns_, misses_});
```
wohe: This unit test to me does not increase confidence in the code. How about Converting to a proto and back and check for equality? Similar for the compressed range data: Extract a function to compare unordered and with limited precision and reuse it for the compressed range data test?

cartographer/sensor/compressed_point_cloud.cc
brandon-northcutt: Done.

cartographer/sensor/compressed_point_cloud.cc
brandon-northcutt: Done.

cartographer/sensor/compressed_point_cloud.cc
brandon-northcutt: Done.

cartographer/sensor/range_data_test.cc
brandon-northcutt: Done.

cartographer/sensor/range_data_test.cc
brandon-northcutt: Done.

cartographer/sensor/range_data_test.cc
brandon-northcutt: Done.

cartographer/sensor/range_data_test.cc
brandon-northcutt: Done.

cartographer/sensor/range_data_test.cc
brandon-northcutt: Done.

cartographer/sensor/range_data_test.cc
```diff
@@ -27,33 +27,79 @@ namespace {
...
+        expected.misses[i].isApprox(actual.misses[i], kPrecision));
+  }
 }
 
+TEST_F(RangeDataTest, CompressedRangeDataToAndFromProto) {
+  const auto expected = CompressedRangeData{
+      origin_, CompressedPointCloud(returns_), CompressedPointCloud(misses_)};
+  const auto actual = FromProto(ToProto(expected));
+
+  EXPECT_TRUE(typeid(expected) == typeid(actual));
```
wohe: This seems to repeat the implementation. Tests should increase confidence in the correctness, not duplicate the implementation. Please remove here and elsewhere.

cartographer/sensor/range_data_test.cc
```diff
@@ -27,33 +27,79 @@ namespace {
...
+  EXPECT_TRUE(expected.origin.isApprox(actual.origin, kPrecision));
+
+  EXPECT_EQ(expected.returns.size(), actual.returns.size());
+  EXPECT_TRUE(expected.returns.empty() == actual.returns.empty());
```
wohe: Obviously true since we already checked that the sizes are equal. Remove?

cartographer/sensor/range_data_test.cc
```diff
@@ -27,33 +27,79 @@ namespace {
...
-TEST(RangeDataTest, Compression) {
-  const std::vector<Eigen::Vector3f> returns = {Eigen::Vector3f(0, 1, 2),
-                                                Eigen::Vector3f(4, 5, 6),
-                                                Eigen::Vector3f(0, 1, 2)};
-  const RangeData range_data = {
-      Eigen::Vector3f(1, 1, 1), returns, {Eigen::Vector3f(7, 8, 9)}};
+class RangeDataTest : public ::testing::Test {
+ protected:
+  RangeDataTest() {
+    origin_ = Eigen::Vector3f(1, 1, 1);
```
wohe: Put this in the initializer list?

cartographer/sensor/range_data_test.cc
```diff
@@ -27,33 +27,79 @@ namespace {
...
 using ::testing::PrintToString;
+constexpr float kPrecision = 0.001f;
 
-// Custom matcher for Eigen::Vector3f entries.
+// TODO (brandon-northcutt): place this function in a common location with
+// identical function in compressed_point_cloud_test.cc
 MATCHER_P(ApproximatelyEquals, expected,
           string("is equal to ") + PrintToString(expected)) {
-  return (arg - expected).isZero(0.001f);
+  return (arg - expected).isZero(kPrecision);
```
wohe: Please revert. Let's not make the test more complicated than it needs to be.

cartographer/sensor/range_data_test.cc
```diff
@@ -27,33 +27,79 @@ namespace {
...
+TEST_F(RangeDataTest, RangeDataToAndFromProto) {
+  const auto expected = RangeData{origin_, returns_, misses_};
+  const auto actual = FromProto(ToProto(expected));
+
+  EXPECT_TRUE(typeid(expected) == typeid(actual));
+  EXPECT_TRUE(expected.origin.isApprox(actual.origin, kPrecision));
+
+  for (uint i = 0; i < expected.returns.size(); ++i) {
+    EXPECT_TRUE(
+        expected.returns[i].isApprox(actual.returns[i], kPrecision));
```
wohe: Can't you write this using the `Pointwise` matcher and `ApproximatelyEquals`?

cartographer/sensor/compressed_point_cloud.h
```diff
@@ -48,11 +49,10 @@ class CompressedPointCloud {
   ConstIterator begin() const;
   ConstIterator end() const;
 
+  const std::vector<int32>& point_data() const { return point_data_; };
```
wohe: This leaks an implementation detail. Implement `operator==` instead?

cartographer/sensor/range_data_test.cc
brandon-northcutt: Done.

cartographer/sensor/range_data_test.cc
brandon-northcutt: Done.

cartographer/sensor/range_data_test.cc
brandon-northcutt: Done.

cartographer/sensor/range_data_test.cc
brandon-northcutt: Done.

cartographer/sensor/compressed_point_cloud.h
brandon-northcutt: Done.

cartographer/sensor/range_data_test.cc
brandon-northcutt: I added a matcher to enable `Pointwise()` and was able to remove two `for` loops.

cartographer/sensor/range_data.cc
```diff
@@ -53,6 +52,21 @@ RangeData CropRangeData(const RangeData& range_data, const float min_z,
...
 }
 
+proto::CompressedRangeData ToProto(
+    const CompressedRangeData& compressed_range_data) {
+  proto::CompressedRangeData proto;
+  *proto.mutable_origin() = transform::ToProto(compressed_range_data.origin);
+  *proto.mutable_returns() = compressed_range_data.returns.ToProto();
+  *proto.mutable_misses() = compressed_range_data.misses.ToProto();
+  return proto;
+}
```
wohe: Style: Empty line to separate the two functions.

cartographer/sensor/range_data_test.cc
```diff
@@ -27,33 +26,64 @@ namespace {
 
 using ::testing::Contains;
 using ::testing::PrintToString;
+using ::std::tr1::get;
```
wohe: Please remove.

cartographer/sensor/range_data_test.cc
```diff
@@ -27,33 +26,64 @@ namespace {
...
 
-// Custom matcher for Eigen::Vector3f entries.
-MATCHER_P(ApproximatelyEquals, expected,
-          string("is equal to ") + PrintToString(expected)) {
-  return (arg - expected).isZero(0.001f);
+MATCHER(NearPointwise, std::string(negation ? "Doesn't" : "Does") + " match.") {
+  return get<0>(arg).isApprox(get<1>(arg), 0.001f);
```
wohe: Please use `std::get` instead of `get`.

cartographer/sensor/range_data_test.cc
```diff
@@ -16,7 +16,6 @@
 
 #include "cartographer/sensor/range_data.h"
 
-#include <utility>
 #include <vector>
```
wohe: Please include `<tuple>` where `std::get` is defined.

cartographer/sensor/range_data_test.cc
```diff
@@ -27,33 +26,64 @@ namespace {
...
+}
+
+TEST_F(RangeDataTest, CompressedRangeDataToAndFromProto) {
+  const auto expected = CompressedRangeData{
+      origin_, CompressedPointCloud(returns_), CompressedPointCloud(misses_)};
+  const auto actual = FromProto(ToProto(expected));
+  EXPECT_THAT(expected.origin, Near(actual.origin));
+  EXPECT_EQ(expected.returns, actual.returns);
+  EXPECT_EQ(expected.misses, actual.misses);
+}
```
wohe: Style: Empty line between closing curly brace of the function and closing the unnamed namespace.

cartographer/sensor/range_data.cc
brandon-northcutt: Done.

cartographer/sensor/range_data_test.cc
brandon-northcutt: Done.

cartographer/sensor/range_data_test.cc
brandon-northcutt: Done.

cartographer/sensor/range_data_test.cc
brandon-northcutt: Done.

cartographer/sensor/range_data_test.cc
brandon-northcutt: Thanks for being patient with this review!

PR #335, opened 2017-06-13
Container placeholder to facilitate full serialization to and from disk.

brandon-northcutt:
This is a first stab at a wrapper container to support full serialization to disk (#253) using the many various protos. Note: there isn't much meat here, more an approach with various steps outlined. Does this seem like the right approach @SirVer, @wohe, @damonkohler, @macmason?

---
This change is [Reviewable](https://reviewable.io/reviews/googlecartographer/cartographer/335)

### DIFF
--- a/cartographer/io/map_io.cc
+++ b/cartographer/io/map_io.cc
@@ -1,0 +1,62 @@
+/*
+ * Copyright 2016 The Cartographer Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "cartographer/io/map_io.h"
+
+namespace cartographer {
+namespace io {
+
+//constructor takes a pointer to the map_builder
+cartographer::io::MapIO::MapIO(cartographer::mapping::MapBuilder & map_builder){
+};
+
+void MapIO::write(const string file_name){
+  //TODO(brandon-northcutt) Pause optimization so that submaps, trajectories,
+  //  and constraints are immutable for serialization of state.
+
+  //TODO(brandon-northcutt) Get a file handle to send protobufs to.
+
+  //TODO(brandon-northcutt) Send map_builder_options protobuf to file first, so
+  //  deserialization constructor can create a map builder object as it's first
+  //  action.
+  
+  //TODO(brandon-northcutt) Serialize all other single copy options protobufs
+  //  ceres_scan_matcher_options, range_data_inserter_options,
+  //  sparse_pose_graph_options, real_time_correlative_scan_matcher_options,
+  //  kalman_local_trajectory_builder_options, local_trajectory_builder_options,
+  //  motion_filter_options, optimizing_local_trajectory_builder_options, 
+  //  adaptive_voxel_filter_options,...)
+
+  //TODO(brandon-northcutt) Iterate trajectories { serialize trajectory builder
+  //  options, serialize sensor configuration proto, serialize all nodes in this
+  //  trajectory.
+  
+  //TODO(brandon-northcutt) Serialize trajectory connectivity.
+
+  //TODO(brandon-northcutt) Iterate submaps(serialize submap options, serialize
+  //  probability grids.
+  
+  //TODO(brandon-northcutt) Serialize the sparse pose graph,  (Should the SPG
+  //  proto contain trajectories?)
+};
+
+void MapIO::read(const string file_name){
+  //TODO(brandon-northcutt) Same steps as write but feed protobufs into
+  //  the appropriate constructors.
+};
+
+} // namespace io
+} // namespace cartographer
--- a/cartographer/io/map_io.cc
+++ b/cartographer/io/map_io.cc
@@ -19,44 +19,46 @@
 namespace cartographer {
 namespace io {
 
-//constructor takes a pointer to the map_builder
-cartographer::io::MapIO::MapIO(cartographer::mapping::MapBuilder & map_builder){
-};
+//constructor takes a const reference to the map_builder
+cartographer::io::MapIO::MapIO(const cartographer::mapping::MapBuilder
+    &map_builder){}
 
-void MapIO::write(const string file_name){
-  //TODO(brandon-northcutt) Pause optimization so that submaps, trajectories,
+bool MapIO::Write(const string file_name){
+  // TODO(brandon-northcutt) Pause optimization so that submaps, trajectories,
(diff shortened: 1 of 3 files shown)


brandon-northcutt:
I signed it!

cartographer/io/map_io.h
```diff
@@ -0,0 +1,41 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef CARTOGRAPHER_IO_MAPPING_IO_H_
+#define CARTOGRAPHER_IO_MAPPING_IO_H_
+
+#include "cartographer/mapping/map_builder.h"
+#include "cartographer/io/file_writer.h"//Re-used filestream wrapper.
```
macmason: Extraneous comment here.

cartographer/io/map_io.h
```diff
@@ -0,0 +1,41 @@
...
+
+namespace cartographer {
+namespace io {
+
+// A class to serialize and deserialize the Cartographer state.
+class MapIO {
+public:
+  MapIO(cartographer::mapping::MapBuilder & map_builder);
```
macmason: GSG requires that `map_builder` either be passed const reference (in the case when it's not modified) or by pointer (in the case where it's modified). 

Also, no space before '&', '*', etc: type modifiers cuddle left in Cartographer code.

cartographer/io/map_io.h
```diff
@@ -0,0 +1,41 @@
...
+  MapIO(const MapIO &) = delete;
+  MapIO &operator=(const MapIO &) = delete;
+  //TODO(brandon-northcutt) Write the full Cartographer state to disk.
```
macmason: Space after // in comments (here and lots of other places).

cartographer/io/map_io.h
```diff
@@ -0,0 +1,41 @@
...
+  void write(const std::string file_name);
```
macmason: GSG specifies that "non-trivial" functions `BeNamedLikeThis`, so you want `Write`.

It should also return a `bool` signifying success, and (depending on what the Googlers say) maybe also set the warn_unused_result attribute.

cartographer/io/map_io.h
```diff
@@ -0,0 +1,41 @@
...
+  //TODO(brandon-northcutt) Read the full cartographer state from disk.
+  void read(const std::string file_name);
```
macmason: Ditto above.

cartographer/io/proto/map_io.proto
```diff
@@ -0,0 +1,61 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+syntax = "proto2";
+
+package cartographer.mapping.proto;
```
macmason: Perhaps `cartographer.io.proto`?

cartographer/io/map_io.cc
```diff
@@ -0,0 +1,62 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "cartographer/io/map_io.h"
+
+namespace cartographer {
+namespace io {
+
+//constructor takes a pointer to the map_builder
```
macmason: Comment is wrong.

brandon-northcutt: Done.

brandon-northcutt:
It looks like the Travis CI build fails with the error:

```bash
cartographer/io/proto/map_io.proto:55:3: Expected "required", "optional", or "repeated".
```
This is on the line defining a `oneof` protobuf message and I get no such error message compiling the file using the current version of protobuf. I think that `oneof` is supported by [proto2](https://developers.google.com/protocol-buffers/docs/proto#oneof). Perhaps the protobuf compiler in Travis CI is due for an update? I'm also fine with using another method that does not require `oneof`  [(#253)](https://github.com/googlecartographer/cartographer/issues/253#issue-225918899).

cartographer/io/proto/map_io.proto
```diff
@@ -0,0 +1,61 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+syntax = "proto2";
+
+package cartographer.io.proto;
+
+import "cartographer/common/proto/ceres_solver_options.proto";
```
wohe: Let's only import what is used.

cartographer/io/map_io.cc
```diff
@@ -0,0 +1,64 @@
...
+#include "cartographer/io/map_io.h"
+
+namespace cartographer {
+namespace io {
+
+//constructor takes a const reference to the map_builder
+cartographer::io::MapIO::MapIO(const cartographer::mapping::MapBuilder
+    &map_builder){}
+
+bool MapIO::Write(const string file_name){
```
wohe: Let's make the serialization part of `MapBuilder`, similar to how there is `SparsePoseGraph::ToProto()`, maybe like this:
```
bool MapBuilder::SerializeState(io::ProtoStreamWriter* writer);
```
There is no need to keep additional state across calls and by putting it at the `MapBuilder` we do not have to expose implementation details.

cartographer/io/map_io.cc
```diff
@@ -0,0 +1,64 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "cartographer/io/map_io.h"
+
+namespace cartographer {
+namespace io {
+
+//constructor takes a const reference to the map_builder
```
wohe: Please do not add comments repeating what the code does.

cartographer/io/map_io.cc
```diff
@@ -0,0 +1,64 @@
...
+cartographer::io::MapIO::MapIO(const cartographer::mapping::MapBuilder
+    &map_builder){}
+
+bool MapIO::Write(const string file_name){
+  // TODO(brandon-northcutt) Pause optimization so that submaps, trajectories,
+  //  and constraints are immutable for serialization of state.
+
+  // TODO(brandon-northcutt) Get a file handle to send protobufs to.
+
+  // TODO(brandon-northcutt) Send map_builder_options protobuf to file first, so
```
wohe: I think we should let the deserialization code use a different configuration if necessary.

cartographer/io/map_io.cc
```diff
@@ -0,0 +1,64 @@
...
+  // TODO(brandon-northcutt) Iterate trajectories { serialize trajectory builder
+  //  options, serialize sensor configuration proto, serialize all nodes in this
+  //  trajectory.
+  
+  // TODO(brandon-northcutt) Serialize trajectory connectivity.
+
+  // TODO(brandon-northcutt) Iterate submaps(serialize submap options, serialize
+  //  probability grids.
+  
+  // TODO(brandon-northcutt) Serialize the sparse pose graph,  (Should the SPG
```
wohe: I'd suggest serializing this first. Then the big picture is clear after looking at the first proto message in the stream.

PR #360, opened 2017-06-23
Consistency fix: trajectory_index -> trajectory_id.

SirVer:


cartographer/io/points_batch.h
```diff
@@ -49,8 +49,8 @@ struct PointsBatch {
   // is unknown.
   string frame_id;
 
-  // Trajectory index that produced this point.
-  int trajectory_index;
+  // Trajectory id that produced this point.
```
wohe: `ID`?

PR #361, opened 2017-06-23
Add IMU orientation correction to the pose graph optimization.

wohe:
Orientation of the IMU, most importantly pitch, is crucial to correctly
track changes in altitude. Each trajectory gets optimization variables
to correct for small errors in the IMU orientation.

Also adds separate gravity constants for each trajectory. The idea is
that different trajectories might use different hardware and slightly
disagree about the strength of gravity.

PAIR=damonkohler

### DIFF
--- a/cartographer/mapping_3d/acceleration_cost_function.h
+++ b/cartographer/mapping_3d/acceleration_cost_function.h
@@ -42,12 +42,12 @@ class AccelerationCostFunction {
   bool operator()(const T* const middle_rotation, const T* const start_position,
                   const T* const middle_position, const T* const end_position,
                   const T* const gravity_constant,
-                  const T* const imu_correction, T* residual) const {
-    const Eigen::Quaternion<T> eigen_imu_correction(
-        imu_correction[0], imu_correction[1], imu_correction[2],
-        imu_correction[3]);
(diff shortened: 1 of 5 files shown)


cartographer/mapping_3d/acceleration_cost_function.h
```diff
@@ -41,9 +41,14 @@ class AccelerationCostFunction {
   template <typename T>
   bool operator()(const T* const middle_rotation, const T* const start_position,
                   const T* const middle_position, const T* const end_position,
-                  const T* const gravity_constant, T* residual) const {
+                  const T* const gravity_constant,
+                  const T* const imu_correction, T* residual) const {
```
SirVer: imu_calibration?

cartographer/mapping_3d/rotation_cost_function.h
```diff
@@ -36,13 +36,17 @@ class RotationCostFunction {
 
   template <typename T>
   bool operator()(const T* const start_rotation, const T* const end_rotation,
-                  T* residual) const {
+                  const T* const imu_correction, T* residual) const {
```
SirVer: here too.

PR #374, opened 2017-06-29
Use multi-res scan matching for loop closures too.

damonkohler:


cartographer/mapping/sparse_pose_graph/proto/constraint_builder_options.proto
```diff
@@ -59,6 +59,13 @@ message ConstraintBuilderOptions {
       ceres_scan_matcher_options = 11;
   optional mapping_3d.scan_matching.proto.FastCorrelativeScanMatcherOptions
       fast_correlative_scan_matcher_options_3d = 10;
+
+  // Voxel filter used to compute a sparser point cloud for matching.
```
wohe: Can you add separate comments for each option and run `scripts/update_configuration_doc.py`? Also mention that this is for 3D loop closure refinement?

PR #375, opened 2017-06-29
Change local 3D SLAM to not use a UKF.

wohe:
This follows the 2D local SLAM in implementing a simple
constant velocity model and allows integrating odometry and
rotation information from an IMU.

### DIFF
--- a/cartographer/mapping_3d/local_trajectory_builder.cc
+++ b/cartographer/mapping_3d/local_trajectory_builder.cc
@@ -32,7 +32,6 @@ LocalTrajectoryBuilder::LocalTrajectoryBuilder(
     const proto::LocalTrajectoryBuilderOptions& options)
     : options_(options),
       active_submaps_(options.submaps_options()),
-      scan_matcher_pose_estimate_(transform::Rigid3d::Identity()),
       motion_filter_(options.motion_filter_options()),
       real_time_correlative_scan_matcher_(
           common::make_unique<scan_matching::RealTimeCorrelativeScanMatcher>(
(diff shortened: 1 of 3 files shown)


cartographer/mapping_3d/local_trajectory_builder.cc
```diff
@@ -32,53 +32,57 @@ LocalTrajectoryBuilder::LocalTrajectoryBuilder(
...
         time);
+    // Use the accelerometer to gravity align the first pose.
+    imu_tracker_->AddImuLinearAccelerationObservation(linear_acceleration);
+    pose_estimate_ = transform::Rigid3d::Rotation(imu_tracker_->orientation());
   }
 
-  pose_tracker_->AddImuLinearAccelerationObservation(time, linear_acceleration);
-  pose_tracker_->AddImuAngularVelocityObservation(time, angular_velocity);
+  Predict(time);
+  imu_tracker_->AddImuLinearAccelerationObservation(linear_acceleration);
```
damonkohler: Doesn't this mean the first observation is added twice?

cartographer/mapping_3d/local_trajectory_builder.cc
```diff
@@ -152,46 +177,62 @@ LocalTrajectoryBuilder::AddAccumulatedRangeData(
...
+         odometry_state_tracker_.newest().state_pose *
+             odometry_prediction.inverse() * pose_estimate_});
+  }
+
+  // Improve the velocity estimate.
+  if (last_scan_match_time_ > common::Time::min() &&
+      time > last_scan_match_time_) {
+    const double delta_t = common::ToSeconds(time - last_scan_match_time_);
+    velocity_estimate_ +=
+        (pose_estimate_.translation() - model_prediction.translation()) /
```
damonkohler: Wouldn't we be better off updating velocity based on the most recent two estimates after refinement by the scan matcher? This appears to be comparing pose_estimate_, as last updated by scan matching, with the previous pose_estimate_, as last updated by prediction.

cartographer/mapping_3d/local_trajectory_builder.cc
wohe: Yes, but since this has no effect: why bother?

cartographer/mapping_3d/local_trajectory_builder.cc
wohe: I'm not sure what you propose. This follows what we do in 2D: The velocity is set to the change in estimates after scan matching divided by the time between the two scan matches.

PR #376, opened 2017-06-30
Submap serialization support.

brandon-northcutt:
This is in pursuit of full state serialization/deserialization (#253).


wohe:
Thanks for the PR. I simplified it a bit and will merge once tests passed.

FYI. `optional` is preferred over `required` for protobufs. Submap objects do not know their `SubmapId`, so this information will be persisted outside of this proto message.

cartographer/mapping/submaps.h
```diff
@@ -72,14 +72,13 @@ class Submap {
       const transform::Rigid3d& global_submap_pose,
       proto::SubmapQuery::Response* response) const = 0;
 
+ protected:
+  void SetNumRangeData(int num_range_data) { num_range_data_ = num_range_data; }
```
damonkohler: const

wohe: Addressed in #380.

PR #380, opened 2017-07-03
Serialize submaps.

wohe:


cartographer/mapping/proto/submap.proto
```diff
@@ -16,15 +16,21 @@ syntax = "proto2";
...
-import "cartographer/transform/proto/transform.proto";
 import "cartographer/mapping_2d/proto/probability_grid.proto";
 import "cartographer/mapping_3d/proto/hybrid_grid.proto";
+import "cartographer/transform/proto/transform.proto";
 
-message Submap {
+message Submap2D {
   optional transform.proto.Rigid3d local_pose = 1;
   optional int32 num_range_data = 2;
   optional bool finished = 3;
```
damonkohler: Maybe add a comment here? It's not clear to me what this should mean. Clearly it's "finished" since it's been serialized...

cartographer/mapping_2d/submaps.cc
```diff
@@ -69,20 +69,19 @@ Submap::Submap(const MapLimits& limits, const Eigen::Vector2f& origin)
...
 }
 
-mapping::proto::Submap Submap::ToProto() const {
-  mapping::proto::Submap proto;
-  *proto.mutable_local_pose() = transform::ToProto(local_pose());
-  proto.set_num_range_data(num_range_data());
-  proto.set_finished(finished_);
-  *proto.mutable_probability_grid() = probability_grid_.ToProto();
-  return proto;
+void Submap::ToProto(mapping::proto::Submap* proto) const {
```
damonkohler: const

cartographer/mapping/proto/submap.proto
wohe: I added general comments how these proto messages are to be interpreted. I think it makes most sense to keep it simple and not change the meaning when serializing: serialized data is matching what was in memory.

Issue #355, opened 2017-06-23
Add a ProbabilityGridPointsProcessor for 2D data into the assets_writer_main


SirVer:
Writing X-Rays does not work for 2D data. For that, an image representation of the probability grid would be best. Adding a PointsProcessor that constructs a probability grid (with trajectory on it) and writes it out would be useful.

There are two useful incarnations of this: 

1) Something similarly to the [XRayPointsProcessor](https://github.com/googlecartographer/cartographer/blob/9fbb65084ba906886e913bf9a9b873d7587ce544/cartographer/io/xray_points_processor.h#L35), but purely 2d. This should write out a PNG (optionally with trajectories on it).
2) Something similar to the [HybridGridPointsProcessor](https://github.com/googlecartographer/cartographer/blob/9fbb65084ba906886e913bf9a9b873d7587ce544/cartographer/io/hybrid_grid_points_processor.h#L22), but writing a 2D [ProbabilityGrid](https://github.com/googlecartographer/cartographer/blob/9fbb65084ba906886e913bf9a9b873d7587ce544/cartographer/mapping_2d/proto/probability_grid.proto) proto.

SirVer:
1. was implemented in #383.
2. seems currently unnecessary. If a ROS map is required, the `occupancy_grid_node` can be used to generate one. Should it be served statically, the `map_writer` can be used to save the output into a PGM and YAML.

Calling this therefore done. @damienrg: Thanks again for implementing this!

PR #383, opened 2017-07-04
Implements a ProbabilityGridPointsProcessor

damienrg:
Hi,
this is wip on issue #355 and for now it should compile but it does not work and I need to do some cleanup.
For now, could you just tell me if:

- the name of the class are ok
- should I use a probability grid inside ImagePointsProcessor? If yes I do not know how to use it correctly so could you elaborate on this?
- in ProbabilityGridPointsProcessor, how should I create the ProbabilityGrid?


damienrg:
I signed it!

SirVer:
> the name of the class are ok

ImagePointsProcessor seems unnecessary. I think you should concentrate on ProbabilityGridPointsProcessor.

> should I use a probability grid inside ImagePointsProcessor? If yes I do not know how to use it correctly so could you elaborate on this?

Yes, I think you want to use it. Have a look at #387. We just implemented that the ProbabilityGrid can be resized. So you would create it at an arbitrary size, have something like `GrowAsNeeded` in your points processor when you process points. Inside `Flush` you would [ComputeCroppedProbabilityGrid](https://github.com/googlecartographer/cartographer/blob/a313448bf01cc10f015ad872098d58b5b9e5162c/cartographer/mapping_2d/submaps.cc#L33) so the white space is removed. This is the data you want to serialize. Now there are many useful ways of serializing it:

1) As a PNG. I suggest doing this first. Please share code with XRayPointsProcessor as much as possible. Also keep in mind that we'd like to draw the trajectory onto this PNG eventually, so you need to keep some understanding how your cropped pixels relate to the map frame.
2) As PGM/YAML for ROS. Once 1) is done, a followup PR can add a PointsProcessor to cartographer_ros, that writes out a YAML and a PGM for the ROS map server. This code would reuse the implementation of [WriteOccupancyGridToPgmAndYaml](https://github.com/googlecartographer/cartographer_ros/blob/764004736e49db70ab8a8d0d73189deedab98c7c/cartographer_ros/cartographer_ros/map_writer.cc#L82) to get to ROS like data.
3) As a proto. I am not sure if anybody cares about the proto format for now, so let's skip this.

> in ProbabilityGridPointsProcessor, how should I create the ProbabilityGrid?

I hope that is answered through #387.

damienrg:
For now I do not handle the trajectory as:
1. I am not sure that the current implementation fit with what you have in mind
2. I need to find how to do it. For now I think that I should start by adding proto::Trajectory like it is done in XRayPointsProcessor.

cartographer/io/image_utils.cc
```diff
@@ -0,0 +1,67 @@
+#include "cartographer/io/image_utils.h"
```
SirVer: I do not agree with *utils.h name of this file. How about just calling it Image?

cartographer/io/image_utils.cc
```diff
@@ -0,0 +1,67 @@
...
+
+#include <memory>
+
+#include "cairo/cairo.h"
+
```
SirVer: remove empty line between includes

cartographer/io/image_utils.cc
```diff
@@ -0,0 +1,67 @@
...
+int stride_for_width(int width)
+{
+  const int stride = cairo_format_stride_for_width(kCairoFormat, width);
+  CHECK_EQ(stride % 4, 0);
+  return stride;
+}
+
+}  // namespace
+
+Image CreateImage(int width, int height)
```
SirVer: It feels like this should now be a class `Image` with a constructor. You can add pixel accessors and hide all of the Cairo implementation in this file.

cartographer/io/image_utils.h
```diff
@@ -0,0 +1,27 @@
+#ifndef CARTOGRAPHER_IO_IMAGE_UTILS_H_
+#define CARTOGRAPHER_IO_IMAGE_UTILS_H_
+
+#include <cstdint>
+#include <vector>
+
+namespace cartographer {
+namespace io {
+
+class FileWriter;
```
SirVer: do not forward declare: https://google.github.io/styleguide/cppguide.html#Forward_Declarations

cartographer/io/image_utils.h
```diff
@@ -0,0 +1,27 @@
...
+{
+  int width;
+  int height;
+  int stride;
+  std::vector<uint32_t> pixels;
+};
+
+Image CreateImage(int width, int height);
+
+void WritePng(Image * const image, FileWriter * const file_writer);
```
SirVer: this can move into the Image class then.

cartographer/io/probability_grid_points_processor.cc
```diff
@@ -0,0 +1,147 @@
...
+#include "cartographer/common/lua_parameter_dictionary.h"
+#include "cartographer/common/make_unique.h"
+#include "cartographer/common/math.h"
+#include "cartographer/io/image_utils.h"
+
+namespace cartographer {
+namespace io {
+namespace {
+
+struct PixelData {
```
SirVer: Use `Color` typedef from points_batch.h instead?

cartographer/io/probability_grid_points_processor.cc
```diff
@@ -0,0 +1,147 @@
...
+  uint8_t r = 0;
+  uint8_t g = 0;
+  uint8_t b = 0;
+};
+
+using PixelDataMatrix =
+  Eigen::Matrix<PixelData, Eigen::Dynamic, Eigen::Dynamic>;
+
+cairo_status_t CairoWriteCallback(void* const closure,
```
SirVer: unused? Remove!

cartographer/io/probability_grid_points_processor.cc
```diff
@@ -0,0 +1,147 @@
...
+      image.pixels[y * image.stride / 4 + x] = (255 << 24) | (r << 16) | (g << 8) | b;
+    }
+  }
+
+  WritePng(&image, file_writer);
+  CHECK(file_writer->Close());
+}
+
+mapping_2d::ProbabilityGrid CreateProbabilityGrid(double resolution) {
+  constexpr int kInitialProbabilityGridSize = 100;
```
SirVer: What a terrible interface ProbabilityGrid provides for construction. I am sorry .... :(.

cartographer/io/probability_grid_points_processor.cc
```diff
@@ -0,0 +1,147 @@
...
+  common::LuaParameterDictionary* const dictionary,
+  PointsProcessor* const next) {
+  return common::make_unique<ProbabilityGridPointsProcessor>(
+    dictionary->GetDouble("resolution"),
+      mapping_2d::CreateRangeDataInserterOptions(
+          dictionary->GetDictionary("range_data_inserter").get()),
+    file_writer_factory(dictionary->GetString("filename") + ".png"), next);
+}
+
+void WriteGrid(const mapping_2d::ProbabilityGrid & probability_grid,
```
SirVer: Can be moved into anonymous namespace?

cartographer/io/probability_grid_points_processor.cc
```diff
@@ -0,0 +1,147 @@
...
+    return Eigen::Array2i(cell_limits.num_x_cells - index(0) - 1,
+                          cell_limits.num_y_cells - index(1) - 1);
+  };
+  int width = cell_limits.num_y_cells;
+  int height = cell_limits.num_x_cells;
+  PixelDataMatrix image(height, width);
+  for (auto xy_index :
+         cartographer::mapping_2d::XYIndexRangeIterator(cell_limits)) {
+    auto index = xy_index + offset;
+    if (probability_grid.IsKnown(index)) {
```
SirVer: Please write constexpr uint8_t kUnknownValue = 128; for unknown space to stay consistent with `cartographer_ros/map_writer.cc`.

cartographer/io/probability_grid_points_processor.cc
```diff
@@ -0,0 +1,147 @@
...
+      const Eigen::Array2i pixel = grid_index_to_pixel(xy_index);
+      PixelData& pixel_data = image(pixel.x(), pixel.y());
+      float probability = probability_grid.GetProbability(index);
```
SirVer: can be const. Also to stay consistent, this should be `1.f - probability_grid...`.

cartographer/io/probability_grid_points_processor.cc
```diff
@@ -0,0 +1,147 @@
...
+      uint8_t value = 255
```
SirVer: `p - min / (max - min)` should read `(p - min) / (max - min)`

cartographer/io/probability_grid_points_processor.cc
```diff
@@ -0,0 +1,147 @@
...
+        * (probability - mapping::kMinProbability
+           / (mapping::kMaxProbability - mapping::kMinProbability));
+      pixel_data.r = value;
+      pixel_data.g = value;
+      pixel_data.b = value;
+    }
+  }
+
```
SirVer: remove empty line

cartographer/io/probability_grid_points_processor.h
```diff
@@ -0,0 +1,56 @@
+#ifndef CARTOGRAPHER_IO_PROBABILITY_GRID_POINTS_PROCESSOR_H_
+#define CARTOGRAPHER_IO_PROBABILITY_GRID_POINTS_PROCESSOR_H_
+
+// Library used for inserting range data points into a probability grid.
```
SirVer: comment doesn't add a ton IMHO. remove?

cartographer/io/probability_grid_points_processor.h
```diff
@@ -0,0 +1,56 @@
...
+#include "cartographer/mapping_2d/probability_grid.h"
+#include "cartographer/mapping_2d/proto/range_data_inserter_options.pb.h"
+#include "cartographer/mapping_2d/range_data_inserter.h"
+
+namespace cartographer {
+namespace io {
+
+// Creates a probability grid with the specified 'resolution'.
+// 'range_data_inserter' options are used to configure the range
+// data ray tracing through the probability grid.
```
SirVer: Mention that the `z` component of the data is ignored, i.e. all points are projected into the x-y plane.

cartographer/io/probability_grid_points_processor.h
```diff
@@ -0,0 +1,56 @@
...
+
+  ~ProbabilityGridPointsProcessor() override {}
+
+  void Process(std::unique_ptr<PointsBatch> batch) override;
+  FlushResult Flush() override;
+
+ private:
+  PointsProcessor* const next_;
+  std::unique_ptr<FileWriter> file_writer_;
+
```
SirVer: remove empty line?

cartographer/io/image_utils.cc
```diff
@@ -0,0 +1,67 @@
...
+  if (static_cast<FileWriter*>(closure)->Write(
+        reinterpret_cast<const char*>(data), length)) {
+    return CAIRO_STATUS_SUCCESS;
+  }
+  return CAIRO_STATUS_WRITE_ERROR;
+}
+
+constexpr cairo_format_t kCairoFormat = CAIRO_FORMAT_ARGB32;
+
+int stride_for_width(int width)
```
SirVer: GetStrideForWidth

cartographer/io/image_utils.h
damienrg: Ok. I see that you seem to repeat includes from the header in the source files. Should I do the same?

damienrg:
In `ProbabilityGridPointsProcessor::Flush() ` do you think that the message of  `LOG(FATAL)...` should contain probability grid instead of image?

cartographer/io/image_utils.h
SirVer: We are not consistent about that. It is enough to include the file in the header.

SirVer:
Damien, thanks for the contribution! I fixed some minor nits and ran clang-format and then merged this. If you disagree with any of my changes we can discuss and change them in a separate PR again.

I see two followups for this PR: 

1) The point 2. I mentioned [above](https://github.com/googlecartographer/cartographer/pull/383#issuecomment-313083104) so that the occupancy grids can be used with map_server for ROS nodes. This would be a PR for `cartographer_ros`. 
2) Drawing the trajectory into the image. I think I wrote something about this already, but GitHub seems to have eaten it? I was musing that you'd need to take the Trajectories similar to how XRayPointsProcessor does it and that it would be cool to write out an SVG or TIF or something else that supports layers, so that the bottom layer would just be the map, the layer above the Trajectories and so on.

Are you interested in tackling any of these?

damienrg:
> If you disagree with any of my changes we can discuss and change them in a separate PR again.

No problem with your changes.

> 1. The point 2. I mentioned [above](https://github.com/googlecartographer/cartographer/pull/383#issuecomment-313083104) so that the occupancy grids can be used with map_server for ROS nodes. This would be a PR for cartographer_ros.

To be sure I understand what you have in mind, the purpose is to be able to use it in the pipeline description file, no?

> 2. Drawing the trajectory into the image. I think I wrote something about this already, but GitHub seems to have eaten it? I was musing that you'd need to take the Trajectories similar to how XRayPointsProcessor does it and that it would be cool to write out an SVG or TIF or something else that supports layers, so that the bottom layer would just be the map, the layer above the Trajectories and so on.

For writing a SVG or TIF, you mean by hand or using a library? I do not know TIF but for SVG it is doable (embed the map in png and add a trajectory layer).

> Are you interested in tackling any of these?

Why not and I am more interested in point 2 but as I have less free time to work on these I cannot guarantee that it will be done soon.

SirVer:
> For writing a SVG or TIF, you mean by hand or using a library? 

A library. We are using Cairo already and it does support SVG (https://www.cairographics.org/manual/cairo-SVG-Surfaces.html), not sure if it does support TIF and also not sure if it does support individual layers in images. That will require some research.

damienrg:
> A library. We are using Cairo already and it does support SVG (https://www.cairographics.org/manual/cairo-SVG-Surfaces.html), not sure if it does support TIF and also not sure if it does support individual layers in images. That will require some research.

It is only for rendering SVG, not writting.

SirVer:
> It is only for rendering SVG, not writing.

No, it does: use [cairo_svg_surface_create_for_stream](https://www.cairographics.org/manual/cairo-SVG-Surfaces.html) and use it similar to cairo_surface_write_to_png_stream.

PR #387, opened 2017-07-05
Grow the ProbabilityGrid as needed instead of preconfiguring its size.

SirVer:
This removes the 'half_length' option.

PAIR=wohe


cartographer/mapping_2d/ray_casting.cc
```diff
@@ -146,13 +146,30 @@ void CastRay(const Eigen::Array2i& begin, const Eigen::Array2i& end,
   CHECK_EQ(current.y(), end.y() / kSubpixelScale);
 }
 
+void GrowAsNeeded(const sensor::RangeData& range_data, ProbabilityGrid* const probability_grid) {
+  Eigen::AlignedBox2f bounding_box(range_data.origin.head<2>());
+  constexpr float kPadding = 1e-6;
```
wohe: `1e-6f`?

PR #389, opened 2017-07-05
Add pure localization parameter to trajectory builder options

Mofef:
This addresses https://github.com/googlecartographer/cartographer_ros/issues/399

At the moment I get an `out_of_range` error after the first trimming is done, I feel/hope this is not related to the way I add the trimmer but to how the trimming is done. Am I missing something?


jihoonl:
I described a bit more about crashing while trimming in #390.

SirVer:
Thanks, I think this is a reasonable way of implementing it for now. I fixed some nits and changed the number of submaps to 3 instead of 10 - we want to keep the computational effort as low as possible. 

I will investigate the crashes next.

PR #405, opened 2017-07-14
Support compiling with CMAKE_BUILD_TYPE set to None

jspricke:
Debian defaults to setting CMAKE_BUILD_TYPE to None and setting CFLAGS
and alike to Debian standards. This allows to set this again.

### DIFF
--- a/cmake/functions.cmake
+++ b/cmake/functions.cmake
@@ -104,6 +104,10 @@ macro(google_initialize_cartographer_project)
         "call CMake with -DFORCE_DEBUG_BUILD=True"
       )
     endif()
+# Support for Debian packaging CMAKE_BUILD_TYPE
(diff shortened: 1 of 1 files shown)


damonkohler:
Thanks for the contribution! This appears to change the default build behavior away from building in release mode. Not building in release mode has been the cause of many user issues and so we'd like to keep it as the default behavior. That said, your change did expose a nice bug that CI is now complaining about.

jspricke:
Ok, I updated the patch to only remove the error when build type is "None". Would that be ok?

damonkohler:
I think it may be better to handle your use case specifically and add a comment about Debian builds. Otherwise, we lose the error checking that is currently preventing typos. Also, have you tested the performance of the system when building in your environment? Please mention your findings in the comment.

jspricke:
Updated. We can still tune the build flags the Debian way, if it's needed.

PR #406, opened 2017-07-17
Fix compile error.

damonkohler:
See #405


SirVer:
This seems fishy - the code was clearly correct with the LOG(FATAL) being noreturn. Why are the other PointsProcessors not blowing up?

damonkohler:
This shouldn't be necessary anymore given the changes to #405. It may crop up again if someone chooses to compile with the same settings.

PR #417, opened 2017-07-21
Towards localization in 3D.

wohe:
Adds trimmer handling (but no trimming yet) and frozen trajectories
to 3D.

### DIFF
--- a/cartographer/mapping_3d/sparse_pose_graph.h
+++ b/cartographer/mapping_3d/sparse_pose_graph.h
@@ -31,6 +31,7 @@
 #include "cartographer/common/mutex.h"
 #include "cartographer/common/thread_pool.h"
 #include "cartographer/common/time.h"
+#include "cartographer/mapping/pose_graph_trimmer.h"
 #include "cartographer/mapping/sparse_pose_graph.h"
 #include "cartographer/mapping/trajectory_connectivity.h"
 #include "cartographer/mapping_3d/sparse_pose_graph/constraint_builder.h"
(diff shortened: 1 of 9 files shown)


cartographer/mapping_3d/sparse_pose_graph.h
```diff
@@ -77,17 +78,13 @@ class SparsePoseGraph : public mapping::SparsePoseGraph {
                   const Eigen::Vector3d& linear_acceleration,
                   const Eigen::Vector3d& angular_velocity);
 
-  void FreezeTrajectory(const int trajectory_id) override {
-    LOG(FATAL) << "Not yet implemented for 3D.";
-  }
+  void FreezeTrajectory(const int trajectory_id) override;
```
damonkohler: No const in prototype.

cartographer/mapping_3d/sparse_pose_graph/optimization_problem.cc
```diff
@@ -154,6 +156,12 @@ void OptimizationProblem::Solve(const std::vector<Constraint>& constraints) {
             translation_parameterization(),
             common::make_unique<ceres::QuaternionParameterization>(), &problem);
       }
+      if (frozen) {
```
damonkohler: Doesn't this mean that nodes could still move around as constraints are added to submaps in other trajectories?

wohe: Yes, this mirrored 2D. Since we currently add no nodes for maps this makes no difference in practice. I fixed in 2D and 3D anyway since we want to add nodes for maps in the future. PTAL

PR #418, opened 2017-07-21
Fix debug build

### LINKED ISSUE #412: Control reaches end of non-void function when building in debug mode
When compiling with `CMAKE_BUILD_TYPE=Debug` (yikes, yes, I know), I get the following error:

```
catkin_ws/src/cartographer/cartographer/io/hybrid_grid_points_processor.cc:64:1: 
error: control reaches end of non-void function [-Werror=return-type]
```

I have no idea why building regularly (with `CMAKE_BUILD_TYPE=Release`) doesn't trigger this.

ojura:
Add a fake return value after LOG(FATAL) in
HybridGridPointsProcessor::Flush(). This keeps gcc versions lower
than 7 happy when building without optimisation (i.e. debug build).
Fixes #412.

### DIFF
--- a/cartographer/io/hybrid_grid_points_processor.cc
+++ b/cartographer/io/hybrid_grid_points_processor.cc
@@ -61,6 +61,9 @@ PointsProcessor::FlushResult HybridGridPointsProcessor::Flush() {
       return FlushResult::kFinished;
   }
   LOG(FATAL) << "Failed to receive FlushResult::kFinished";
+  // The following unreachable return statement is needed to avoid a GCC bug
+  // described at https://gcc.gnu.org/bugzilla/show_bug.cgi?id=81508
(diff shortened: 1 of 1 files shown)


cartographer/io/hybrid_grid_points_processor.cc
```diff
@@ -61,6 +61,7 @@ PointsProcessor::FlushResult HybridGridPointsProcessor::Flush() {
       return FlushResult::kFinished;
   }
   LOG(FATAL) << "Failed to receive FlushResult::kFinished";
+  return FlushResult::kFinished;
```
SirVer: Please add a comment to the GCC bug report why this should not be removed.

ojura: Done.

Issue #414, opened 2017-07-20
ProbabilityPointsProcessor is mirroring the output


SirVer:
Compare XRayPointsProcessors output to ProbabilityGridPointsProcessor. The second one is flipping its output horizontally which make the map incorrect.

@damienrg Are you interested in looking into this?

damienrg:
I can look at this. Could you post images of XRayPointsProcessor and ProbabilityGridPointsProcessor showing the problem?

SirVer:
Yup, here are the images. I created them using the changes in [this commit](https://github.com/SirVer/cartographer_ros/commit/67a270e2aa2ec019b5274a5ec928adc4ed4ac195). First, I created the `.pbstream` using:

~~~
roslaunch cartographer_ros offline_backpack_2d.launch \
    bag_filenames:=${HOME}/Downloads/cartographer_paper_deutsches_museum.bag
~~~

then generating the assets using

~~~
roslaunch cartographer_ros assets_writer_backpack_2d.launch \
   bag_filenames:=$HOME/Downloads/cartographer_paper_deutsches_museum.bag  \
   pose_graph_filename:=$HOME/Downloads/cartographer_paper_deutsches_museum.bag.pbstream
~~~

The assets will be written into `~/.ros`. Notice how the probability grid is differently rotated and also flipped over. The X-Ray represents the world as watched from the top, which is also what RViz and ROS are using, so the probability grid should follow this convention.

![probability_grid](https://user-images.githubusercontent.com/140115/28457465-829ce9bc-6e06-11e7-811e-ee76e0fc629b.png)
![xray_xy_all](https://user-images.githubusercontent.com/140115/28457464-829bf958-6e06-11e7-99f6-3f21468aa6a7.png)

PR #419, opened 2017-07-21
Fix ProbabilityPointsProcessor mirroring output (#414)

damienrg:


SirVer:
Thank you very much, Damien!

PR #420, opened 2017-07-24
Pull some functionality from cartographer_ros into cartographer.

SirVer:
- Moved color.[h|cc].
- Moved Cairo types.
- Simplifications in xray_points_processor.cc.


cartographer/io/xray_points_processor.cc
```diff
@@ -46,9 +47,8 @@ double Mix(const double a, const double b, const double t) {
 }
 
 // Write 'mat' as a pleasing-to-look-at PNG into 'filename'
```
damonkohler: Comment is wrong

SirVer: Will fix in followup.

PR #423, opened 2017-07-24
Change Color to Uint8Color and FloatColor

damienrg:
As almost all computations for color are done in float, use this format
by default and convert to uint8_t only when needed.

### DIFF
--- a/cartographer/io/image.cc
+++ b/cartographer/io/image.cc
@@ -9,6 +9,19 @@ namespace cartographer {
 namespace io {
 namespace {
 
+uint32 Uint8ColorToCairo(const Uint8Color& color) {
+  return {static_cast<uint32>(255) << 24 |
+          static_cast<uint32>(color[0]) << 16 |
+          static_cast<uint32>(color[1]) << 8 | color[2]};
+}
+
+Uint8Color CairoToUint8Color(uint32 color) {
+  uint8 r = color >> 16;
+  uint8 g = color >> 8;
+  uint8 b = color;
+  return {{r, g, b}};
+}
+
 cairo_status_t CairoWriteCallback(void* const closure,
                                   const unsigned char* data,
                                   const unsigned int length) {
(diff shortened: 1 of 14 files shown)


cartographer/io/image.cc
```diff
@@ -9,6 +9,21 @@ namespace cartographer {
 namespace io {
 namespace {
 
+uint32_t ColorToCairo(const Color& color) {
+  uint32_t r = ColorComponentToUint8(color[0]);
+  uint32_t g = ColorComponentToUint8(color[1]);
+  uint32_t b = ColorComponentToUint8(color[2]);
```
wohe: Would it make sense to have separate types for `UInt8Color` and `FloatColor` and provide functions converting between the two?

cartographer/io/xray_points_processor.cc
```diff
@@ -81,10 +81,9 @@ Image IntoImage(const PixelDataMatrix& mat) {
       double mix_g = Mix(1., mean_g_in_column, saturation);
       double mix_b = Mix(1., mean_b_in_column, saturation);
 
-      const uint8_t r = common::RoundToInt(mix_r * 255.);
-      const uint8_t g = common::RoundToInt(mix_g * 255.);
-      const uint8_t b = common::RoundToInt(mix_b * 255.);
-      image.SetPixel(x, y, {{r, g, b}});
+      image.SetPixel(x, y,
+                     {{static_cast<float>(mix_r), static_cast<float>(mix_g),
```
wohe: Any reason to do the computations for `mix_?` in `double` and cast here? Seems the computations could be done in `float` without any problem.

cartographer/io/image.cc
```diff
@@ -9,6 +9,21 @@ namespace cartographer {
...
+  return (255 << 24) | (r << 16) | (g << 8) | b;
```
wohe: The (255 << 24) is undefined behavior: Cast the 255 to `uint32` before shifting?

wohe:
Thanks for the PR!

damienrg:
I created Uint8Color and FloatColor but I do not know if it is better, just let me what you think about that.
I still parse color as uint8 values in coloring_points_processor.cc to not break compatibility.

cartographer/io/image.cc
```diff
@@ -9,6 +9,19 @@ namespace cartographer {
 namespace io {
 namespace {
 
+uint32 Uint8ColorToCairo(const Uint8Color& color) {
+  return {static_cast<uint32>(255) << 24 |
```
wohe: No need for the curly braces. Remove?

damienrg: Fixed.

damienrg:
Do you think that I should rename Uint8Color to RgbUint8Color and FloatColor to RgbFloatColor?

wohe:
No, let's keep the names concise for readability.

Issue #409, opened 2017-07-18
Trimming a submap in the middle


jihoonl:
This is a necessary work to replace an outdated submap with a newly generated submap that covers the same area. With this feature, I think cartographer can actually do the lifelong mapping over time. I would like to open a discussion to start the development.

I think we can break this down into two separate problems, trimming trajectory nodes and trimming a submap.

Submap trimming seems relatively easy. All we have to do is changing vector<deque> to vector<map> just as in cartographer_rviz (and of course some following works).

Trimming trajectory nodes seems to need a design decision(or discussion) to decide how cartographer keeps the separated trajectories. Should cartographer keep the separated trajectories in the nested map structure? should it keep the entire trajectory without trimming until no constraint references this trajectory? Or any other options?

How do you think?

SirVer:
We thought that this should be modeled by 

1) Split trajectory.
2) Trim one of the from the beginning.

2) is done, 1) should be relatively simple (famous last words).

If you trim nodes in the middle you have an awkward trajectory with missing data and non-even timestamp increases.

jihoonl:
Alright. I will try to split trajectory.

wohe:
I was trying to review #425, but I think we need some discussion first how we approach trimming in the middle. How I see it, there are a couple of things that matter:

- we label data presented to the user with pairs of trajectory ID and node index (e.g. in RViz) where it is useful to look at individual trajectories (similarly we use ID/index pairs for submaps)
- we also use these ID pairs to label input data, i.e. to tell the SLAM system which trajectory builder some data is for
- we need to be able to trim arbitrary nodes and submaps for life-long mapping, and this should be efficient
- the internal representation of the pose graph is an implementation detail, therefore it does not have to use ID/index pairs

I see two very different general approaches to implement trimming in the middle:
1. Change the so called "trajectory IDs" to only identify where data came from and allow "trajectories" to contain holes. "Trajectories" with holes are no longer a single trajectory. We must detect and handle these gaps in node indices, e.g. in RViz and SLAM. By construction this maintains labels, e.g. in RViz it is easy for users to understand what is happening. In this case, the data structures for nodes/submaps could be changed from `std::vector<std::deque<..>> to `std::vector<std::map<int,...>>`.
[...]

PR #425, opened 2017-07-25
Get new trajectory id from sparse pose graph

jihoonl:
Related to #409, #437 

To split a trajectory in sparse pose graph level.


jihoonl:
We decided not to continue in this direction.

PR #428, opened 2017-07-25
Better tuning for 2D.

SirVer:


SirVer:
before:
xray_xy_all

after:
xray_xy_all

koelsinha:
Hello!
I have just started using cartographer and have a basic question. In the before and after, the change that is immediately noticeable is the reduction in the number of the circular rings around the map.  I just want to understand what these rings are and what parameter affects this the most?

PR #429, opened 2017-07-26
rate_timer.h allow user specified pulse time

ahundt:
I want to collect my second "pulse" time from a source other than the current time on the local clock, this change defaults to the current behavior but allows the user to specify their own time.

With the default behavior this also has the convenient side effect of collecting the time before starting the function call.

### DIFF
--- a/cartographer/common/rate_timer.h
+++ b/cartographer/common/rate_timer.h
@@ -68,8 +68,8 @@ class RateTimer {
   }
 
   // Records an event that will contribute to the computed rate.
-  void Pulse(common::Time time) {
-    events_.push_back(Event{time, ClockType::now()});
+  void Pulse(common::Time time, typename ClockType::time_point compare_time = ClockType::now()) {
+    events_.push_back(Event{time, compare_time});
     while (events_.size() > 2 &&
            (events_.back().wall_time - events_.front().wall_time) >
                window_duration_) {


wohe:
Can you give a bit of context what you want to do, i.e. where will the functionality be used?

ahundt:
I'm using this class to collect converted time stamps from hardware clocks on an optical tracker and robot arm in relation to a local CPU clock, and I'm collecting the timestamp immediately before and after my driver receives the data, but I don't want my driver to depend on cartographer. For that reason I'd like to supply both time stamps myself.

wohe:
If you're using this outside of the cartographer library itself, I suggest forking. We want to move towards a public API and my guess is that we will stop exposing `rate_timer.h` before a 1.0 release.

ahundt:
I installed the cartographer API using the CMake scripts and I'm including it from my own code.

The following is very preliminary and subject to change. I'm working on improving timestamp estimates as a way to improve localization for an optical tracker, computer, and robot arm. I expect very similar time correction algorithms incorporating physical measurements could provide similar benefits for the IMU, LIDAR, and RGBD hardware used by cartographer applications to better estimate and minimize mapping error in slam software. I found cartographer to be well written with relatively few dependencies, especially since ROS is optional. This makes for a very good starting point for testing across platforms and minimizing infrastructure work.

I hoped some of this might become directly useful to cartographer users, so I was hoping to minimize the difference between my fork and cartographer by pulling useful changes upstream early in a collaborative fashion. However, if you think it is best that I simply make a permanent fork I will understand and close my PRs accordingly.

wohe:
Just to be clear: I was suggesting you fork the `RateTimer` class and put it in your code base, not forking all of Cartographer. The way I see it, `RateTimer` is currently only used internally for logging/debugging and will not become part of the public API. The public API will focus on SLAM.

I think it would make sense for helper classes that are useful specifically for SLAM users to become part of the public API. So if you can make that case for the `RateTimer`, I'm happy to reconsider.

ahundt:
Thank your for your explanation, your motivation makes sense to me since you're focusing on SLAM end users. In my own words, Cartographer is not middleware for algorithm developers and extending the logging/debug components is not part of the current mission of Cartographer. It may even prove distracting!

Since you mentioned you're definitely interested in something that proves concretely useful for SLAM I'll keep that in mind in case I eventually end up with something worthwhile. In the meantime I'll just put any useful cartographer utilities I employ and modify directly in my own code base. Thanks for your time & feedback! :-)

PR #430, opened 2017-07-26
Add a extrapolator for poses.

wohe:
This will be used in Cartographer ROS to extrapolate poses for tf.

PAIR=damonkohler

### DIFF
--- a/cartographer/mapping/pose_extrapolator.cc
+++ b/cartographer/mapping/pose_extrapolator.cc
@@ -1,0 +1,73 @@
+/*
+ * Copyright 2017 The Cartographer Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "cartographer/mapping/pose_extrapolator.h"
+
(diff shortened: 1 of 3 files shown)


cartographer/mapping/pose_extrapolator.cc
```diff
@@ -0,0 +1,73 @@
...
+  if (timed_pose_queue_.empty()) {
+    return common::Time::min();
+  }
+  return timed_pose_queue_.back().time;
+}
+
+void PoseExtrapolator::AddPose(const common::Time time,
+                               const transform::Rigid3d& pose) {
+  timed_pose_queue_.push_back(TimedPose{time, pose});
+  while (timed_pose_queue_.size() > 2 &&
```
ensonic: I would not expect a method called Add* to pop items off the vector.

cartographer/mapping/pose_extrapolator.cc
```diff
@@ -0,0 +1,73 @@
...
+  const auto last_time = last_timed_pose.time;
+  CHECK(time >= last_time);
+  if (timed_pose_queue_.size() == 1) {
+    return last_timed_pose.pose;
+  }
+  const double queue_delta =
+      common::ToSeconds(last_time - timed_pose_queue_.front().time);
+  const transform::Rigid3d& last_pose = last_timed_pose.pose;
+  const Eigen::Vector3d linear_velocity =
+      (last_pose.translation() - timed_pose_queue_.front().pose.translation()) /
```
ensonic: since you use a local var for last_timed_pose, maybe also do:
recent_timed_pose = timed_pose_queue_.front()

cartographer/mapping/pose_extrapolator.cc
```diff
@@ -0,0 +1,73 @@
...
+      queue_delta;
```
ensonic: can this be 0?

cartographer/mapping/pose_extrapolator.cc
wohe: This is consistent with the code base. The interesting fact is that this adds new pose information.

PR #432, opened 2017-07-27
color.cc constexpr std::sqrt is not ISO C++

ahundt:
```
Semantic Issue Group
/Users/athundt/source/cartographer/cartographer/io/color.cc:32:17: Constexpr variable 'kGoldenRatioConjugate' must be initialized by a constant expression
```

https://stackoverflow.com/questions/8622256/in-c11-is-sqrt-defined-as-constexpr

### DIFF
--- a/cartographer/io/color.cc
+++ b/cartographer/io/color.cc
@@ -29,7 +29,6 @@ namespace {
 constexpr float kInitialHue = 0.69f;
 constexpr float kSaturation = 0.85f;
 constexpr float kValue = 0.77f;
-constexpr float kGoldenRatioConjugate = (std::sqrt(5.f) - 1.f) / 2.f;
 
 Color CreateRgba(const float r, const float g, const float b) {
   return Color{{static_cast<uint8_t>(common::RoundToInt(r * 255.)),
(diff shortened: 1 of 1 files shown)


cartographer/io/color.cc
```diff
@@ -69,6 +68,7 @@ Color GetColor(int id) {
   CHECK_GE(id, 0);
   // Uniform color sampling using the golden ratio from
   // http://martin.ankerl.com/2009/12/09/how-to-create-random-colors-programmatically/
+  float kGoldenRatioConjugate = (std::sqrt(5.f) - 1.f) / 2.f;
```
SirVer: const float golden_ration_conjugate

ahundt: maybe static const? then it should only be calculated once

Or could add something like [this compile time implementation](https://baptiste-wicht.com/posts/2014/07/compile-integer-square-roots-at-compile-time-in-cpp.html) to math.h:

```c++
#define MID(a, b) ((a+b)/2)
#define POW(a) (a*a)

template<std::size_t res, std::size_t l = 1, std::size_t r = res>
struct ct_sqrt;

template<std::size_t res, std::size_t r>
struct ct_sqrt<res, r, r> : std::integral_constant<std::size_t, r> {};

template <std::size_t res, std::size_t l, std::size_t r>
struct ct_sqrt : std::integral_constant<std::size_t, ct_sqrt<res,
        (POW(MID(r, l)) >= res ? l : MID(r, l)+1),
        (POW(MID(r, l)) >= res ? MID(r, l) : r)>::value> {};
```

ahundt: Or even better just use the constant provided by boost:
http://www.boost.org/doc/libs/1_63_0/libs/math/doc/html/math_toolkit/constants.html

macmason: Or evaluate it in python, and then write the result down: `0.6180339887498949f`

ahundt: ah I resent you and your practical solutions @macmason. :-)

PR #434, opened 2017-07-28
Use IMU data when extrapolating poses if available.

wohe:


cartographer/mapping/pose_extrapolator.cc
```diff
@@ -71,8 +77,57 @@ transform::Rigid3d PoseExtrapolator::ExtrapolatePose(common::Time time) {
...
+}
+
+void PoseExtrapolator::TrimImuData() {
+  while (imu_data_.size() > 2 && !timed_pose_queue_.empty() &&
+         imu_data_[1].time <= timed_pose_queue_.back().time) {
+    imu_data_.pop_front();
+  }
+}
+
+Eigen::Quaterniond PoseExtrapolator::RotationFromAngularVelocity(
```
SirVer: anon namespace and remove from header.

PR #446, opened 2017-08-10
Fix ClangTidy warnings

zjwoody:
Fix the ClangTidy warnings, build and passed all test.


damonkohler:
@wohe please carefully consider the change to cartographer/mapping_2d/sparse_pose_graph.cc

cartographer/mapping_2d/sparse_pose_graph.cc
```diff
@@ -432,8 +432,9 @@ void SparsePoseGraph::RunOptimization() {
   if (optimization_problem_.submap_data().empty()) {
     return;
   }
-  optimization_problem_.Solve(constraints_, frozen_trajectories_);
+
   common::MutexLocker locker(&mutex_);
+  optimization_problem_.Solve(constraints_, frozen_trajectories_);
```
wohe: Please revert this change.

By construction, no other thread is accessing the `optimization_problem_`, `constraints_` and `frozen_trajectories_` while we execute this line. `Solve()` can take a significant amount of time, so taking the mutex here would block foreground processing. We do not have reader locks, so I see no easy way to improve the situation besides maybe adding a comment.

zjwoody: Done.
Thanks for the review!

zjwoody:
I sent another PR, close this one.

PR #448, opened 2017-08-10
Use PoseExtrapolator in the 2D trajectory builder.

wohe:
This replaces the Predict() logic in the 2D trajectory builder.

### DIFF
--- a/cartographer/mapping_2d/local_trajectory_builder.cc
+++ b/cartographer/mapping_2d/local_trajectory_builder.cc
@@ -31,8 +31,7 @@ LocalTrajectoryBuilder::LocalTrajectoryBuilder(
       motion_filter_(options_.motion_filter_options()),
       real_time_correlative_scan_matcher_(
           options_.real_time_correlative_scan_matcher_options()),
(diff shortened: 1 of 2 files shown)


cartographer/mapping_2d/local_trajectory_builder.cc
```diff
@@ -223,64 +193,29 @@ LocalTrajectoryBuilder::pose_estimate() const {
...
-            transform::GetYaw(pose_estimate_.rotation()) - last_yaw,
-            Eigen::Vector3d::UnitZ()) *
-        imu_tracker_->orientation();
-    pose_estimate_ = transform::Rigid3d(translation, rotation);
+void LocalTrajectoryBuilder::InitializeExtrapolator(const common::Time time) {
+  if (extrapolator_ != nullptr) {
+    return;
   }
-  time_ = time;
+  constexpr double kExtrapolationEstimationTimeSec = 0.001;  // 1 ms
```
damonkohler: A comment here about what this means would be useful. Based on my understanding, this should perhaps not even be a parameter but rather just a constant in the PoseExtrapolator?

PR #450, opened 2017-08-11
Use PoseExtrapolator in the 3D trajectory builder.

wohe:
This replaces the Predict() logic in the 3D trajectory builder.

### DIFF
--- a/cartographer/mapping/pose_extrapolator.h
+++ b/cartographer/mapping/pose_extrapolator.h
@@ -40,6 +40,10 @@ class PoseExtrapolator {
   PoseExtrapolator(const PoseExtrapolator&) = delete;
   PoseExtrapolator& operator=(const PoseExtrapolator&) = delete;
 
+  static std::unique_ptr<PoseExtrapolator> InitializeWithImu(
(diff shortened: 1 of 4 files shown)


cartographer/mapping/pose_extrapolator.h
```diff
@@ -49,6 +49,16 @@ class PoseExtrapolator {
   void AddOdometryData(const sensor::OdometryData& odometry_data);
   transform::Rigid3d ExtrapolatePose(common::Time time);
 
+  // Must be called after the first IMU data and no pose were added. Estimates
+  // an initial gravity aligned orientation from the linear acceleration of the
+  // first IMU data.
+  void AddInitialGravityAlignedPose();
```
damonkohler: Shouldn't this be called in 2D as well? We also use it for gravity alignment there I think?

Also, at that point, doesn't it make sense to just require that, if IMU data is added, it must be added first so that this can be done as the first action? Maybe taking one reading in the constructor or something?

wohe: I addressed the second issue. I'll change 2D in a follow up PR since this will slightly change 2D SLAM behavior.

PR #455, opened 2017-08-15
Fix ClangTidy warnings

zjwoody:


ojura:
@SirVer has mentioned that he believes `const` in headers to be an unnecessary implementation detail (in the definition of the function it is okay, in fact, it is desirable to have it; but not in the interface).

zjwoody:
Hi @ojura , 
Thanks for your comment!
I agree that top-level const on function parameters in declarations that are not definitions is meaningless and best omitted because it will be ignored by the compiler and mislead readers.
My change is using the const reference so it is significant and can be used to distinguish overloaded function declarations.

ojura:
Sorry, I haven't been looking carefully - now I see that all the places where you inserted `const`, you also turned them into references. Never mind. :-)

Issue #370, opened 2017-06-28
Add an odometry term in 2D/3D modeled after 2D's consecutive scan term


damonkohler:
In 2D, since we don't tie poses together with IMU reading, we tie poses together by position. That is, from one pose to the next, the position shouldn't change drastically without sufficient evidence.

The same approach could be used to integrate odometry information. That is, relative odometry readings from one position to the next should not be violated without sufficient evidence.

BrannonKing:
It sounds like this is the same bug as #242 and #467 and https://github.com/googlecartographer/cartographer_ros/issues/488, or that they have all morphed together as PoseExtrapolator has taken over the odometry->velocity role. PoseExtrapolator::AddImuData is just too sensitive to noise in the odometry data. Most odometry data comes from wheel encoders; they slip and they often aren't regularly periodic. There is some noisy calculus to convert from multiple wheel encoders into a single odometry point. Simulating them (from my experience) isn't noise-free either -- the timing in the simulation tools is lousy.

SirVer:
@cschuet @wohe Is this fixed with your recent 2D/3D unification?

cschuet:
Fixed in #570.

PR #456, opened 2017-08-16
Add odometry term to optimization problem (wip)

Mofef:
This is to address #370 , but it's still work in progress. However since there are a lot of changes related to odometry happening these days I felt it's better to come forward with some intermediate state to avoid double work.
A detailed review on this is probably a waste of time but i'm happy for some directions

Unaddressed problems are:
* New parameters should be extracted to configuration files.
* So far I don't pay attention to the time stamps.
* Removing the old consecutive scan distance minimization terms results in a broken trajectory
* The translational scaling factor seams to take whatever lower bound I offer it without a visible effect to the map. So probably I have a mistake in the problem formulation.
* Probably hard bounds on those parameters are not the best way to go. A penalty with a scaling parameter could be better.
* Maybe you prefer odometry parameters to stay outside of cartographer entirely. I could then remove the related code from this PR and open a new issue.


cartographer/mapping_2d/sparse_pose_graph.cc
```diff
@@ -164,6 +164,15 @@ void SparsePoseGraph::AddImuData(const int trajectory_id,
   });
 }
 
+void SparsePoseGraph::AddOdometerData(const int trajectory_id,
+                                      const sensor::OdometryData odometry_data) {
+
```
wohe: Remove empty line?

cartographer/mapping_2d/sparse_pose_graph.h
```diff
@@ -97,6 +97,9 @@ class SparsePoseGraph : public mapping::SparsePoseGraph {
       override EXCLUDES(mutex_);
   std::vector<Constraint> constraints() override EXCLUDES(mutex_);
 
+  void AddOdometerData(const int trajectory_id,
```
wohe: Move right after `AddImuData()` similar to the `.cc` file?

cartographer/mapping_2d/sparse_pose_graph/optimization_problem.cc
```diff
@@ -23,6 +23,7 @@
 #include <memory>
 #include <string>
 #include <vector>
+#include <cartographer/sensor/odometry_data.h>
```
wohe: Style: This belongs to the next group of includes in alphabetical order.

cartographer/mapping_2d/sparse_pose_graph/optimization_problem.h
```diff
@@ -78,6 +79,9 @@ class OptimizationProblem {
   int num_trimmed_nodes(int trajectory_id) const;
   int num_trimmed_submaps(int trajectory_id) const;
 
+  void AddOdometerData(const int trajectory_id,
```
wohe: Move right after to `AddImuData()` as in the `.cc`?

cartographer/mapping_2d/sparse_pose_graph/optimization_problem.h
```diff
@@ -22,6 +22,7 @@
 #include <map>
 #include <set>
 #include <vector>
+#include <cartographer/sensor/odometry_data.h>
```
wohe: Style: Move to the next block of includes.

cartographer/mapping_2d/sparse_pose_graph/optimization_problem.cc
```diff
@@ -192,6 +209,27 @@ void OptimizationProblem::Solve(const std::vector<Constraint>& constraints,
...
+      problem.AddResidualBlock(
+          new ceres::AutoDiffCostFunction<WarpedSpaCostFunction, 3, 3, 3, 2>(
+              new WarpedSpaCostFunction(Constraint::Pose{
+                  odometry_data_[trajectory_id][odometry_data_index - 1]
+                      .pose.inverse() *
+                  odometry_data_[trajectory_id][odometry_data_index]
+                      .pose,
+                  0.3, 30.})),
+          nullptr /* loss function */,
+          C_nodes[trajectory_id][odometry_data_index - 1].data(),
```
wohe: This seems wrong: indices of nodes and odometry data are different in general since usually odometry has a higher frequency. I think what you want to do is look at the time stamps of two consecutive nodes, then use the data in `odometry_data_` to compute the relative change expected according to odometry. Since you most likely do not have the exact timestamp in the odometry data, you'd have to interpolate between two odometry poses, once for the timestamp of the first node and once for the timestamp of the second node.

cartographer/mapping_2d/sparse_pose_graph/spa_cost_function.h
```diff
@@ -81,6 +81,56 @@ class SpaCostFunction {
   const Constraint::Pose pose_;
 };
 
+class WarpedSpaCostFunction {
```
wohe: I suggest leaving this out for now. Let's keep the focus of this PR on introducing odometry to the optimization. Adding scaling factors can be done in a follow up PR.

cartographer/mapping_2d/sparse_pose_graph/optimization_problem.cc
```diff
@@ -192,6 +209,27 @@ void OptimizationProblem::Solve(const std::vector<Constraint>& constraints,
...
+          C_nodes[trajectory_id][odometry_data_index].data(),
+          C_odometry[trajectory_id].data());
+    }
+  }
+
```
wohe: Since we now tie together consecutive scans using odometry, the following code should be dropped/changed. As I see it odometry can take the place of
```
node_data_[trajectory_id][node_data_index - 1]
    .initial_point_cloud_pose.inverse() *
node_data_[trajectory_id][node_data_index]
    .initial_point_cloud_pose
```
I'd suggest changing the code to have one loop and either use a new computation of this relative pose using odometry if available, or the old code otherwise.

wohe:
If you would like to merge a smaller PR sooner, feel free to send a PR containing just the code that moves odometry data to the optimization problem without making use of it yet.

cartographer/mapping_2d/sparse_pose_graph/optimization_problem.cc
Mofef: Yes, thanks. I'm currently working on it.

Mofef:
Please take another look :)

cartographer/mapping_2d/sparse_pose_graph/optimization_problem.cc
```diff
@@ -192,25 +201,47 @@ void OptimizationProblem::Solve(const std::vector<Constraint>& constraints,
...
-                  options_.consecutive_scan_rotation_penalty_factor()})),
-          nullptr /* loss function */,
-          C_nodes[trajectory_id][node_data_index - 1].data(),
-          C_nodes[trajectory_id][node_data_index].data());
+      if (trajectory_id < odometry_data_.size() &&
+          odometry_data_[trajectory_id].Has(
+              node_data_[trajectory_id][node_data_index].time) &&
+          odometry_data_[trajectory_id].Has(
+              node_data_[trajectory_id][node_data_index - 1].time)) {
+        problem.AddResidualBlock(
```
wohe: This is a lot of duplication. How about extracting the `relative_pose`, e.g.
```
const transform::Rigid2d relative_pose = odometry_available ? ... : ...
problem.AddResidualBlock(
...

Mofef:
Ok ready for another round :)

Mofef:
Done

PR #458, opened 2017-08-16
Estimate angular velocities from odometry.

### LINKED ISSUE #453: Drastically decreased quality with the new PoseExtrapolator in LocalTrajectoryBuilder
Hi @wohe, the newest head (currently 11dbdf9) looks like this on a couple of my bags. You can see that the quality of the local trajectory builder has drastically decreased:

screenshot from 2017-08-13 20-38-16

And here is how it looked before the pose extrapolator got merged:
screenshot from 2017-08-13 21-23-41

Here are the bags and configuration files: https://drive.google.com/file/d/0B695JUjxohhNVWFpUUFuaG95SG8/view?usp=sharing

To reproduce:
1) Clone https://github.com/larics/cartographer and checkout the `rangefinder-decimation` branch (decimation is needed for these bags, or the pose estimate diverges with the old velocity model - however, when decimated, it worked great, as you can see from the second picture.). Use vanilla `cartographer_ros`.
2) Build and run the offline node (the command line arguments for running the offline node have been provided in `offline_node_arguments`)
3) To revert the extrapolator commits and get back good behaviour, revert the extrapolator commits in `cartographer`: `git revert -n 73b0e^..11dbdf9`

wohe:
If no IMU is available we will now use odometry to estimate
angular velocities if available instead of the last poses.

Fixes #453.

### DIFF
--- a/cartographer/mapping/pose_extrapolator.cc
+++ b/cartographer/mapping/pose_extrapolator.cc
@@ -90,22 +90,29 @@ void PoseExtrapolator::AddOdometryData(
         odometry_data.time >= timed_pose_queue_.back().time);
   odometry_data_.push_back(odometry_data);
   TrimOdometryData();
-  if (odometry_data_.size() < 2 || timed_pose_queue_.empty()) {
+  if (odometry_data_.size() < 2) {
     return;
   }
   // TODO(whess): Improve by using more than just the last two odometry poses.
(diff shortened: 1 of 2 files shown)


ojura:
Thanks on taking the time to debug and fix this. I have a question: why not use the velocity from the twist part of the odometry message?

wohe:
This is currently not available in Cartographer as we decided to keep the API minimal and just contain poses.

PR #472, opened 2017-08-23
Use vector<map> for node_data

jihoonl:
~~Note that I have added `todo` in TrimTrajectoryNode to trim out imu and odometry data. I wasn't sure how to handle them when trajectory node in the middle get trimmed.~~

~~Also, the difference between node indices as another factor to penalize consecutive scans. You might want to change this too.~~

### DIFF
--- a/cartographer/mapping_2d/sparse_pose_graph.cc
+++ b/cartographer/mapping_2d/sparse_pose_graph.cc
@@ -193,9 +193,8 @@ void SparsePoseGraph::ComputeConstraint(const mapping::NodeId& node_id,
               .pose.inverse() *
           optimization_problem_.node_data()
               .at(node_id.trajectory_id)
-              .at(node_id.node_index - optimization_problem_.num_trimmed_nodes(
-                                           node_id.trajectory_id))
-              .pose;
+              .at(node_id.node_index)
+              .point_cloud_pose;
       constraint_builder_.MaybeAddConstraint(
           submap_id, submap_data_.at(submap_id).submap.get(), node_id,
           trajectory_nodes_.at(node_id).constant_data.get(),
@@ -210,14 +209,9 @@ void SparsePoseGraph::ComputeConstraintsForOldScans(
   const auto& node_data = optimization_problem_.node_data();
   for (size_t trajectory_id = 0; trajectory_id != node_data.size();
        ++trajectory_id) {
-    for (size_t node_data_index = 0;
-         node_data_index != node_data[trajectory_id].size();
-         ++node_data_index) {
-      const int node_index =
-          node_data_index +
-          optimization_problem_.num_trimmed_nodes(trajectory_id);
+    for (const auto& index_node_data : node_data[trajectory_id]) {
       const mapping::NodeId node_id{static_cast<int>(trajectory_id),
-                                    static_cast<int>(node_index)};
+                                    index_node_data.first};
       CHECK(!trajectory_nodes_.at(node_id).trimmed());
       if (submap_data.node_ids.count(node_id) == 0) {
         ComputeConstraint(node_id, submap_id);
@@ -248,9 +242,9 @@ void SparsePoseGraph::ComputeConstraintsForScan(
               optimization_problem_.node_data().size()
           ? static_cast<int>(optimization_problem_.node_data()
                                  .at(matching_id.trajectory_id)
-                                 .size()) +
-                optimization_problem_.num_trimmed_nodes(
-                    matching_id.trajectory_id)
+                                 .rbegin()
+                                 ->first +
+                             1)
           : 0};
   const auto& scan_data = trajectory_nodes_.at(node_id).constant_data;
   optimization_problem_.AddTrajectoryNode(
@@ -439,15 +433,14 @@ void SparsePoseGraph::RunOptimization() {
   const auto& node_data = optimization_problem_.node_data();
   for (int trajectory_id = 0;
        trajectory_id != static_cast<int>(node_data.size()); ++trajectory_id) {
-    int node_data_index = 0;
     const int num_nodes = trajectory_nodes_.num_indices(trajectory_id);
-    int node_index = optimization_problem_.num_trimmed_nodes(trajectory_id);
-    for (; node_data_index != static_cast<int>(node_data[trajectory_id].size());
-         ++node_data_index, ++node_index) {
-      const mapping::NodeId node_id{trajectory_id, node_index};
+    int last_node_data_index = 0;
+    for (auto it = node_data.at(trajectory_id).begin();
+         it != node_data.at(trajectory_id).end(); ++it) {
+      const mapping::NodeId node_id{trajectory_id, it->first};
       trajectory_nodes_.at(node_id).pose =
-          transform::Embed3D(node_data[trajectory_id][node_data_index].pose) *
-          trajectory_nodes_.at(node_id).constant_data->tracking_to_tracking_2d;
+          transform::Embed3D(it->second.point_cloud_pose);
+      last_node_data_index = it->first;
     }
     // Extrapolate all point cloud poses that were added later.
     const auto local_to_new_global =
@@ -456,7 +449,8 @@ void SparsePoseGraph::RunOptimization() {
         optimized_submap_transforms_, trajectory_id);
     const transform::Rigid3d old_global_to_new_global =
         local_to_new_global * local_to_old_global.inverse();
-    for (; node_index < num_nodes; ++node_index) {
+    for (int node_index = last_node_data_index + 1; node_index < num_nodes;
+         ++node_index) {
       const mapping::NodeId node_id{trajectory_id, node_index};
       trajectory_nodes_.at(node_id).pose =
           old_global_to_new_global * trajectory_nodes_.at(node_id).pose;
--- a/cartographer/mapping_2d/sparse_pose_graph.cc
+++ b/cartographer/mapping_2d/sparse_pose_graph.cc
@@ -194,7 +194,7 @@ void SparsePoseGraph::ComputeConstraint(const mapping::NodeId& node_id,
           optimization_problem_.node_data()
               .at(node_id.trajectory_id)
               .at(node_id.node_index)
-              .point_cloud_pose;
+              .pose;
       constraint_builder_.MaybeAddConstraint(
           submap_id, submap_data_.at(submap_id).submap.get(), node_id,
           trajectory_nodes_.at(node_id).constant_data.get(),
@@ -438,7 +438,7 @@ void SparsePoseGraph::RunOptimization() {
     for (const auto& node_data_index : node_data.at(trajectory_id)) {
       const mapping::NodeId node_id{trajectory_id, node_data_index.first};
       trajectory_nodes_.at(node_id).pose =
-          transform::Embed3D(node_data_index.second.point_cloud_pose);
+          transform::Embed3D(node_data_index.second.pose);
     }
     // Extrapolate all point cloud poses that were added later.
     const auto local_to_new_global =
--- a/cartographer/mapping_2d/sparse_pose_graph.cc
+++ b/cartographer/mapping_2d/sparse_pose_graph.cc
@@ -239,7 +239,8 @@ void SparsePoseGraph::ComputeConstraintsForScan(
   const mapping::NodeId node_id{
       matching_id.trajectory_id,
       static_cast<size_t>(matching_id.trajectory_id) <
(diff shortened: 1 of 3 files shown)


cartographer/mapping_2d/sparse_pose_graph.cc
```diff
@@ -444,23 +438,25 @@ void SparsePoseGraph::RunOptimization() {
...
-    int node_data_index = 0;
     const int num_nodes = trajectory_nodes_.num_indices(trajectory_id);
-    int node_index = optimization_problem_.num_trimmed_nodes(trajectory_id);
-    for (; node_data_index != static_cast<int>(node_data[trajectory_id].size());
-         ++node_data_index, ++node_index) {
-      const mapping::NodeId node_id{trajectory_id, node_index};
-      trajectory_nodes_.at(node_id).pose = transform::Embed3D(
-          node_data[trajectory_id][node_data_index].point_cloud_pose);
+    int last_node_data_index = 0;
+    for (auto it = node_data.at(trajectory_id).begin();
```
wohe: Use a range based for loop?

cartographer/mapping_2d/sparse_pose_graph.cc
```diff
@@ -444,23 +438,25 @@ void SparsePoseGraph::RunOptimization() {
...
+         it != node_data.at(trajectory_id).end(); ++it) {
+      const mapping::NodeId node_id{trajectory_id, it->first};
+      trajectory_nodes_.at(node_id).pose =
+          transform::Embed3D(it->second.point_cloud_pose);
+      last_node_data_index = it->first;
```
wohe: Why update this? In the end this is `node_data.at(trajectory_id).rbegin()->first` or `0` if the trajectory contains no optimized nodes.

cartographer/mapping_2d/sparse_pose_graph.cc
```diff
@@ -444,23 +438,25 @@ void SparsePoseGraph::RunOptimization() {
...
        trajectory_id != static_cast<int>(node_data.size()); ++trajectory_id) {
-    int node_data_index = 0;
     const int num_nodes = trajectory_nodes_.num_indices(trajectory_id);
-    int node_index = optimization_problem_.num_trimmed_nodes(trajectory_id);
-    for (; node_data_index != static_cast<int>(node_data[trajectory_id].size());
-         ++node_data_index, ++node_index) {
-      const mapping::NodeId node_id{trajectory_id, node_index};
-      trajectory_nodes_.at(node_id).pose = transform::Embed3D(
-          node_data[trajectory_id][node_data_index].point_cloud_pose);
+    int last_node_data_index = 0;
```
wohe: Move down where it is used. `last_optimized_node_index`?

cartographer/mapping_2d/sparse_pose_graph/optimization_problem.cc
```diff
@@ -194,47 +190,55 @@ void OptimizationProblem::Solve(const std::vector<Constraint>& constraints,
...
+        const NodeData& node_data = index_node_data->second;
+        ++index_node_data;
+        if (index_node_data == node_data_[trajectory_id].end()) {
+          break;
+        }
+
+        const int next_node_index = index_node_data->first;
+        const NodeData& next_node_data = index_node_data->second;
+
+        const int index_diff = next_node_index - node_index;
```
wohe: Not sure what this is about. Move to a separate PR?

cartographer/mapping_2d/sparse_pose_graph/optimization_problem.cc
```diff
@@ -257,16 +261,14 @@ void OptimizationProblem::Solve(const std::vector<Constraint>& constraints,
...
   for (size_t trajectory_id = 0; trajectory_id != node_data_.size();
        ++trajectory_id) {
-    for (size_t node_data_index = 0;
-         node_data_index != node_data_[trajectory_id].size();
-         ++node_data_index) {
-      node_data_[trajectory_id][node_data_index].point_cloud_pose =
-          ToPose(C_nodes[trajectory_id][node_data_index]);
+    for (auto& index_node_data : node_data_[trajectory_id]) {
+      index_node_data.second.point_cloud_pose =
+          ToPose(C_nodes[trajectory_id][index_node_data.first]);
```
wohe: Consistency: Use `.at(index_node_data.first)` which checks that the data we want to copy exists?

cartographer/mapping_2d/sparse_pose_graph/optimization_problem.cc
```diff
@@ -194,47 +190,55 @@ void OptimizationProblem::Solve(const std::vector<Constraint>& constraints,
...
-      problem.AddResidualBlock(
-          new ceres::AutoDiffCostFunction<SpaCostFunction, 3, 3, 3>(
-              new SpaCostFunction(Constraint::Pose{
-                  relative_pose,
-                  options_.consecutive_scan_translation_penalty_factor(),
-                  options_.consecutive_scan_rotation_penalty_factor()})),
-          nullptr /* loss function */,
-          C_nodes[trajectory_id][node_data_index - 1].data(),
-          C_nodes[trajectory_id][node_data_index].data());
+    if (node_data_[trajectory_id].size() > 1) {
```
wohe: Isn't this check redundant? Remove?

cartographer/mapping_2d/sparse_pose_graph/optimization_problem.cc
```diff
@@ -194,47 +190,55 @@ void OptimizationProblem::Solve(const std::vector<Constraint>& constraints,
...
+      for (auto index_node_data = node_data_[trajectory_id].begin();;) {
```
wohe: Rename `index_node_data` since this is actually an iterator. `node_it`?

cartographer/mapping_2d/sparse_pose_graph/optimization_problem.cc
```diff
@@ -194,47 +190,55 @@ void OptimizationProblem::Solve(const std::vector<Constraint>& constraints,
...
+        const int node_index = index_node_data->first;
+        const NodeData& node_data = index_node_data->second;
+        ++index_node_data;
+        if (index_node_data == node_data_[trajectory_id].end()) {
+          break;
+        }
+
+        const int next_node_index = index_node_data->first;
+        const NodeData& next_node_data = index_node_data->second;
+
```
wohe: `continue` if `next_node_index != node_index + 1`, i.e. there is a gap?

cartographer/mapping_2d/sparse_pose_graph/optimization_problem.cc
```diff
@@ -170,14 +163,17 @@ void OptimizationProblem::Solve(const std::vector<Constraint>& constraints,
   for (size_t trajectory_id = 0; trajectory_id != node_data_.size();
        ++trajectory_id) {
     const bool frozen = frozen_trajectories.count(trajectory_id);
-    // Reserve guarantees that data does not move, so the pointers for Ceres
-    // stay valid.
-    C_nodes[trajectory_id].reserve(node_data_[trajectory_id].size());
-    for (const NodeData& node_data : node_data_[trajectory_id]) {
-      C_nodes[trajectory_id].push_back(FromPose(node_data.point_cloud_pose));
-      problem.AddParameterBlock(C_nodes[trajectory_id].back().data(), 3);
+
```
wohe: Remove empty line?

cartographer/mapping_2d/sparse_pose_graph/optimization_problem.cc
```diff
@@ -85,26 +85,19 @@ void OptimizationProblem::AddTrajectoryNode(
...
+  ++trajectory_data.next_node_index;
 }
 
 void OptimizationProblem::TrimTrajectoryNode(const mapping::NodeId& node_id) {
-  auto& trajectory_data = trajectory_data_.at(node_id.trajectory_id);
-  // We only allow trimming from the start.
-  CHECK_EQ(trajectory_data.num_trimmed_nodes, node_id.node_index);
   auto& node_data = node_data_.at(node_id.trajectory_id);
-  CHECK(!node_data.empty());
-  if (node_id.trajectory_id < static_cast<int>(imu_data_.size())) {
```
wohe: Retain similar logic to remove all IMU data that is no longer needed?

cartographer/mapping_2d/sparse_pose_graph.cc
jihoonl: It is being assigned in right after the declaration.([here](https://github.com/magazino/cartographer/blob/use_map_trajectory_node/cartographer/mapping_2d/sparse_pose_graph.cc#L448)). Or you think it makes sense to assign it right before [this line](https://github.com/magazino/cartographer/blob/use_map_trajectory_node/cartographer/mapping_2d/sparse_pose_graph.cc#L458) with `node_data.at(trajectory_id).rbegin()->first`?

jihoonl: nevermind. you have answered below.

cartographer/mapping_2d/sparse_pose_graph/optimization_problem.cc
jihoonl: No, it isn't. `++node_it` causes a problem when `node_data_[trajectory_id]` is empty. For loop doesn't have an initial condition check before `++node_it`

jihoonl: I will replace it with `if(!node_data_[trajectory_id].empty())`

cartographer/mapping_2d/sparse_pose_graph.cc
```diff
@@ -253,9 +247,9 @@ void SparsePoseGraph::ComputeConstraintsForScan(
               optimization_problem_.node_data().size()
           ? static_cast<int>(optimization_problem_.node_data()
                                  .at(matching_id.trajectory_id)
-                                 .size()) +
-                optimization_problem_.num_trimmed_nodes(
-                    matching_id.trajectory_id)
+                                 .rbegin()
```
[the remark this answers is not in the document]
jihoonl: Done.

jihoonl:
Done.

cartographer/mapping_2d/sparse_pose_graph.cc
```diff
@@ -250,12 +244,13 @@ void SparsePoseGraph::ComputeConstraintsForScan(
   const mapping::NodeId node_id{
       matching_id.trajectory_id,
       static_cast<size_t>(matching_id.trajectory_id) <
-              optimization_problem_.node_data().size()
+                  optimization_problem_.node_data().size() &&
+              !optimization_problem_.node_data().empty()
```
wohe: `!optimization_problem_.node_data()[matching_id.trajectory_id].empty()`?

cartographer/mapping_2d/sparse_pose_graph/optimization_problem.cc
```diff
@@ -85,26 +85,44 @@ void OptimizationProblem::AddTrajectoryNode(
...
   auto& node_data = node_data_.at(node_id.trajectory_id);
-  CHECK(!node_data.empty());
   if (node_id.trajectory_id < static_cast<int>(imu_data_.size())) {
-    const common::Time node_time = node_data.front().time;
-    auto& imu_data = imu_data_.at(node_id.trajectory_id);
-    while (imu_data.size() > 1 && imu_data[1].time <= node_time) {
-      imu_data.pop_front();
+    auto node_it = node_data.begin();
+    for (; std::next(node_it, 1) != node_data.end(); ++node_it) {
+      if (std::next(node_it, 1)->first == node_id.node_index) {
```
wohe: This seems a complicated way to implement `find(node_id.node_index)`. Most importantly for trimming the first node of trajectory (which is what we do right now), it seems to never trim any IMU data. Rewrite? Maybe along these lines:

- Remove node from `node_data`.
- If the trajectory is not yet empty, find the time of the now first node and call it `node_time`.
- following logic as before

Does that make sense?

jihoonl: I don't see how it deletes imu_date in the middle. But, it should be fine since there is only `PureLocalizationTrimmer`. Ok, I will follow the suggestion.

jihoonl:
Done.

cartographer/mapping_2d/sparse_pose_graph/optimization_problem.cc
wohe: It does not and it might make sense to add this functionality in the future. I just wanted to make sure there is no regression.

PR #478, opened 2017-08-24
Add histograms for pose residuals

cschuet:
This PR adds code for the calculation of histograms for the translational and rotational intra submap constraint residuals.

The idea is that the observation of these histograms helps tune parameters such as `translation_weight` and `rotation_weight` in the `ceres_scan_matcher`.


cartographer/mapping_3d/sparse_pose_graph.cc
```diff
@@ -423,6 +423,34 @@ void SparsePoseGraph::RunFinalOptimization() {
...
+    if (constraint.tag == Constraint::Tag::INTRA_SUBMAP) {
+      // in map frame
+      const cartographer::transform::Rigid3d& optimized_pose_map =
+          trajectory_nodes_.at(constraint.node_id).pose;
+      // in submap frame
+      const cartographer::transform::Rigid3d& original_pose_submap =
+          constraint.pose.zbar_ij;
+      // find the submap to map transform.
+      const cartographer::transform::Rigid3d& submap_transform =
+          optimized_submap_transforms_.at(constraint.submap_id.trajectory_id)
```
cschuet: Is it save here to assume the submap_transform structure always contains the correct transform already? Or do we need bound checks here?

cartographer/mapping_3d/sparse_pose_graph.cc
```diff
@@ -423,6 +423,34 @@ void SparsePoseGraph::RunFinalOptimization() {
           .max_num_iterations());
 }
 
+void SparsePoseGraph::CalculateHistograms() {
+  translational_residual_ = common::Histogram();
```
cschuet: Should we add a ::reset() method to the histogram here?

cartographer/mapping_3d/sparse_pose_graph.cc
```diff
@@ -423,6 +423,34 @@ void SparsePoseGraph::RunFinalOptimization() {
...
+      // original pose in the map frame.
+      const cartographer::transform::Rigid3d original_pose_map =
+          submap_transform * original_pose_submap;
+
+      translational_residual_.Add(
+          (optimized_pose_map.inverse() * original_pose_map)
+              .translation()
+              .norm());
+    }
+  }
```
cschuet: Is is ok to simply output the histogram here? Do you want to control creation and display of the histogram by a flag?_

cartographer/mapping_3d/sparse_pose_graph.cc
wohe: Any reason we keep these histograms around? Seems they have to be recomputed every time, so we could make this a local variable?

cartographer/mapping_3d/sparse_pose_graph.cc
wohe: Yes, just output. Might make sense to add an options. `log_residual_histograms`?

cartographer/mapping_3d/sparse_pose_graph.cc
```diff
@@ -423,6 +423,34 @@ void SparsePoseGraph::RunFinalOptimization() {
           .max_num_iterations());
 }
 
+void SparsePoseGraph::CalculateHistograms() {
+  translational_residual_ = common::Histogram();
+  for (const Constraint& constraint : constraints_) {
+    if (constraint.tag == Constraint::Tag::INTRA_SUBMAP) {
+      // in map frame
```
wohe: Consistency: map <- node (since this is the transform from the node frame to the map frame)
Even better: Name the variables `optimized_node_to_map`, `node_to_submap_constraint`, `optimized_submap_to_map`.

cartographer/mapping_3d/sparse_pose_graph.cc
wohe: Since we run this code after `Solve()` was called with the `constraints_` this should always be the case.

cartographer/mapping_3d/sparse_pose_graph.cc
```diff
@@ -423,6 +423,34 @@ void SparsePoseGraph::RunFinalOptimization() {
...
+      const cartographer::transform::Rigid3d& original_pose_submap =
+          constraint.pose.zbar_ij;
+      // find the submap to map transform.
+      const cartographer::transform::Rigid3d& submap_transform =
+          optimized_submap_transforms_.at(constraint.submap_id.trajectory_id)
+              .at(constraint.submap_id.submap_index)
+              .pose;
+
+      // original pose in the map frame.
+      const cartographer::transform::Rigid3d original_pose_map =
```
wohe: Might be more readable to combine the two optimized poses to an `optimized_node_to_submap`, then compare to the `node_to_submap_constraint`.

cartographer/mapping_3d/sparse_pose_graph.cc
```diff
@@ -423,6 +423,34 @@ void SparsePoseGraph::RunFinalOptimization() {
...
+          submap_transform * original_pose_submap;
+
+      translational_residual_.Add(
+          (optimized_pose_map.inverse() * original_pose_map)
```
wohe: Is this the same as `optimized_node_to_submap.translation() - node_to_submap_constraint.translation()`?

cartographer/mapping/proto/sparse_pose_graph_options.proto
```diff
@@ -47,4 +47,7 @@ message SparsePoseGraphOptions {
   // Rate at which we sample a single trajectory's scans for global
   // localization.
   optional double global_sampling_ratio = 5;
+
+  // Whether to calculate and output histograms for the pose residuals.
```
wohe: Drop the `calculate and`? Sounds like an implementation detail. Please run `scripts/update_configuration_doc.py`.

configuration_files/sparse_pose_graph.lua
```diff
@@ -92,4 +92,5 @@ SPARSE_POSE_GRAPH = {
   },
   max_num_final_iterations = 200,
   global_sampling_ratio = 0.003,
+  log_residual_histograms = true,
```
wohe: You also have to add this option to `mapping_2d/sparse_pose_graph_test.cc`.

cartographer/mapping_3d/sparse_pose_graph.h
```diff
@@ -158,6 +158,9 @@ class SparsePoseGraph : public mapping::SparsePoseGraph {
   mapping::SparsePoseGraph::SubmapData GetSubmapDataUnderLock(
       const mapping::SubmapId& submap_id) REQUIRES(mutex_);
 
+  // Calculates histograms for the translational residual of scan poses.
+  void CalculateResidualHistograms();
```
wohe: `LogResidualHistograms()`?

cartographer/mapping_3d/sparse_pose_graph.cc
```diff
@@ -423,6 +423,41 @@ void SparsePoseGraph::RunFinalOptimization() {
           .max_num_iterations());
 }
 
+void SparsePoseGraph::CalculateResidualHistograms() {
+  common::Histogram rotational_residual;
+  common::Histogram translational_residual;
+
+  for (const Constraint& constraint : constraints_) {
+    if (constraint.tag == Constraint::Tag::INTRA_SUBMAP) {
+
```
wohe: Remove most of these empty lines?

cartographer/mapping_3d/sparse_pose_graph.cc
```diff
@@ -423,6 +423,41 @@ void SparsePoseGraph::RunFinalOptimization() {
...
+      const cartographer::transform::Rigid3d& optimized_node_to_map =
```
wohe: Consistency: Remove the `&`? Similar below.

PR #481, opened 2017-08-25
Custom constraints

ojura:
As fixed frame constraints are now an active topic, I decided to chime in with my implementation of something similar for 2D (I called them '~~manual~~ custom constraints'). It's quite hack-ish (~~they are inserted with respect to  submap 0~~ not anymore) and not nearly considered complete (~~I wanted to also implement editing/removing them~~ removing implemented). Although this is conceptually similar to adding fixed pose measurements, its indended use is different - it bypasses the trajectory builder completely and enables the library user to directly edit the sparse pose graph. After a block of custom constraints is inserted, optimization akin to loop closures is re-run.  This enables the user to see the immediate effects of the added constraints.

Todo:
- update ConnectedComponents and TrajectoryConnectivityState?
- check that there are no unused custom constraints after final optimization

### DIFF
--- a/cartographer/mapping/proto/sparse_pose_graph.proto
+++ b/cartographer/mapping/proto/sparse_pose_graph.proto
@@ -37,7 +37,7 @@ message SparsePoseGraph {
     enum Tag {
       INTRA_SUBMAP = 0;
       INTER_SUBMAP = 1;
-      MANUAL = 2;
+      CUSTOM = 2;
     }
 
     optional SubmapId submap_id = 1;  // Submap ID.
--- a/cartographer/mapping/proto/sparse_pose_graph.proto
+++ b/cartographer/mapping/proto/sparse_pose_graph.proto
@@ -37,6 +37,7 @@ message SparsePoseGraph {
     enum Tag {
       INTRA_SUBMAP = 0;
       INTER_SUBMAP = 1;
+      MANUAL = 2;
     }
 
     optional SubmapId submap_id = 1;  // Submap ID.
--- a/cartographer/mapping/sparse_pose_graph.cc
+++ b/cartographer/mapping/sparse_pose_graph.cc
@@ -44,6 +44,8 @@ SparsePoseGraph::Constraint::Tag FromProto(
       return SparsePoseGraph::Constraint::Tag::INTRA_SUBMAP;
     case proto::SparsePoseGraph::Constraint::INTER_SUBMAP:
       return SparsePoseGraph::Constraint::Tag::INTER_SUBMAP;
+    case proto::SparsePoseGraph::Constraint::CUSTOM:
+      return SparsePoseGraph::Constraint::Tag::CUSTOM;
   }
   LOG(FATAL) << "Unsupported tag.";
 }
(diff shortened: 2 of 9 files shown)


ojura:
Hi @wohe, I have returned from vacation. I still have some catching up to do on what has been going on in Cartographer (and there is a lot :-)), especially with the fixed pose constraints. Have you managed to take a look at this?

Mofef:
Hey would you like to explain the use case for those manual constraints?

damienrg:
If this pull request will be merged in the future, I think that you should add method to add constraint by time because for now the user have to know the current submap_id and node_id.
`void SparsePoseGraph::RemoveCustomConstraint` remove all constraints for submap_id and node_id. Maybe add an id to constraint (not necessarily unique)?
I have a question in optimization_problem, in 3D they handle fixed frame pose with specific code:
https://github.com/googlecartographer/cartographer/blob/3744988fd2f18cf03f10b828e4be70b0ac01a9f5/cartographer/mapping_3d/sparse_pose_graph/optimization_problem.cc#L306
and this is is not the case in your pull request, can you explain the difference?

ojura:
@Mofef thanks on the inquiry. We are looking into enabling user intervention in the mapping process. Unfortunately, at this point I can't yet publicly go into exact details about this (but when I do, I will send you an update). We have decided to contribute back this modification to libcartographer because we hoped it would otherwise be useful (and so we don't have to maintain it in a fork :-)).

@damienrg 
> I think that you should add method to add constraint by time because for now the user have to know the current submap_id and node_id.

The same libcartographer user can call `GetAllSubmapData()` and `GetTrajectoryNodes()` methods of the `SparsePoseGraph` to get the information they need. The functions `Add/RemoveCustomConstraints` were envisioned to be a tool for exact manipulation of the sparse pose graph, and the language of the SPG are node and submap IDs, not time. So I think what you proposed is outside the scope of this PR, but something like `LookupTrajectoryNodeByTime` could be added in another PR. You are welcome to take a shot :-)

> void SparsePoseGraph::RemoveCustomConstraint remove all constraints for submap_id and node_id. Maybe add an id to constraint (not necessarily unique)?

Currently, it removes all _custom_ constraints with that submap_id and node_id (it does not touch the internally generated `INTRA` and `INTER` constraints). I agree that adding IDs to constraints might be a good idea, but in that case I would make them unique, otherwise it does not provide much benefit - the tuple [submap_id, node_id] already serves well as a non-unique constraint ID, and the tag can be used to distinguish them further apart.

> I have a question in optimization_problem, in 3D they handle fixed frame pose with specific code:

The code in this PR is actually older than the fixed frame pose measurements that you mentioned. Nonetheless, the intent is different: fixed frame poses are appropriate for handling a stream of fixed frame pose measurements in an unknown global coordinate system, such as GPS - if I have correctly understood the code, the first fixed frame pose measurement, tied to a certain trajectory node, is taken as a reference for the subsequent measurements.

On the other hand, the functionality in this PR is much more flexible - it enables you not only to add global constraints, but also local constraints as well - you can choose the ID of the submap with which you wish to constrain a certain node. Think more of it like loop closure. 

Also, fixed frame pose measurements, IMU data and odometry bypass the pose graph and are used directly in the optimization problem. I think that is not a very good thing, and that it would be better if we had a more "universal" constraint which could encompass both classic submap inter/intra constraints, and IMU data, odometry and fixed poses. That way, they could all be members of the sparse pose graph. Perhaps this universal constraint could be a submap-to-node constraint or a node-to-node constraint.

Okay, I went really off tangent there. I wanted to point out that the custom constraints are "well behaved" citizens of the pose graph and they can be visualized like other constraints because they _are_ constraints, just with a different tag.

SirVer:
@ojura This sounds elaborated enough to go through an RFC. Can we close this PR until we discussed its design and motivation in there a bit more?

ojura:
Sure.

PR #485, opened 2017-08-29
Fix Lint and ClangTidy Warinings

zjwoody:


wohe:
I think we do not want to change the `.svg` files. They were copied verbatim. Revert this part?

zjwoody:
Will the svg file change in future? Because I need to fix the trailing whitespace lint error when sync them, if they won't change anymore, I can avoid sync them.

wohe:
I don't expect them to change.

PR #486, opened 2017-08-29
Add Cartographer high level structure diagram

zjwoody:
I'm thinking maybe add a shared doc is better than pdf, so that everybody can update the diagram for the new changes.


jihoonl:
I agree with you that an image file or shared doc is better than pdf.

Odometry is also used in optimization problem since #456 is merged.

zjwoody:
Thanks @cschuet ! Could you please merge this pr, because I don't have the write access to this repository. Thanks:)

jihoonl:
I think the embedded image would be better than the link. How do you think

zjwoody:
Done. Thanks for the review!

PR #487, opened 2017-08-29
Add mutex locker in LogResidualHistograms().

zjwoody:


wohe:
Isn't this just unlocking the mutex to immediately lock it again to output the histograms? Seems easier without the change. Can you explain the rationale?

zjwoody:
I have to lock the mutex in LogResidualHistograms() to make it compile in google, and this will cause dead lock in RunOptimization() because both of them need to lock this mutex.

wohe:
@zjwoody Can you please verify whether this is fixed now?

zjwoody:
Yes, it is fixed now, thanks!

PR #488, opened 2017-08-30
Refactoring: Do not pass 'constant_data' in pieces.

wohe:
Also removes the no longer used range data from trajectory nodes
in 2D.

### DIFF
--- a/cartographer/mapping_2d/local_trajectory_builder.cc
+++ b/cartographer/mapping_2d/local_trajectory_builder.cc
@@ -17,6 +17,7 @@
 #include "cartographer/mapping_2d/local_trajectory_builder.h"
 
 #include <limits>
+#include <memory>
 
 #include "cartographer/common/make_unique.h"
 #include "cartographer/sensor/range_data.h"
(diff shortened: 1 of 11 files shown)


cartographer/mapping_2d/local_trajectory_builder.cc
```diff
@@ -186,10 +187,15 @@ LocalTrajectoryBuilder::AddAccumulatedRangeData(
...
       adaptive_voxel_filter.Filter(range_data_in_tracking_2d.returns);
 
-  return common::make_unique<InsertionResult>(
-      InsertionResult{time, std::move(insertion_submaps),
-                      tracking_to_tracking_2d, range_data_in_tracking_2d,
-                      pose_estimate_2d, filtered_point_cloud_in_tracking_2d});
+  return common::make_unique<InsertionResult>(InsertionResult{
+      std::make_shared<const mapping::TrajectoryNode::Data>(
+          mapping::TrajectoryNode::Data{time,
+                                        {},
```
cschuet: {} /* what is empty here? */

and below

ojura:
Just a quick question: how will the asset writer build a map from the final optimized trajectory if the range data is no longer being kept around?

wohe:
The asset writer does not make use of this data right now already. It is replaying the sensor data from a bag file and transforms it according to the optimized trajectory.

PR #492, opened 2017-08-31
Add thread annotation to LogResidualHistograms().

wohe:


wohe:
This should resolve #487.

PR #496, opened 2017-08-31
Pass initial pose of nodes to the 3D optimization problem.

wohe:
Also drops 'point_cloud' from the name since the pose is also
relevant to other sensor data.


wohe:
Yes, that is intended. This is towards adding odometry-like consecutive pose constraints to 3D.

PR #501, opened 2017-09-04
Allow rotating histograms in the rotational scan matcher.

wohe:
This allows reusing histograms to match for different yaws.
Adds a unit test to test that rotated histograms match as expected.

### DIFF
--- a/cartographer/mapping_3d/scan_matching/fast_correlative_scan_matcher.cc
+++ b/cartographer/mapping_3d/scan_matching/fast_correlative_scan_matcher.cc
@@ -125,6 +125,24 @@ struct Candidate {
   bool operator>(const Candidate& other) const { return score > other.score; }
 };
 
+namespace {
+
+scan_matching::RotationalScanMatcher CreateRotationalScanMatcher(
+    const std::vector<mapping::TrajectoryNode>& nodes,
+    const int histogram_size) {
+  Eigen::VectorXf histogram = Eigen::VectorXf::Zero(histogram_size);
+  for (const mapping::TrajectoryNode& node : nodes) {
+    histogram += scan_matching::RotationalScanMatcher::ComputeHistogram(
+        sensor::TransformPointCloud(
+            node.constant_data->range_data.returns.Decompress(),
+            node.pose.cast<float>()),
+        histogram_size);
+  }
+  return scan_matching::RotationalScanMatcher({{histogram, 0.f}});
+}
+
+}  // namespace
+
 FastCorrelativeScanMatcher::FastCorrelativeScanMatcher(
     const HybridGrid& hybrid_grid,
     const HybridGrid* const low_resolution_hybrid_grid,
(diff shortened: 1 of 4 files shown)


cartographer/mapping_3d/scan_matching/fast_correlative_scan_matcher.cc
```diff
@@ -286,7 +305,10 @@ std::vector<DiscreteScan> FastCorrelativeScanMatcher::GenerateDiscreteScans(
     angles.push_back(rz * angular_step_size);
   }
   const std::vector<float> scores = rotational_scan_matcher_.Match(
-      sensor::TransformPointCloud(fine_point_cloud, initial_pose), angles);
+      RotationalScanMatcher::ComputeHistogram(
+          sensor::TransformPointCloud(fine_point_cloud, initial_pose),
+          options_.rotational_histogram_size()),
+      0.f, angles);
```
cschuet: document constant /* ... */ pleaase

cartographer/mapping_3d/scan_matching/rotational_scan_matcher.cc
```diff
@@ -122,68 +117,78 @@ sensor::PointCloud SortSlice(const sensor::PointCloud& slice) {
...
-  for (const Eigen::Vector3f& point : point_cloud) {
-    slices[common::RoundToInt(point.z() / kSliceHeight)].push_back(point);
+// Rotates the given 'histogram' by the given 'angle'. This might lead to
+// rotations of a fractional bucket which is handled by linearly interpolating.
+Eigen::VectorXf RotateHistogram(const Eigen::VectorXf& histogram,
+                                const float angle) {
+  const float rotate_by_buckets = -angle * histogram.size() / M_PI;
+  int full_buckets = common::RoundToInt(rotate_by_buckets - 0.5f);
+  const float fraction = rotate_by_buckets - full_buckets;
+  while (full_buckets < 0) {
```
cschuet: while -> if and then assert full_buckets > 0?

cartographer/mapping_3d/scan_matching/rotational_scan_matcher_test.cc
```diff
@@ -0,0 +1,68 @@
...
+  RotationalScanMatcher matcher({{Eigen::VectorXf::Unit(kNumBuckets, 3), 0.f}});
+  for (float t = 0.f; t < 1.f; t += 0.1f) {
+    // 't' is the fraction of overlap and we have to divide by the norm of the
+    // histogram to get the expected score.
+    const float expected_score = t / std::hypot(t, 1 - t);
+    // We rotate the 't'-th fraction of a bucket into the matcher's histogram.
+    auto scores = matcher.Match(Eigen::VectorXf::Unit(kNumBuckets, 2), 0.f,
+                                {t * kAnglePerBucket});
+    ASSERT_EQ(1, scores.size());
+    EXPECT_NEAR(expected_score, scores[0], 1e-6);
```
cschuet: std::numeric_limits<float>::eps()?

cartographer/mapping_3d/scan_matching/rotational_scan_matcher.cc
wohe: Isn't that just more brittle? I would prefer to make no accuracy assumption here, though I agree that a sane floating point implementation should always be done in one iteration.

cartographer/mapping_3d/scan_matching/rotational_scan_matcher_test.cc
wohe: Similar here: There are multiple operations involved, so accuracy might be lower. While I imagine the test might pass: why `EXPECT` so much? If it is within `1e-6` I'm convinced of the correctness in this test case. Are you not?

cartographer/mapping_3d/scan_matching/rotational_scan_matcher.cc
wohe: Sorry, I confused this with the fractional part. This is repeated because the angle might be smaller than -pi.

PR #502, opened 2017-09-06
Check ratios in fixed rate sampler

ojura:
Edit: Since sensor sampling has been moved to cartographer_ros, I edited this PR to keep only the sampler ratio check.

~~In https://github.com/googlecartographer/cartographer_ros/issues/488, @BrannonKing and @jpapon have suggested that they would like to be able to decimate odometry data (which sounds weird to me, but they probably have a reason to do so).~~

~~You may remember that I also have been using decimation of rangefinder data for a long time, since Cartographer was crashing due to unstable velocity estimation. That does not happen anymore with the new pose extrapolator; however, I still find it very useful to be able to decimate rangefinder data because it still produces good results, while significantly speeding up Cartographer.~~

~~I have extended this to odometry, and just for completion sake, IMU data as well, in case someone is crazy enough to do that.~~


cschuet:
Hi Juraj,

Thanks for your contribution. Is there a specific reason why you prefer to do the filtering in cartographer rather than cartographer_ros? Since this data originates outside of cartographer doing all the ROS serialization / deserialization seems wasteful to me when the goal is to eventually throw some of that data away.

Also I would argue that in cartographer the typical use case keeps all of the data. It seems slightly disproportionate to force these samplers on everybody if most people will set the sampling ratio to 1.

So what do you think about moving this into cartographer_ros? Another option would be to pipe data through a subsampling ROS node that republishes the subsampled data to cartographer, but I also understand that this comes at the cost of additional ROS overhead.

ojura:
Hi, thanks for the answer. It seemed to me that doing it in the global trajectory builder seemed the most natural place to do it in the pipeline, with the minimum amount of additional code needed (there are only three cases here).

I am strongly focused on using the offline node, so piping data through another ROS node is not an option for me. If I were to do it in `cartographer_ros`, it would have to be in either `Node` or `SensorBridge`. Perhaps it would better to do it in `Node` to reduce as much deserialization as possible, as you suggested.

Also, I do not believe that pulsing a sampler with a unity ratio introduces much overhead. If you are worried about that, perhaps that case can be handled with an `if (ratio < 1.) { sampler.Pulse() }`, although from my experience, the Cartographer maintainers dislike premature optimization very much :-)

BrannonKing:
This drops a percentage of the data or it only allows data in on a fixed time period? The issue I'm seeing with the odometry is caused by a large variance on the message period. The PoseExtrapolator is very unforgiving to this; it explodes that variance out with large translations on occasion. If you have a large jump in odometry position over a short period of time -- there is nothing to filter that out specifically. Pure sampling of the data might filter that out -- assuming it never sends in two of the originally consecutive messages.

ojura:
This drops a fixed percentage of data. Although it wouldn't be difficult to write a different sampler.

damonkohler:
Thanks, @ojura, and sorry for the delay. While the checks against 0 and 1 are clear, could you perhaps make the warning a bit more informative? Why is a small ratio bad?

ojura:
I imagined that it could only occur because of a mistake, e.g. someone initializes a ratio float to zero and forgets to update it. The sampler would then silently drop all data.

damonkohler:
I guess then it makes the most sense to just change the check against 0 to CHECK_GT() and drop the additional warning log. Sound reasonable?

ojura:
Okay, but there is a test which initializes the sampler with zero and checks if it drops all the data. Should that part of the test be removed as well? That is the reason why I opted to just soft-check if the ratio is small.

damonkohler:
Rather than remove that part of the test, you could make it a death test and add another death tests for ratios greater than 1.

ojura:
Done.

PR #503, opened 2017-09-06
Store rotational histogram as part of the node data.

wohe:
The compressed point cloud is no longer stored as part of the
node data to reduce memory consumption.

### DIFF
--- a/cartographer/mapping/map_builder.cc
+++ b/cartographer/mapping/map_builder.cc
@@ -181,9 +181,7 @@ void MapBuilder::SerializeState(io::ProtoStreamWriter* const writer) {
         // TODO(whess): Handle trimmed data.
         range_data_proto->mutable_node_id()->set_trajectory_id(trajectory_id);
         range_data_proto->mutable_node_id()->set_node_index(node_index);
-        const auto& data = *node_data[trajectory_id][node_index].constant_data;
-        *range_data_proto->mutable_range_data() =
-            sensor::ToProto(data.range_data);
+        range_data_proto->mutable_range_data();
         // TODO(whess): Only enable optionally? Resulting pbstream files will be
         // a lot larger now.
         writer->WriteProto(proto);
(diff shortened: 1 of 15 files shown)


cartographer/mapping/map_builder.cc
```diff
@@ -181,9 +181,7 @@ void MapBuilder::SerializeState(io::ProtoStreamWriter* const writer) {
         // TODO(whess): Handle trimmed data.
         range_data_proto->mutable_node_id()->set_trajectory_id(trajectory_id);
         range_data_proto->mutable_node_id()->set_node_index(node_index);
-        const auto& data = *node_data[trajectory_id][node_index].constant_data;
-        *range_data_proto->mutable_range_data() =
-            sensor::ToProto(data.range_data);
+        range_data_proto->mutable_range_data();
```
cschuet: I don't understand. Is this not a no-op?

cartographer/mapping/trajectory_node.h
```diff
@@ -46,6 +43,7 @@ struct TrajectoryNode {
     // Used for loop closure in 3D.
     sensor::PointCloud high_resolution_point_cloud;
     sensor::PointCloud low_resolution_point_cloud;
+    Eigen::VectorXf rotational_scan_matcher_data;
```
cschuet: Mention that this is histogram data?

cartographer/mapping_3d/local_trajectory_builder.h
```diff
@@ -68,6 +68,7 @@ class LocalTrajectoryBuilder {
       const Eigen::Quaterniond& gravity_alignment,
       const sensor::PointCloud& high_resolution_point_cloud,
       const sensor::PointCloud& low_resolution_point_cloud,
+      const Eigen::VectorXf& rotational_scan_matcher_data,
```
cschuet: histogram?

cartographer/mapping/map_builder.cc
wohe: We don't have any range data anymore and I would like to send a follow up PR for serializing the new data to keep this PR small and focused. This is adding the proto message without adding any data in it. So strictly speaking not a no-op.

PR #504, opened 2017-09-07
Serialize trajectory node data.

wohe:
Related to #253.

### DIFF
--- a/cartographer/mapping/proto/serialization.proto
+++ b/cartographer/mapping/proto/serialization.proto
@@ -18,7 +18,7 @@ package cartographer.mapping.proto;
 
 import "cartographer/mapping/proto/sparse_pose_graph.proto";
 import "cartographer/mapping/proto/submap.proto";
-import "cartographer/sensor/proto/sensor.proto";
+import "cartographer/mapping/proto/trajectory_node.proto";
 
 message Submap {
   optional SubmapId submap_id = 1;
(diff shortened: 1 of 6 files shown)


cartographer/mapping/proto/serialization.proto
```diff
@@ -18,21 +18,21 @@ package cartographer.mapping.proto;
...
+message NodeData {
   optional NodeId node_id = 1;
-  optional sensor.proto.CompressedRangeData range_data = 4;
+  optional TrajectoryNode trajectory_node = 5;
 }
 
 message SerializedData {
   optional Submap submap = 1;
-  optional RangeData range_data = 2;
+  optional NodeData node_data = 2;
```
cschuet: Is it fine to break proto backwards compatibility? You could also add [deprecated=true] to range_data.

cartographer/mapping/trajectory_node.h
```diff
@@ -56,6 +57,8 @@ struct TrajectoryNode {
   transform::Rigid3d pose;
 };
 
+proto::TrajectoryNode ToProto(const TrajectoryNode::Data& constant_data);
```
cschuet: I don't see FromProto. Where are you planning to deserialize the data? Add FromProto and a round-trip unit test?

cartographer/mapping/proto/serialization.proto
wohe: Yes. We are before version 1.0 and make no guarantees.

ojura:
Please have a look at #508.

PR #506, opened 2017-09-07
Update trajectory connection in WhenDone callback.

cschuet:
This modifies the ConstraintBuilder interface so that Maybe*Constraint
is not responsible anymore for updating the TrajectoryConnectivity
member of SparsePoseGraph and moves this responsibility into the
WhenDone callback.

### DIFF
--- a/cartographer/mapping_2d/sparse_pose_graph.h
+++ b/cartographer/mapping_2d/sparse_pose_graph.h
@@ -172,7 +172,7 @@ class SparsePoseGraph : public mapping::SparsePoseGraph {
       GUARDED_BY(mutex_);
 
   // How our various trajectories are related.
-  mapping::TrajectoryConnectivity trajectory_connectivity_ GUARDED_BY(mutex_);
+  mapping::TrajectoryConnectivity trajectory_connectivity_;
 
   // We globally localize a fraction of the scans from each trajectory.
   std::unordered_map<int, std::unique_ptr<common::FixedRatioSampler>>
--- a/cartographer/mapping_3d/sparse_pose_graph.cc
+++ b/cartographer/mapping_3d/sparse_pose_graph.cc
@@ -210,7 +210,7 @@ void SparsePoseGraph::ComputeConstraint(const mapping::NodeId& node_id,
     constraint_builder_.MaybeAddGlobalConstraint(
         submap_id, submap_data_.at(submap_id).submap.get(), node_id,
         trajectory_nodes_.at(node_id).constant_data.get(), submap_nodes,
-        initial_relative_pose.rotation(), &trajectory_connectivity_);
+        initial_relative_pose.rotation());
   } else {
     const bool scan_and_submap_trajectories_connected =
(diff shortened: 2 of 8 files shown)


cartographer/mapping_3d/sparse_pose_graph.cc
```diff
@@ -328,6 +328,19 @@ void SparsePoseGraph::HandleScanQueue() {
           common::MutexLocker locker(&mutex_);
           constraints_.insert(constraints_.end(), result.begin(), result.end());
         }
+
+        // Update the trajectory connectivity structure with the new
+        // constraints.
+        for (Constraint constraint : result) {
+          int trajectory_id_node = constraint.node_id.trajectory_id;
```
wohe: Inline since there is a single use? Otherwise, this should be `const`.

cartographer/mapping_3d/sparse_pose_graph.cc
```diff
@@ -328,6 +328,19 @@ void SparsePoseGraph::HandleScanQueue() {
           common::MutexLocker locker(&mutex_);
           constraints_.insert(constraints_.end(), result.begin(), result.end());
         }
+
+        // Update the trajectory connectivity structure with the new
+        // constraints.
+        for (Constraint constraint : result) {
```
wohe: `const Constraint&`?

cartographer/mapping_3d/sparse_pose_graph.cc
```diff
@@ -328,6 +328,19 @@ void SparsePoseGraph::HandleScanQueue() {
...
+          int trajectory_id_node = constraint.node_id.trajectory_id;
+          int trajectory_id_submap = constraint.submap_id.trajectory_id;
+          if (constraint.tag ==
```
wohe: You could make this a `CHECK`. The constraint builder only adds `INTER_SUBMAP` constraints.

cartographer/mapping_3d/sparse_pose_graph.cc
cschuet: Done

cartographer/mapping_3d/sparse_pose_graph.cc
cschuet: Done

cartographer/mapping_3d/sparse_pose_graph.cc
cschuet: Done

cartographer/mapping_2d/sparse_pose_graph.h
```diff
@@ -172,7 +172,7 @@ class SparsePoseGraph : public mapping::SparsePoseGraph {
       GUARDED_BY(mutex_);
 
   // How our various trajectories are related.
-  mapping::TrajectoryConnectivity trajectory_connectivity_ GUARDED_BY(mutex_);
```
wohe: Did you consider retaining this and making `mapping::TrajectoryConnectivity` thread-compatible?

PR #509, opened 2017-09-07
Remove connected_components_ from SPG.

cschuet:
This PR removes connected_components_ from the SparsePoseGraph. Since
the connectivity structure is no longer updated by the
ConstraintBuilders it is no longer necessary for SPG to keep a copy of
the connected components for data consistency.


cartographer/mapping_3d/sparse_pose_graph.cc
```diff
@@ -536,7 +535,7 @@ transform::Rigid3d SparsePoseGraph::GetLocalToGlobalTransform(
 
 std::vector<std::vector<int>> SparsePoseGraph::GetConnectedTrajectories() {
   common::MutexLocker locker(&mutex_);
```
wohe: Remove since `trajectory_connectivity_` is thread-safe? Same in 2D.

cschuet: Done

PR #510, opened 2017-09-07
Rename TrajectoryConnectivity ConnectedComponents.

cschuet:


cartographer/mapping/connected_components.h
```diff
@@ -21,7 +21,7 @@
 #include <unordered_map>
```
wohe: Update the include guards?

cschuet: Done

PR #480, opened 2017-08-25
Add fixed frame pose constraints to 3D problem.

zjwoody:

PR #511, opened 2017-09-07
First implementation of initial pose

apr8:
This is the first basic implementation of initial pose. I have implemented this similar to what we have in AMCL(for pure localization). I am currently using the following strategy when an initial pose is given. 
1.Need to writer a subscriber in node.cc that would subscribe to /initialpose (eg from:rviz).
2.Whenever we receive an /initialpose we need to first finish the current Trajectory and then run final optimization.
3.Then a new trajectory needs to be started from the pose given by /initialpose and need to make sure it is with respect to map frame.
4.Push this pose all the way back to LocalTrajectoryBuilder class where the PoseEstimate is defined and change the value from identity to this pose.
5.Then pass the current pose estimate from localtrajectorybuilder to the sparse_pose_graph and then constraint_builder and finally to fast correlative scan matcher (MatchFullSubmap)
6.And instead of starting to scan match from the origin(or centre) of the submap, scan match from the current best pose estimate(only when robot is in or near that submap).

I haven't sent a request for  cartographer_ros changes yet. Will do once this is fixed.

Further Work:
1.Delete all submaps from the unwanted trajectory or delete the trajectory itself.
2.Ordering the submaps based on their proximity to the current pose of the scan.

Would love to have your feedback. 

Thanks,
Akshay

### DIFF
--- a/cartographer/mapping/global_trajectory_builder.h
+++ b/cartographer/mapping/global_trajectory_builder.h
@@ -33,6 +33,14 @@ class GlobalTrajectoryBuilder
       : trajectory_id_(trajectory_id),
         sparse_pose_graph_(sparse_pose_graph),
         local_trajectory_builder_(options) {}
+
+  GlobalTrajectoryBuilder(const LocalTrajectoryBuilderOptions& options,
+                          const int trajectory_id,
+                          SparsePoseGraph* const sparse_pose_graph,
+                          const transform::Rigid3d& initialpose_data)
+      : trajectory_id_(trajectory_id),
+        sparse_pose_graph_(sparse_pose_graph),
+        local_trajectory_builder_(options, initialpose_data) {}
   ~GlobalTrajectoryBuilder() override {}
 
   GlobalTrajectoryBuilder(const GlobalTrajectoryBuilder&) = delete;
--- a/cartographer/mapping/map_builder.cc
+++ b/cartographer/mapping/map_builder.cc
@@ -103,6 +103,44 @@ int MapBuilder::AddTrajectoryBuilder(
   return trajectory_id;
 }
 
+int MapBuilder::AddTrajectoryBuilderWithInitialpose(
+    const std::unordered_set<string>& expected_sensor_ids,
+    const proto::TrajectoryBuilderOptions& trajectory_options,
+    common::Time time,
+    const transform::Rigid3d& initialpose_data) {
+  const int trajectory_id = trajectory_builders_.size();
+  if (options_.use_trajectory_builder_3d()) {
+    CHECK(trajectory_options.has_trajectory_builder_3d_options());
+    trajectory_builders_.push_back(
+        common::make_unique<CollatedTrajectoryBuilder>(
+            &sensor_collator_, trajectory_id, expected_sensor_ids,
+            common::make_unique<mapping::GlobalTrajectoryBuilder<
+                mapping_3d::LocalTrajectoryBuilder,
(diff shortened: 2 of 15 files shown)


ojura:
Have you checked out the recently merged fixed pose constraints by @zjwoody (#471, #476, #477, #480)? I think that being able to constrain the beginning of the trajectory is very specific. Using fixed pose constraints, you might be able to constrain any part of the second trajectory with respect to the first trajectory's origin (if I understood that implementation correctly)

Also, I have an experimental pull request (#481) which enables you to directly inject fixed (I called them "manual") constraints into the optimization problem. If you did this for the first node of the new trajectory, you might again be able to accomplish what you want. I wonder what will maintainers have to say.

apr8:
I have checked the recent merge although I started dev work for this before that, so I dont know what the correct way is. However I feel this is not the best way though you can do that in addition to what I have done. I might be wrong. 
1.Constraints first of all are optimized every n scans. In most cases like in the case of service bots or warehouse robots as in our case we cannot afford to get a wrong position of the robot and wait till the pose is optimized. However, one can improve this  by optimizing very frequently, but again you would slow down the process(there are trade offs) which we need not deal with if we do this. 
2.The pose gets optimized mostly dependent on what weights the optimized poses have that were calculated in the residual block. For example having loop closure weight equivalent to the consecutive scan weight we can see the robot jumping around in the map.
I just changed these weights in sparse_pose_graph.lua and ran their bag file(for localization):demo_backpack_2d_localization.launch
loop_closure_translation_weight = 1.1e3,
loop_closure_rotation_weight = 1e3
We see the bot localize well when the weight for these loop closure constraints are high. Drawing similar conclusions for initial pose implementation we might have to give them higher weights, however if there were any other fixed frame poses we might have to change its weight again as we are mostly not that confident about those poses.
3.I see this is not implemented in 2D, however in 3D, fixed frame poses has a residual block added in the optimization problem. Similar to the argument above, having an common data structure might be problematic as we would require the weight for initial pose far more higher than other fixed frame poses(gps etc) as in most cases we know the initial pose of the robot(at docking station or with respect to some ladmark or qrcode etc). If the way you suggest is done then I guess we need to add separate data structure similar to the one you suggested.
4.In this commit I have also changed the point from which the scan matching is done for global scan matcher(in MatchFullSubmap). Before the scan matching is started from the centre of the submap but now as we already have a good estimate from the initial pose we can scan match from that pose. I havent really decreased the search radius in this commit. But we can do this. Moreover, it wouldnt matter when we are trying to scan match in the wrong submap. I guess this feature much be useful and also ordering the submaps with respect to its proximity to the current scan as it would help find a match faster when we have large umber of submaps. For example if we have 100 submaps and we start in the 100th submap. It might take some time for us before we can scan match against that submap. Current if you can see in the ComputeConstraintsForScan method in sparse pose graph we are starting from submap 0 and then moving to submap 100. If we know we are already at submap 100, we can start to compute constraints from there directly instead.
I wrote this drawing similarity to how it is done AMCL where we re-initialize the pf everytime we give an initial pose. In our case we cannot afford to wait for the robot to localize as that would completely change the path that we generated for it to implement tasks or recoveries. There are multiple drawbacks in the robot localization is bad even for a sec. We have a setup such that everytime the localization is bad or unknown initial, the initial pose feature localizes it at best possible pose immediately.

Thank you so much for your feedback. Would love to hear your thoughts on this.

apr8:
I signed it!

cschuet:
Merged #606. Closing this one.

PR #514, opened 2017-09-08
remove hooray. it spams during pure localization

jihoonl:
I would like to remove this info since it spams in pure localization mode. 

Or could you suggest any other way of disabling this log? I don't want to set `minloglevel` higher because it would also disable logs that I set in my custom cartographer_ros.


ojura:
what a killjoy pull request :-)

wohe:
I'd also be a bit sad to see this go. If you really want to change it, we should make sure that debugging does not become harder. This is right now the best signal whether the background processing is running at real-time. You could switch it to output the `work_queue_` length if background processing falls behind. What do you think?

damonkohler:
@jihoonl, I think @wohe's suggestion is a good one. Do you plan to implement that or should we drop this PR?

jihoonl:
Outputing the length of `work_queue_` if it falls behind sounds better to me. Ok. I will update the PR as you suggested.

Another question, I see [this logging](https://github.com/googlecartographer/cartographer/blob/master/cartographer/mapping/collated_trajectory_builder.cc#L71-L77) also outputs pretty often. Do you have any good idea to disable this log? I wanted to make it as `ROS_DEBUG` but _glog_ doesn't seem to support this level.

jihoonl:
commits got messed up while rebasing. Should I recreate the PR?

cschuet:
Someone more knowledgeable about git than me could probably figure out how to repair this, but as far as I am concerned you can just create a new PR. I don't think there is too much context that we are loosing.

jihoonl:
Replaced by #547

ojura:
No need for a new PR, you can just force push to this branch. You should probably rebase everything on top of the current upstream master. I can assist with anything git related if anyone needs help :)

PR #529, opened 2017-09-14
fix odometry delta calculation

jihoonl:
I am not sure you intentionally have new -> old delta.


wohe:
This was intentional. We also use it for the translational part and the way the code is written we can only easily extrapolate the tracking frame orientation at the newer odometry time. The time delta is then also swapped so that we do not have to invert the value.

jihoonl:
ok. thanks for the explanation

Issue #515, opened 2017-09-08
Drastic decrease in quality after #499


ojura:
This is using the same dataset as in #453, which was another regression that caused a drastic decrease in quality on that dataset.

I did a git bisect:
```
3948943b6491ad07dd4c6ba8e71cb4d03b27a5bf is the first bad commit
commit 3948943b6491ad07dd4c6ba8e71cb4d03b27a5bf
Author: Wolfgang Hess <whess@google.com>
Date:   Mon Sep 4 11:14:51 2017 +0200

    Use EstimateGravityOrientation() in 2D SLAM. (#499)
```

Here's with e3b6f0a:
![e3b6f0a](https://user-images.githubusercontent.com/9062353/30221686-88a8f988-94c4-11e7-8ae6-f2cfbb8eeba9.png)

...and with 3948943:
![3948943](https://user-images.githubusercontent.com/9062353/30221693-90b64752-94c4-11e7-8313-b0c6daaa9192.png)

[Here is the updated `pioneer-cartographer.lua`](https://gist.github.com/ojura/e1bfeef9216e9bec0ba8abcfa24ebe4f) (please use this one instead of the one in #453). Also, please use `cartographer_ros` with rangefinder decimation (if it doesn't get merged by the time you are reading this): googlecartographer/cartographer_ros@e8962a7.

jihoonl:
We are also investigating the same problem at the moment. The orientation becomes very strange when `use_odometry=true`.

ojura:
BTW, bisecting is really made difficult by the fact that there are two separate repos. I know it's a long shot, but would you be willing to consider moving everything to one repo? In the very probable case that you won't, in order to facilitate bisecting, I was thinking of making a script that would prepare a "zipped" linear repo with commits from both `cartographer` and `cartographer_ros` ordered by time of commiting.

ojura:
I have also observed a minor decrease in quality also after 0ef372d ([before](https://user-images.githubusercontent.com/9062353/30225110-934105ea-94d1-11e7-89a6-3a0ce882b107.png), [after](https://user-images.githubusercontent.com/9062353/30225117-97517480-94d1-11e7-84cd-51da5b866c22.png)). But I guess that's a tuning question, and I can open another issue for that after resolve this one, which is far more serious.

ojura:
[...]

PR #530, opened 2017-09-14
use oldest and newest instead of last two odometry data

jihoonl:
it reduces some hiccup behavior while local trajectory building. I think using delta between last two poses were not able to compute proper linear and angular velocity when odom is published in 100hz.

I referenced [angular_velocity_from_poses](https://github.com/googlecartographer/cartographer/blob/master/cartographer/mapping/pose_extrapolator.cc#L140).


jihoonl:
This might be relevant to #515

PR #532, opened 2017-09-15
Apply gravity alignment to odometry data passed to SPG

### LINKED ISSUE #515: Drastic decrease in quality after #499

ojura:
Fixes #515.

I am not sure if this also belongs in 3D (I just mirrored the change that I did in 2D).

### DIFF
--- a/cartographer/mapping_2d/local_trajectory_builder.cc
+++ b/cartographer/mapping_2d/local_trajectory_builder.cc
@@ -198,14 +198,19 @@ void LocalTrajectoryBuilder::AddImuData(const sensor::ImuData& imu_data) {
   extrapolator_->AddImuData(imu_data);
 }
 
-void LocalTrajectoryBuilder::AddOdometerData(
+std::unique_ptr<sensor::OdometryData> LocalTrajectoryBuilder::AddOdometerData(
     const sensor::OdometryData& odometry_data) {
   if (extrapolator_ == nullptr) {
     // Until we've initialized the extrapolator we cannot add odometry data.
     LOG(INFO) << "Extrapolator not yet initialized.";
-    return;
+    return nullptr;
   }
+  const transform::Rigid3d gravity_alignment = transform::Rigid3d::Rotation(
+      extrapolator_->EstimateGravityOrientation(odometry_data.time));
   extrapolator_->AddOdometryData(odometry_data);
+  // Return gravity aligned odometry data.
+  return common::make_unique<sensor::OdometryData>(sensor::OdometryData{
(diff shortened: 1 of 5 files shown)


cartographer/mapping_2d/local_trajectory_builder.cc
```diff
@@ -198,14 +198,19 @@ void LocalTrajectoryBuilder::AddImuData(const sensor::ImuData& imu_data) {
...
+std::unique_ptr<sensor::OdometryData> LocalTrajectoryBuilder::AddOdometerData(
     const sensor::OdometryData& odometry_data) {
   if (extrapolator_ == nullptr) {
     // Until we've initialized the extrapolator we cannot add odometry data.
     LOG(INFO) << "Extrapolator not yet initialized.";
-    return;
+    return nullptr;
   }
+  const transform::Rigid3d gravity_alignment = transform::Rigid3d::Rotation(
+      extrapolator_->EstimateGravityOrientation(odometry_data.time));
```
ojura: Should this be called before or after adding the odometry data to the extrapolator?

wohe:
Thanks for the PR, this definitely goes into the right direction. As far as the concrete implementation goes, I think there is a different approach more in line with the rest of the codebase. I'll prepare a PR together with @cschuet soon.

1.  3D does not have any issue and should not be changed. In the 3D pose graph all poses are in tracking frame.
2. The approach you take to change the odometry data is brittle. Since we do not want to make guarantees about the yaw, interpolating (which we later do in the pose graph optimization) this changed odometry data might introduce errors.
3. A better approach would be to take the gravity alignment of the poses in the 2D pose graph optimization into account when computing the predicted relative change from one pose to the next. We'll look into implementing this now.

PR #533, opened 2017-09-15
Fix odometry in the 2D pose graph optimization.

### LINKED ISSUE #515: Drastic decrease in quality after #499

wohe:
Before, the relative change in pose due to odometry in the
2D pose graph optimization problem was incorrectly done in
the tracking frame. Instead, the gravity aligned frame in
which the 2D poses are stored must be used.

Fixes #515.

PAIR=cschuet

### DIFF
--- a/cartographer/mapping_2d/sparse_pose_graph.cc
+++ b/cartographer/mapping_2d/sparse_pose_graph.cc
@@ -250,9 +250,10 @@ void SparsePoseGraph::ComputeConstraintsForScan(
                                  ->first +
                              1)
           : 0};
-  const auto& scan_data = trajectory_nodes_.at(node_id).constant_data;
-  optimization_problem_.AddTrajectoryNode(
-      matching_id.trajectory_id, scan_data->time, pose, optimized_pose);
+  const auto& node_data = trajectory_nodes_.at(node_id).constant_data;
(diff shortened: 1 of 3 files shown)


wohe:
@ojura Could you please verify that this fixes the quality issues you observed?

Mofef:
For us this fixes the problem. Thanks!

PR #537, opened 2017-09-18
Fix typo in comment

chantivlad:
Just an initial commit.


ojura:
`iff` is an abbreviation for "if and only if"

chantivlad:
Thanks, i was not aware.

PR #540, opened 2017-09-26
Rename serialization NodeData proto for consistency

ojura:
I think that node_data should be a member of a node, not the other way around.

By the way, I think that Cartographer's code here is a bit messy at times. `SparsePoseGraph` has lots of vectors keyed by trajectory ID, instead of encapsulating everything in a class e.g. `Trajectory` and having a single vector of those. It's also a bit confusing that some classes are deserialized by passing a proto to their constructor, while others using `FromProto` (is one for structs, and other for constructed classes?). Some classes handle their own serialization, while others do not (e.g. `SparsePoseGraph`'s nodes and submaps are serialized by `MapBuilder`). It made sense that poses of trajectory nodes were serialized as a part of the sparse pose graph, and constant data separately in the proto stream, but the naming was really confusing, so I attempted to correct that in this PR.


wohe:
Regarding your comments: If there are changes that clearly improve readability, please send PRs.

Let's not bikeshed, though. Having data keyed by trajectory ID for example has certain advantages: We can pass around only some of the data, it allows us to improve readability using abstractions like `NestedVectorsById`.

Regarding free-standing or member functions: This is simply following Google-style and due to the fact that some of the data are simple structs.

ojura:
Sorry if the comments in this PR came as bit abrasive. At that point I was just getting started on serialization.

About the code style, just to confirm - classes have a `ToProto` method and a proto constructor, while for structs there is an overload of `FromProto` and `ToProto`?

damonkohler:
Correct. Google style requires that structs have no methods.

PR #541, opened 2017-09-26
Deserialize trajectory nodes, support multiple trajectories in LoadMap

ojura:
~~Depends on #540~~ (merged).


cartographer/mapping/trajectory_node.h
```diff
@@ -54,6 +54,7 @@ struct TrajectoryNode {
   // node is being trimmed, it must survive until all use finishes.
   std::shared_ptr<const Data> constant_data;
 
+  const transform::Rigid3d& initial_pose;
```
wohe: This does not belong here. It does not take part in the optimization, so it should be part of the `constant_data`. Similar for the proto. Also: Do not keep a dangling reference, this does not work.

ojura: Oops, the reference was a typo. The initial pose _is_ used in the optimization problem: https://github.com/googlecartographer/cartographer/blob/master/cartographer/mapping_2d/sparse_pose_graph/optimization_problem.cc#L239

wohe: Used yes, but not as an optimization variable: It is similar to e.g. `gravity_alignment`. It is taken into account but not allowed to change.

wohe:
Also, can you split off parts of this PR to get it merged quicker, whatever you would like to see merged first, maybe
1. The change to the serialization of `initial_pose`?
2. Adding deserialization of nodes. Maybe include both 2D and 3D, then.
3. Change `LoadMap()` to put things into separate trajectories. Should probably be a separate function `DeserializeState()`?
What do you think?

ojura:
3. is already incoming in the last PR, when everything is ready. Will do 1 and 2 as separate PRs.

ojura:
Okay, I'm closing this pull request. It has been superceded by #567, #568 and #569.

Issue #433, opened 2017-07-27
Serialized SparsePoseGraphs can easily be too big for default proto deserialization


macmason:
I'm working with some _very large_ multi-trajectory SLAM problems (10s of kms, indoor), and the resulting serialized `SparsePoseGraph` objects are too big to deserialize, as they're much larger than the default 64MB limit. (Seriously: I'm up to about 1.3GB for the SPG alone.)

There are three solutions:

1. Modify the serialize of SPG to use the pbstream, writing the pieces out one-by-one.
2. Improve pbstream to use a `google::protobuf::io::CodedInputStream` directly, so we can call `SetTotalBytesLimit` with a bigger value.
3. "Mac, you're out of your mind, make yourself a branch and leave the rest of us out of this."

(I've currently implemented 3, via 2, in `macmason/cartographer` and `macmason/cartographer_ros`, in branches both called `mass_map`.)

Thoughts?

FYI @brandon-northcutt

SirVer:
Are we talking about [this proto](https://github.com/googlecartographer/cartographer/blob/master/cartographer/mapping/proto/sparse_pose_graph.proto#L32)? Can you figure out what exactly takes up all that space?

Because: having a SPA this big means that sometime before a computer actually had to keep all these constraints (that should be of similar size in memory as in proto) in memory and ran an optimization. This seems like it should take years to optimize - that does not seem too practical.

macmason:
Yes, that proto.

The problem is that I have a large number of trajectories, each revisiting the same building, in roughly the same order. As a result, I have _lots_ of loop closures. (The map  looks great!)

For example, my most-recent pbstream serializes 53 trajectories and has 13,206,458 constraints. Optimization takes hours and hours, as expected.

That sounds like a vote for option 3. :-D

wohe:
Longer term, option 1 seems the most reasonable to me. In a first step we could write a `Trajectories` proto containing all trajectories as the first message to minimize the diff. We would have multiple messages containing `Constraints`.

Most interesting change would seem to be necessary in `autogenerate_ground_truth_main.cc` which so far is the only place where we are deserializing constraints.

PR #544, opened 2017-09-27
Deserialize SPG constraints

ojura:
Depends on ~~#541~~ ~~#569~~ and ~~#567~~ (merged).

Also, I would prefer if we didn't have to multiply by the gravity alignment inverse during deserialization. Consider removing it from serialization of constraints?

### DIFF
--- a/cartographer/mapping/map_builder.cc
+++ b/cartographer/mapping/map_builder.cc
@@ -174,11 +174,12 @@ void MapBuilder::SerializeState(io::ProtoStreamWriter* const writer) {
 }
 
 void MapBuilder::LoadMap(io::ProtoStreamReader* const reader) {
-  proto::SparsePoseGraph pose_graph;
-  CHECK(reader->ReadProto(&pose_graph));
+  const std::shared_ptr<proto::SparsePoseGraph> pose_graph =
+      std::make_shared<proto::SparsePoseGraph>();
+  CHECK(reader->ReadProto(pose_graph.get()));
 
   std::map<int, int> trajectory_remapping;
-  for (auto& trajectory_proto : *pose_graph.mutable_trajectory()) {
+  for (auto& trajectory_proto : *pose_graph->mutable_trajectory()) {
     const int new_trajectory_id = AddTrajectoryForDeserialization();
     CHECK(trajectory_remapping
               .emplace(trajectory_proto.trajectory_id(), new_trajectory_id)
@@ -189,7 +190,7 @@ void MapBuilder::LoadMap(io::ProtoStreamReader* const reader) {
   }
 
   // Apply the calculated remapping to constraints in the SparsePoseGraph proto
-  for (auto& constraint_proto : *pose_graph.mutable_constraint()) {
+  for (auto& constraint_proto : *pose_graph->mutable_constraint()) {
     constraint_proto.mutable_submap_id()->set_trajectory_id(
         trajectory_remapping.at(constraint_proto.submap_id().trajectory_id()));
     constraint_proto.mutable_node_id()->set_trajectory_id(
@@ -197,7 +198,7 @@ void MapBuilder::LoadMap(io::ProtoStreamReader* const reader) {
   }
 
   MapById<SubmapId, transform::Rigid3d> submap_poses;
-  for (const proto::Trajectory& trajectory_proto : pose_graph.trajectory()) {
+  for (const proto::Trajectory& trajectory_proto : pose_graph->trajectory()) {
     for (const proto::Trajectory::Submap& submap_proto :
          trajectory_proto.submap()) {
       submap_poses.Insert(SubmapId{trajectory_proto.trajectory_id(),
@@ -207,7 +208,7 @@ void MapBuilder::LoadMap(io::ProtoStreamReader* const reader) {
   }
 
   MapById<NodeId, transform::Rigid3d> node_poses;
-  for (const proto::Trajectory& trajectory_proto : pose_graph.trajectory()) {
+  for (const proto::Trajectory& trajectory_proto : pose_graph->trajectory()) {
     for (const proto::Trajectory::Node& node_proto : trajectory_proto.node()) {
       node_poses.Insert(
           NodeId{trajectory_proto.trajectory_id(), node_proto.node_index()},
--- a/cartographer/mapping/map_builder.cc
+++ b/cartographer/mapping/map_builder.cc
@@ -188,14 +188,6 @@ void MapBuilder::LoadMap(io::ProtoStreamReader* const reader) {
     sparse_pose_graph_->FreezeTrajectory(new_trajectory_id);
   }
 
-  // Apply the calculated remapping to constraints in the SparsePoseGraph proto
-  for (auto& constraint_proto : *pose_graph.mutable_constraint()) {
-    constraint_proto.mutable_submap_id()->set_trajectory_id(
-        trajectory_remapping.at(constraint_proto.submap_id().trajectory_id()));
-    constraint_proto.mutable_node_id()->set_trajectory_id(
-        trajectory_remapping.at(constraint_proto.node_id().trajectory_id()));
-  }
-
   MapById<SubmapId, transform::Rigid3d> submap_poses;
   for (const proto::Trajectory& trajectory_proto : pose_graph.trajectory()) {
     for (const proto::Trajectory::Submap& submap_proto :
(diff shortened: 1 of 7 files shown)


cartographer/mapping_2d/sparse_pose_graph.cc
```diff
@@ -426,6 +426,39 @@ void SparsePoseGraph::AddNodeFromProto(const transform::Rigid3d& global_pose,
...
+          constraint_proto.rotation_weight()};
+      const Constraint::Tag tag = mapping::FromProto(constraint_proto.tag());
+
+      CHECK(trajectory_nodes_.Contains(node_id));
+      CHECK(submap_data_.Contains(submap_id));
+      CHECK(trajectory_nodes_.at(node_id).constant_data != nullptr);
+      CHECK(submap_data_.at(submap_id).submap != nullptr);
+
+      if (tag == Constraint::Tag::INTRA_SUBMAP) {
+        submap_data_.at(submap_id).node_ids.emplace(node_id);
```
ojura: CHECK(emplace().second)?

cartographer/mapping/map_builder.cc
```diff
@@ -174,11 +174,12 @@ void MapBuilder::SerializeState(io::ProtoStreamWriter* const writer) {
 }
 
 void MapBuilder::LoadMap(io::ProtoStreamReader* const reader) {
-  proto::SparsePoseGraph pose_graph;
-  CHECK(reader->ReadProto(&pose_graph));
+  const std::shared_ptr<proto::SparsePoseGraph> pose_graph =
```
wohe: Using a `shared_ptr` to avoid copy of this comparatively small amount of data seems like premature optimization, especially given that we copy the constraints in other parts of the code.

cartographer/mapping/map_builder.cc
```diff
@@ -230,6 +239,7 @@ void MapBuilder::LoadMap(io::ProtoStreamReader* const reader) {
     }
   }
   CHECK(reader->eof());
+  sparse_pose_graph_->AddConstraintsFromProto(pose_graph);
```
wohe: Pass along only the repeated constraints? The other parts of the `pose_graph` are not used by `AddConstraintsFromProto()`, right?

cartographer/mapping/sparse_pose_graph.h
```diff
@@ -120,6 +125,9 @@ class SparsePoseGraph {
   virtual std::vector<Constraint> constraints() = 0;
 };
 
+SparsePoseGraph::Constraint::Tag FromProto(
```
wohe: How about exposing instead a function that creates a vector of `SparsePoseGraph::Constraint` from the proto?

cartographer/mapping/map_builder.cc
```diff
@@ -188,8 +189,16 @@ void MapBuilder::LoadMap(io::ProtoStreamReader* const reader) {
     sparse_pose_graph_->FreezeTrajectory(new_trajectory_id);
   }
 
+  // Apply the calculated remapping to constraints in the SparsePoseGraph proto
+  for (auto& constraint_proto : *pose_graph->mutable_constraint()) {
```
wohe: When the map is frozen deserializing the constraints provides no benefit but comes at a cost regarding performance and memory. Either make it possible to load unfrozen maps and conditionally deserialize constraints in this case. Or first introduce a function for deserializing the state, i.e. not intended to provide a frozen map but to continue SLAMming. What do you think?

cartographer/mapping/map_builder.cc
ojura: According to #433, it can be quite big.

cartographer/mapping/map_builder.cc
ojura: Not yet, but they will. `SPG::AddConstraintsFromProto` gets renamed to `AddDataFromProto` in the next PR (#548), and loading IMU data, odometry and trajectory connectivity will also be added to this function in #548, #550 and #552.

cartographer/mapping/sparse_pose_graph.h
ojura: In other words, you want to extract [lines 433-446](https://github.com/googlecartographer/cartographer/pull/544/files#diff-8148ee0da898a26c8b927bc8079d8900R433) into a separate function?

cartographer/mapping/map_builder.cc
ojura: That is true. In #554 `LoadMap` gets renamed to `LoadState` and gets a boolean parameter whether to freeze or not. Could we delay this optimization until then? Alternatively, we can add the parameter now, skip deserialization if `frozen = true`, and quit if `frozen = false` because it's not ready yet.

cartographer/mapping/map_builder.cc
cschuet: Right, but using a `shared_ptr` here does not really help with that problem. If we wanted to properly support the use case of deserializing 100s of MBs of constraints we would need to move the constraints out of the `proto::SparsePoseGraph` and support loading them in chunks. So let's do a copy here.

BTW personally I would only use a `shared_ptr` if the lifetime of the object is known only at runtime. I would always opt for `unique_ptr`.

cartographer/mapping/map_builder.cc
cschuet: Ok, so then let's move this block of code over to #554.

cartographer/mapping/sparse_pose_graph.h
cschuet: Yes. I find it always hard to read overly long closures.

cartographer/mapping_2d/sparse_pose_graph.cc
cschuet: Sure. Why not.

cartographer/mapping/map_builder.cc
cschuet: Ok, let's refactor it then. Let's pass only the constraints here.

cartographer/mapping/map_builder.cc
ojura: You can't easily move a `unique_ptr` into a lambda in C++11. You would have to release it and capture a regular pointer. And we don't exactly know how long the function is going to sit around on the worker queue (does this fit your definition of knowing it only at runtime)?

But sure, I can change it here to copying the SPG proto when adding the function to the worker queue, if you prefer.

cschuet: Cool. Let's do that and get this thing merged.

cartographer/mapping/map_builder.cc
ojura: So, you want me to pass an
`::google::protobuf::RepeatedPtrField< ::cartographer::mapping::proto::SparsePoseGraph_Constraint >` here?

ojura: How do you want me to pass only the constraints? Pass a `::google::protobuf::RepeatedPtrField< ::cartographer::mapping::proto::SparsePoseGraph_Constraint >`?

cartographer/mapping/sparse_pose_graph.h
ojura: Creating a constraint is not possible to express as a pure function. It requires access to per-node `gravity_alignment` for 2D (see line [442](https://github.com/googlecartographer/cartographer/pull/544/files#diff-8148ee0da898a26c8b927bc8079d8900R442)).

ojura:
Updated, PTAL

cartographer/mapping/sparse_pose_graph.cc
```diff
@@ -35,6 +35,40 @@ proto::SparsePoseGraph::Constraint::Tag ToProto(
...
+  LOG(FATAL) << "Unsupported tag.";
+}
+
+std::vector<SparsePoseGraph::Constraint> FromProto(
+    const ::google::protobuf::RepeatedPtrField<
+        ::cartographer::mapping::proto::SparsePoseGraph::Constraint>&
+        constraint_protos) {
+  std::vector<SparsePoseGraph::Constraint> constraints;
+  for (const auto& constraint_proto : constraint_protos) {
+    const mapping::SubmapId submap_id{
```
ojura: Perhaps we should `CHECK` components of `node_id` and `submap_id` for being `>= 0` here.

cschuet: MapById already checks that, so let's not add redundant checks.

ojura:
Ok, sure. As I mentioned, I plan to rename `AddSerializedConstraints` to `AddDataFromProto` in #548. I can write a test after we complete reviewing that PR, so that I don't have to change the test.

cschuet:
SG thanks.

PR #548, opened 2017-09-28
Serialize IMU data in SPG

ojura:
Depends on ~~#544~~ (merged).

~~~This replaces ~~`AddConstraintsFromProto`~~ `AddSerializedConstraints` (from #544) with `AddDataFromProto`, so that all functionality concerning loading optimization data from the SPG proto is in one function (soon to be added is odometry).~~~

### DIFF
--- a/cartographer/mapping/map_builder.cc
+++ b/cartographer/mapping/map_builder.cc
@@ -229,6 +229,22 @@ void MapBuilder::LoadMap(io::ProtoStreamReader* const reader) {
       sparse_pose_graph_->AddSubmapFromProto(submap_pose, proto.submap());
     }
   }
+
+  // Add information about which nodes belong to which submap.
+  for (const proto::SparsePoseGraph::Constraint& constraint_proto :
+       pose_graph.constraint()) {
+    if (constraint_proto.tag() !=
+        mapping::proto::SparsePoseGraph::Constraint::INTRA_SUBMAP) {
+      continue;
+    }
+    const NodeId node_id{
+        trajectory_remapping.at(constraint_proto.node_id().trajectory_id()),
+        constraint_proto.node_id().node_index()};
+    const SubmapId submap_id{
+        trajectory_remapping.at(constraint_proto.submap_id().trajectory_id()),
+        constraint_proto.submap_id().submap_index()};
+    sparse_pose_graph_->AddNodeToSubmap(node_id, submap_id);
+  }
   CHECK(reader->eof());
 }
 
--- a/cartographer/mapping/map_builder.cc
+++ b/cartographer/mapping/map_builder.cc
@@ -23,6 +23,7 @@
 #include <utility>
 
 #include "cartographer/common/make_unique.h"
+#include "cartographer/common/time.h"
 #include "cartographer/mapping/collated_trajectory_builder.h"
 #include "cartographer/mapping/global_trajectory_builder.h"
 #include "cartographer/mapping_2d/local_trajectory_builder.h"
@@ -100,6 +101,14 @@ int MapBuilder::AddTrajectoryBuilder(
     sparse_pose_graph_->AddTrimmer(common::make_unique<PureLocalizationTrimmer>(
         trajectory_id, kSubmapsToKeep));
   }
+  if (trajectory_options.has_initial_trajectory_pose()) {
+    const auto& initial_trajectory_pose =
+        trajectory_options.initial_trajectory_pose();
+    sparse_pose_graph_->SetInitialTrajectoryPose(
+        trajectory_id, initial_trajectory_pose.to_trajectory_id(),
+        transform::ToRigid3(initial_trajectory_pose.relative_pose()),
+        common::FromUniversal(initial_trajectory_pose.timestamp()));
+  }
   return trajectory_id;
 }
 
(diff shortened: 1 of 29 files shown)


ojura:
Okay, this should be ready again (@wohe merged in `MapByTime` for imu data in #631 , so I had to follow here).

cartographer/mapping/proto/trajectory.proto
```diff
@@ -48,4 +49,7 @@ message Trajectory {
 
   // Submaps associated with the trajectory.
   repeated Submap submap = 2;
+
+  // Time-ordered IMU data.
+  repeated sensor.proto.ImuData imu_data = 4;
```
wohe: The `Trajectory` message is contained in the `SparsePoseGraph` proto of which we have a single one at the beginning of each `.pbstream`. Putting a lot of data into a single proto message is not a good idea, so we should not put this here. Put in `SerializedData`?

cartographer/mapping/sparse_pose_graph.cc
```diff
@@ -131,6 +131,14 @@ proto::SparsePoseGraph SparsePoseGraph::ToProto() {
         transform::ToProto(submap_id_data.data.pose);
   }
 
+  const auto all_imu_data = GetImuData();
+  for (const int trajectory_id : all_imu_data.trajectory_ids()) {
+    for (const auto& imu_data : all_imu_data.trajectory(trajectory_id)) {
+      auto* imu_data_proto = trajectory(trajectory_id)->add_imu_data();
```
wohe: No need for the local variable?

cartographer/mapping/sparse_pose_graph.h
```diff
@@ -87,10 +89,9 @@ class SparsePoseGraph {
   virtual void AddNodeFromProto(const transform::Rigid3d& global_pose,
                                 const proto::Node& node) = 0;
 
-  // Adds serialized constraints. The corresponding trajectory nodes and submaps
-  // have to be deserialized before calling this function.
-  virtual void AddSerializedConstraints(
-      const std::vector<Constraint>& constraints) = 0;
+  // Adds constraints and IMU data from a proto. The corresponding trajectory
+  // nodes and submaps have to be deserialized before calling this function.
+  virtual void AddDataFromProto(const proto::SparsePoseGraph& proto) = 0;
```
wohe: This is not a great interface since it combines two orthogonal concerns: constraints and IMU data. `AddImuData()` already exists, so there is also no need to change `AddSerializedConstraints()`.

ojura: My reasoning was the following: the sparse pose graph owns the optimization problem and the IMU data contained therein, so I thought that `SPG::ToProto` should be responsible for serializing that. Following that, since we would be serializing everything (constraints and imu data and later odometry) in `SPG::ToProto`, I thought that the single inverse function (resembling `FromProto`) should also handle everything, so I proposed the rename.

Another concern was that if I use `AddImuData` to load all IMU messages, this would be slower since there's a really big number of them and we would need to push a work item for each one.

I considered storing it into the proto stream instead, as you wrote in the review, but I thought that since we're already storing all node poses inside the SPG proto, IMU data wouldn't make much of a difference. But sure, if you would prefer to store it in the proto stream, we can do that as well.

wohe: Yes, please update to store IMU data in the proto stream. We rather want to split the current `SparsePoseGraph` message into multiple messages, not add more data. There is also much more IMU data (e.g. 250 Hz in our demo bags) usually then poses. You might run into the proto message default limit for long demo bags already.

ojura:
Updated.

cartographer/mapping/map_builder.cc
```diff
@@ -237,6 +250,12 @@ void MapBuilder::LoadMap(io::ProtoStreamReader* const reader) {
                                    proto.submap().submap_id().submap_index()});
       sparse_pose_graph_->AddSubmapFromProto(submap_pose, proto.submap());
     }
+    if (proto.has_imu_data()) {
```
wohe: No code in comments, please. At best this adds to the maintenance burden, at worst it gets out of sync. Remove here and add in the appropriate follow up PR?

cartographer/mapping/proto/serialization.proto
```diff
@@ -31,8 +32,14 @@ message Node {
   optional TrajectoryNodeData node_data = 5;
 }
 
+message ImuData {
+  optional int32 trajectory_id = 1;
+  optional cartographer.sensor.proto.ImuData imu_data = 2;
```
wohe: Remove the redundant `cartographer.` prefix?

cartographer/mapping/sparse_pose_graph.h
```diff
@@ -135,6 +137,13 @@ class SparsePoseGraph {
   // Returns the collection of constraints.
   virtual std::vector<Constraint> constraints() = 0;
 
+  // Returns the IMU data.
```
wohe: Consistency with the implementations: Move right before `constraints()`.

cartographer/mapping/sparse_pose_graph.h
```diff
@@ -135,6 +137,13 @@ class SparsePoseGraph {
...
+  virtual sensor::MapByTime<sensor::ImuData> GetImuData() = 0;
+
+  // Inserts an IMU measurement.
```
wohe: Similar here: Move above `FreezeTrajectory()`?

ojura:
@wohe Updated.

PR #550, opened 2017-09-28
Serialize and deserialize odometry data in SPG

ojura:
Depends on #548.

### DIFF
--- a/cartographer/transform/proto/transform_interpolation_buffer.proto
+++ b/cartographer/transform/proto/transform_interpolation_buffer.proto
@@ -1,0 +1,28 @@
+// Copyright 2017 The Cartographer Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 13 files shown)


cartographer/transform/proto/transform_interpolation_buffer.proto
```diff
@@ -0,0 +1,28 @@
...
+// limitations under the License.
+
+syntax = "proto2";
+
+package cartographer.transform.proto;
+
+import "cartographer/transform/proto/transform.proto";
+
+message TransformInterpolationBuffer {
+  message StampedTransform {
```
ojura: There is already a NodeData proto in the sensor package, but I think it semantically makes sense to introduce a stamped transform proto for serializing the interpolation buffer.

ojura:
@wohe OK, this waits until odometry is refactored.

PR #551, opened 2017-09-28
fix ceres dependency name to match rosdep key

mikaelarguedas:
use `libceres-dev` as it is the key defined in the rosdep database for the ceres-solver package: 
https://github.com/ros/rosdistro/blob/0eb70bb82cc6b0dcb613ad8c6e045ab9eee8d24e/rosdep/base.yaml#L1426-L1432

### DIFF
--- a/package.xml
+++ b/package.xml
@@ -37,9 +37,9 @@
   <build_depend>python-sphinx</build_depend>
 
   <depend>boost</depend>
-  <depend>ceres-solver</depend>
   <depend>eigen</depend>
   <depend>libcairo2-dev</depend>
+  <depend>libceres-dev</depend>
   <depend>libgflags-dev</depend>
   <depend>libgoogle-glog-dev</depend>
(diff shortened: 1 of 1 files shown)


damonkohler:
We require a newer version of Ceres than is available on all supported platforms. It's intentionally built from source instead.

damonkohler:
See https://discourse.ros.org/t/ceres-package/872/20

mikaelarguedas:
Thanks @damonkohler for the reminder, I repackaged [ceres-solver 1.12](https://github.com/googlecartographer/cartographer_ros/pull/477) only for [Ubuntu 16.04 and newer](https://discourse.ros.org/t/ceres-package/872/19), so ceres-solver still needs to be built from source on Ubuntu 14.04 and Debian Jessie. I'll keep this change on the release repository then because the rosdep key needs to be resolvable for the package to be built on the ROS buildfarm for ROS Lunar

Issue #549, opened 2017-09-28
Full deserialization from .pbstream task list


ojura:
Things to serialize/deserialize (please add if I have missed anything):

- [x] Submaps (#382)
- [x] Trajectory nodes (~~#541~~ #569)
- [x] Multiple trajectories (#567)
- [x] Constraints (#544)
- [x] IMU data (#548)
- [x] Odometry data (~~#550~~ #666)
- [x] Trajectory connectivity (#552)
- [x] Trajectory builder options (~~#553~~ #859)
- [x] 3D: fixed frame poses (#689)
- [x] Landmarks, landmark observations?

Finally:
- [x] Enable optional loading of unfrozen state (~~#554~~ #891 )

ojura:
@wohe, I would like to ping you to get your input about the big picture of this.

To summarize, it's a linear series of commits which has been broken into multiple PRs in hope that it would get merged faster, but #554 is basically one big PR (currently with a diff of +377 −71) which contains all of the above (although it does not contain #576, which was my attempt at unblocking #567). The final goal of the last PR, #554, is to enable full unfrozen deserialization, but also to retain the possibility of loading frozen trajectories for pure localization (`LoadMap`, renamed to `LoadState`, accepts a boolean parameter whether to freeze loaded trajectories or not).

Since all these PRs form a continous story of full deserialization, I did not think it would make sense to base them all on top of master, in parallel, and then resolve the unavoidable conflicts -- that's why they all are in a chain. Despite that, you can review the relative changesets from one PR to other by viewing the diffs of their respective branches:

#567: https://github.com/googlecartographer/cartographer/compare/master...larics:loadmap-multi-traj
#544: https://github.com/larics/cartographer/compare/loadmap-multi-traj...spg-deserialization-constraints
#548: https://github.com/larics/cartographer/compare/spg-deserialization-constraints...spg-serialization-imudata
#550: https://github.com/larics/cartographer/compare/spg-serialization-imudata...spg-serialization-odometry
#552: https://github.com/larics/cartographer/compare/spg-serialization-odometry...spg-serialization-misc
#553: https://github.com/larics/cartographer/compare/spg-serialization-misc...spg-serialization-trajectorybuilders
#554: https://github.com/larics/cartographer/compare/spg-serialization-trajectorybuilders...load-unfrozen-state

[...]

Issue #249, opened 2017-05-03
Refactor LocalTrajectoryBuilder to distinguish between loaded/finished trajectories and currently building ones


SirVer:

wohe:
Made obsolete by #605.

PR #553, opened 2017-09-29
Serialize and deserialize trajectory builder options

ojura:
The options proto is written as a second proto in the stream, behind the pose graph proto.

### DIFF
--- a/cartographer/mapping/map_builder.cc
+++ b/cartographer/mapping/map_builder.cc
@@ -129,12 +129,24 @@ int MapBuilder::AddTrajectoryBuilder(
         transform::ToRigid3(initial_trajectory_pose.relative_pose()),
         common::FromUniversal(initial_trajectory_pose.timestamp()));
   }
+  proto::TrajectoryBuilderOptionsWithSensorIds options_with_sensor_ids_proto;
+  for (const auto& sensor_id : expected_sensor_ids) {
+    options_with_sensor_ids_proto.add_sensor_id(sensor_id);
+  }
+  *options_with_sensor_ids_proto.mutable_trajectory_builder_options() =
+      trajectory_options;
+  all_trajectory_builder_options_.push_back(options_with_sensor_ids_proto);
+  CHECK_EQ(trajectory_builders_.size(), all_trajectory_builder_options_.size());
   return trajectory_id;
 }
 
-int MapBuilder::AddTrajectoryForDeserialization() {
+int MapBuilder::AddTrajectoryForDeserialization(
+    const proto::TrajectoryBuilderOptionsWithSensorIds&
+        options_with_sensor_ids_proto) {
   const int trajectory_id = trajectory_builders_.size();
   trajectory_builders_.emplace_back();
+  all_trajectory_builder_options_.push_back(options_with_sensor_ids_proto);
+  CHECK_EQ(trajectory_builders_.size(), all_trajectory_builder_options_.size());
   return trajectory_id;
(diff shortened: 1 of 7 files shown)


SirVer:
@ojura Is this something you want to get in? It seems to me that the TrajectoryBuilderOptions are configuration, expressed in Lua and not State. It seems peculiar to me to serialize it as state.

SirVer:
@cschuet @wohe I have no strong feelings either way about this. What do you think?

wohe:
It seems this reverts the fix to #249. Why do we deserialize the options that were once used but no longer are? @ojura I also do not understand why your asset writer wants to reuse the resolution that was used for SLAM. Is this not a separate choice, i.e. why do you not add this as an option to your asset writer?

ojura:
Sure, I can handle the resolution option externally, but that is not the only thing from here that is used: for example, also `RangeDataInserterOptions`.

The submaps that are deserialized into libcartographer were previously built with certain (obviously immutable) `LocalTrajectoryBuilderOptions` (including `RangeDataInserterOptions`). After deserialization, the built submaps are restored, but you don't know which `LocalTrajectoryBuilderOptions` were used for building them, should you need it to inspect it for some reason. (And I do, because I want to reuse all these options for rebuilding the map.)

I can of course handle all of this externally, of course. That's what I'm going to do if the PR doesn't go through :-)

> It seems this reverts the fix to #249.

Yes, because since we have the trajectory builder options data available in the proto stream, and restoring it is very cheap, I don't see a reason not to restore these options.

ojura:
I think I came up with a good analogy: when a camera takes a picture, it records the focal length, aperture size, exposure and ISO. You can't really change these parameters, because this would mean taking another picture. But you _can_ inspect these parameters and make some conclusions on how they influence the quality of the picture.

Cartographer recording the options used for building the trajectory and the submaps would be analogous.

cschuet:
@ojura I have looked at this again and I must say that I am really deeply uncomfortable with deserializing the `TrajectoryBuilderOptions`. I completely get your analogy and I agree that we should add the options when writing the pbstream, so that your pbstream management GUI can display them or whatever you want to do with them, but I wouldn't want to deserialize them in LoadMap.

ojura:
Okay, sure. I will update this to avoid full deserialization.

ojura:
@cschuet I reworked this a little, can you take a look? You didn't like full deserialization of a new trajectory builder using the saved options, so I removed it. I also reverted removing `AddTrajectoryForDeserialization`, and expanded it to just pass through the already previously serialized options.

cartographer/mapping/map_builder.cc
```diff
@@ -227,12 +251,20 @@ void MapBuilder::SerializeState(io::ProtoStreamWriter* const writer) {
 }
 
 void MapBuilder::LoadMap(io::ProtoStreamReader* const reader) {
-  proto::PoseGraph pose_graph;
-  CHECK(reader->ReadProto(&pose_graph));
+  proto::PoseGraph pose_graph_proto;
+  CHECK(reader->ReadProto(&pose_graph_proto));
```
ojura: I think it would make sense to introduce a pbstream serialization header proto, which would include the pose graph and trajectory builder options protos.

ojura:
@cschuet updated. This is blocking #549 for almost two months, can we resolve this?

ojura:
Shoot, I messed up by not overriding the interface method. Will fix and reopen.

PR #554, opened 2017-09-29
Enable loading unfrozen state

ojura:
Depends on #553.

This is the big finish to the serialization series of PRs. If we managed to restore everything properly, we should be able to load the prior state unfrozen and enable full optimization, as if we built all the deserialized data in the current run.


ojura:
Reminder to myself: don't forget to apply the trajectory ID remapping here (https://github.com/googlecartographer/cartographer/pull/544/commits/8df868c9a29bc84a025efa181619256747bb0842).

ojura:
Closing, as this is not ready yet.

Issue #543, opened 2017-09-27
Implementation of an initial pose for new trajectories


Mofef:
Hey, I was searching for a way to initialize new trajectories with a pose estimate.

The behaviour I would like to get is that cartographer searches for constraints in a certain window around  the given the initial pose in order to connect to trajectory 0 (lets say trajectory 0 is enough for now). At some point the SPG should be optimized and the initial pose estimate would then be improved. If no constraint is found inside the window it should either fail or fall back to global localization.

So, what do you think about saving the initial pose as the current pose of node 0 and submap 0 of the new trajectory by saving it to the respective `pose` in `node_data_` and `submap_data_` of the optimization problem? Further I would `TrajectoryConnectivityState::Connect()` the new trajectory to trajectory 0.
This would make respective calls of `SparsePoseGraph::ComputeConstraint()` to fall into the local search case with the given initial pose as the center of the local search window.

This would be an alternative to #511

cschuet:
Thanks, there is already #469. Having prior knowledge about where the robot is and using that to speed up localization definitely is something we want to add.

jihoonl:
@cschuet you just closed this without answering the question... @Mofef is basically asking for the initialpose feature development direction. Please comment at least if you have a different idea of developing initialpose or if it is similar to what you plan on initialpose.

Mofef:
Alternatively I can repost my suggestions there. 
I just opened a new issue, because I wanted to discuss a potential implementation with the goal of preparing a pull request afterwards.

cschuet:
Thanks, Jihoon and Moritz. I am trying to keep the discussion around initial poses contained in the original issue as I would prefer not to have it spread out over multiple.

I had added an (admittedly short) comment on #469 about the direction for initial pose development we are thinking, that I have just expanded.

jihoonl:
Sorry if I was too aggressive. And thanks for the explanation. It explains enough to check if we are moving to the same direction or somewhere else. :)

cschuet:
Oh, I did not have that impression, so all good :) And I think you made a good point in that I should start live streaming more of my thoughts into the tracker even if I am not confident yet that they make sense :)

jihoonl:
Yes. It would be really helpful for us to sync thoughts and plan what features to contribute in cartographer.

Issue #469, opened 2017-08-22
please add support for an initial world pose


BrannonKing:
Please allow us to set the initial value of `pose_estimate_` as discussed here: https://groups.google.com/forum/#!topic/google-cartographer/dqgz9CNCPKA

BrannonKing:
I suppose that exposing the quality of the pose_estimate_ may reduce the need for this feature -- I could ignore pose output until I'm sufficiently certain of my position. However, even after that this may be handy for situations where the startup laser scan is ambiguous (imaging facing a tight corner on startup, etc.)

TysonUQ:
I too would appreciate the ability to set the initial pose.

I can read in a previously saved map, and eventually (after travelling around for a few minutes) locate myself within it.

It would be great to read in the map and tell Cartographer the pose X_0 (which I know) that the platform should be starting from within it, rather than wait for it to work it out.  Alternatively, would it be possible to apply inv(X_0) to the saved map as it is read in, effectively putting the prior around the (0, 0) starting pose?

Entropy512:
I've recently seen some commits referencing the ability to add fixed-frame reference data to the optimization/constraint problem - this might handle the issue of initial pose estimate quality when localizing in an existing map?

cschuet:
Right, we are thinking that fixed-frame poses might also the right way to allow people to inject prior knowledge about the position of the system e.g. when it starts up.

I am hoping to get to that next week.

ojura:
@cschuet check out my Custom constraints PR - it allows adding constraints which involve nodes and submaps that do not yet exist. It might be what you're looking for.

cschuet:
Thanks @ojura I will take a look. In reply to #543 that I just closed to keep the discussion focussed to this issue.

I haven't discussed this with @wohe yet and would like to hear his thoughts but,

* I think fixed frame poses are the right vehicle to allow cartographer clients to inject prior knowledge about where the system is currently located into the problem. Initial pose seems like a special case of that.
* At least for the initial pose use case after having added a fixed frame pose for it, we want local (windowed) searches (at least for a limited time) around that position rather than global searches. If we fail to localize (e.g. because the initial pose was just wrong) we want to be able to recover from that by switching back to global searches.
[...]

Issue #254, opened 2017-05-03
Change the fixed point of Cartographer's map away from being the initial pose of the first submap


SirVer:
/cc @dcconner @brandon-northcutt

This was pulled out from https://github.com/googlecartographer/cartographer/issues/197 to focus discussion. 

In https://github.com/googlecartographer/cartographer/issues/197#issuecomment-298477329 the issue was raised that the robot position jumps whenever loop closure occurs. The suggestion was to update prior submap poses relative to the current robot pose estimate.

ROS solves this differently and Cartographer follows this convention. This is documented [in REP 105](http://www.ros.org/reps/rep-0105.html). The local pose is mapped in a different global frame usually called `odom` and Cartographer can publish this. `odom -> base_link` never jumps and can therefore be used as local pose. `map -> odom` always changes when loop closure happens (and jumps) to keep the robots pose correct in the current world frame.

This is usually sufficient to express the needs of algorithms and observers. My inclination is therefore to close this issue as `wontfix`. Am I missing something here?

Entropy512:
At least from my perspective, the benefit of having the origin of the map somewhere else other than the initial pose of the first robot to begin mapping is less of a benefit for robot navigation and more of a potential human interface/UI perspective if Cartographer is integrated into a higher-level navigation solution that presents a representation of a map to a human - there's likely a strong desire to have the origin point of the map at a location known to be the center (or a corner) of a facility.  There's an even stronger desire to have the x/y coordinate system axes be aligned with the typical horizontal grid structure of most facilities.

One solution to the latter is simply to ensure that the first robot to start mapping the facility starts aligned along the desired axis.  I've done this at least once - but my robot's typical parking spot is at an odd angle that results in a human-unfriendly coordinate system.

Another possibility might be to just publish a transform between "map" and a "human_friendly_map" coordinate system - but then the challenge becomes deciding and designating which submap is the anchor point for this transform.

Currently it's not a big issue since you can just rotate the map image written out by the finish_trajectory service, but once you start doing persistent mapping there could be some UI issues if the map coordinate system's origin isn't easily "anchorable" to a reference point that's human-friendly, whether by choosing an alternate origin coordinate for map, or by publishing a transform between map and that particular anchor point.

SirVer:
> Another possibility might be to just publish a transform between "map" and a "human_friendly_map" coordinate system - but then the challenge becomes deciding and designating which submap is the anchor point for this transform.

[...]

PR #558, opened 2017-10-02
Initial pose for new trajectories

Mofef:
This is a proposal to #469 based on the idea described in #543. There is not yet the enum @cschuet proposed in the discussion in #469.
We are happy to implement this if you consider this PR to be a suitable approach in general. In this case there is also a respective branch ready for cartographer_ros.

### DIFF
--- a/cartographer/mapping_2d/sparse_pose_graph.cc
+++ b/cartographer/mapping_2d/sparse_pose_graph.cc
@@ -62,6 +62,8 @@ std::vector<mapping::SubmapId> SparsePoseGraph::GrowSubmapTransformsAsNeeded(
         submap_data[trajectory_id].empty()) {
       optimization_problem_.AddSubmap(
           trajectory_id,
+          pose_initialization_map.find(trajectory_id) != pose_initialization_map.end() ?
+          transform::Project2D(pose_initialization_map.at(trajectory_id)) :
           sparse_pose_graph::ComputeSubmapPose(*insertion_submaps[0]));
     }
     CHECK_EQ(submap_data[trajectory_id].size(), 1);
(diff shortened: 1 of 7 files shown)


cartographer/mapping_2d/sparse_pose_graph.cc
```diff
@@ -322,6 +325,15 @@ common::Time SparsePoseGraph::GetLatestScanTime(
   return time;
 }
 
+void SparsePoseGraph::SetInitialTrajectoryPose(const int trajectory_id,
+                                               const transform::Rigid3d& pose,
+                                               const common::Time& time) {
+  common::MutexLocker locker(&mutex_);
+  pose_initialization_map[trajectory_id] = pose;
+  trajectory_connectivity_state_.Add(trajectory_id);
+  trajectory_connectivity_state_.Connect(0, trajectory_id, time);
```
jihoonl: Here actually assumes that `Trajectory 0` exist. Should we skip trajectory_connectivity part if the given trajectory is 0?

ojura: Does setting an initial pose for Trajectory 0 make sense? Isn't the global coordinate system actually defined to be the origin of trajectory 0?

jihoonl: It is not necessarily the same. It happened to the same because trajectories always start at (0,0,0).

Mofef: I think throwing an error in that case might also be a viable option. 
@ojura the start of trajectory 0 is so far fixed to zero at least, yes. Also in #254 it was discussed and decided to keep it. I personally don't care, but since the interface for starting a trajectory would change anyway and the alternative option would be to throw an error, we could reconsider this option.

cschuet:
Merged #606. Closing this one.

Issue #560, opened 2017-10-02
Trim all submaps when trajectory finishes if pure localization = true


jihoonl:
If a trajectory is in a pure localization mode, it should remove all submaps when it finishes.

wohe:
I think I'm missing context here. Can you elaborate a bit on what you are trying to achieve? I also do not understand the rationale behind #562.

jihoonl:
Assume that you loaded a frozen trajectory and started a pure localization enabled trajectory. Then, robot moved around for a while and the trajectory is finished by calling /finish_trajectory service. The last 3 submaps generated by the pure localization trajectory will remain in the sparse pose graph, and they will be optimized and serialized into pbstream file. 

Since it is _"pure localization"_, I think it should remove trajectory related data when it is finished. Otherwise, the sparse pose graph and pbstream file would get dirtier everytime pure localization trajectory is started and finished.

PR #562, opened 2017-10-04
Add FinishTrajectory() in SparsePosegraph

jihoonl:
Towards #560


cartographer/mapping_2d/sparse_pose_graph.cc
```diff
@@ -396,6 +401,23 @@ void SparsePoseGraph::WaitForAllComputations() {
   locker.Await([&notification]() { return notification; });
 }
 
+void SparsePoseGraph::FinishTrajectory(const int trajectory_id) {
+  AddWorkItem([this, trajectory_id]() REQUIRES(mutex_) {
+    CHECK_EQ(finished_trajectories_.count(trajectory_id), 0);
+    finished_trajectories_.insert(trajectory_id);
+
+    const auto& submap_data = optimization_problem_.submap_data();
```
ojura: Can you make a pass through all submaps in `submap_data_` which match the given trajectory ID without accessing the `optimization_problem_`? Then you would not need to use the work queue at all.

jihoonl: Submap states are updated within work queue thread always. See [ComputeConstraintsForScan](https://github.com/googlecartographer/cartographer/blob/master/cartographer/mapping_2d/sparse_pose_graph.cc#L293), [AddSubmapFromProto](https://github.com/googlecartographer/cartographer/blob/master/cartographer/mapping_2d/sparse_pose_graph.cc#L437), and [MarkSubmapAsTrimmed](https://github.com/googlecartographer/cartographer/blob/master/cartographer/mapping_2d/sparse_pose_graph.cc#L670).

ojura: Ah yes, good point. I actually posted [the same question](https://github.com/googlecartographer/cartographer/commit/7434e963697c2876add9658e1ff1af6399589215#commitcomment-24545844) to @wohe some time ago.

cartographer/mapping/sparse_pose_graph.h
```diff
@@ -74,6 +74,9 @@ class SparsePoseGraph {
   SparsePoseGraph(const SparsePoseGraph&) = delete;
   SparsePoseGraph& operator=(const SparsePoseGraph&) = delete;
 
+  // Finish a trajectory
```
wohe: `Finishes the given trajectory.`?

cartographer/mapping_2d/sparse_pose_graph.cc
```diff
@@ -365,6 +365,11 @@ void SparsePoseGraph::WaitForAllComputations() {
   locker.Await([&notification]() { return notification; });
 }
 
+void SparsePoseGraph::FinishTrajectory(const int trajectory_id) {
+  // TODO(jihoonl): Add a logic to notify trimmers to finish the given
+  // trajectory
```
wohe: nit: End sentence with a period.

cartographer/mapping_3d/sparse_pose_graph.cc
```diff
@@ -386,6 +386,10 @@ void SparsePoseGraph::WaitForAllComputations() {
   locker.Await([&notification]() { return notification; });
 }
 
+void SparsePoseGraph::FinishTrajectory(const int trajectory_id) {
+  LOG(FATAL) << "Not yet implemented for 3D.";
```
wohe: This would break 3D localization which we now support. Change to a `TODO` comment similar to 2D?

PR #563, opened 2017-10-04
Trim all submaps when pure localization trajectory is finished.

### LINKED ISSUE #560: Trim all submaps when trajectory finishes if pure localization = true

jihoonl:
Fixes #560 
based on #562

### DIFF
--- a/cartographer/mapping/pose_graph.h
+++ b/cartographer/mapping/pose_graph.h
@@ -94,7 +94,7 @@ class PoseGraph {
   // Finishes the given trajectory.
   virtual void FinishTrajectory(int trajectory_id) = 0;
 
-  // Checks if the given trajectory is finished
+  // Checks if the given trajectory is finished.
   virtual bool IsTrajectoryFinished(int trajectory_id) = 0;
 
   // Freezes a trajectory. Poses in this trajectory will not be optimized.
--- a/cartographer/mapping/pose_graph.h
+++ b/cartographer/mapping/pose_graph.h
@@ -94,6 +94,9 @@ class PoseGraph {
   // Finishes the given trajectory.
   virtual void FinishTrajectory(int trajectory_id) = 0;
 
+  // Checks if the given trajectory is finished
+  virtual bool IsTrajectoryFinished(int trajectory_id) = 0;
+
   // Freezes a trajectory. Poses in this trajectory will not be optimized.
   virtual void FreezeTrajectory(int trajectory_id) = 0;
 
--- a/cartographer/mapping/pose_graph_trimmer.h
+++ b/cartographer/mapping/pose_graph_trimmer.h
@@ -37,7 +37,7 @@ class Trimmable {
   // Submaps and nodes are only marked, the numbering remains unchanged.
   virtual void MarkSubmapAsTrimmed(const SubmapId& submap_id) = 0;
 
-  // Check if the given trajectory is finished or not.
+  // Checks if the given trajectory is finished or not.
   virtual bool IsFinished(int trajectory_id) const = 0;
 };
 
@@ -49,7 +49,7 @@ class PoseGraphTrimmer {
   // Called once after each pose graph optimization.
   virtual void Trim(Trimmable* pose_graph) = 0;
 
-  // Checks if this trimmer is terminatable state.
+  // Checks if this trimmer is in a terminatable state.
   virtual bool IsFinished() = 0;
 };
 
--- a/cartographer/mapping/pose_graph_trimmer.h
+++ b/cartographer/mapping/pose_graph_trimmer.h
@@ -36,6 +36,10 @@ class Trimmable {
   // will no longer take part in scan matching, loop closure, visualization.
   // Submaps and nodes are only marked, the numbering remains unchanged.
   virtual void MarkSubmapAsTrimmed(const SubmapId& submap_id) = 0;
+
+  // Check if the given trajectory is finished or not
+  virtual bool IsFinished(int trajectory_id) const = 0;
+
 };
 
 // An interface to implement algorithms that choose how to trim the pose graph.
@@ -45,9 +49,6 @@ class PoseGraphTrimmer {
 
   // Called once after each pose graph optimization.
   virtual void Trim(Trimmable* pose_graph) = 0;
-
-  // Gets finished trajectory
-  virtual void FinishTrajectory(int trajectory_id) = 0;
 
   // Checks if this trimmer is terminatable state
   virtual bool IsFinished() = 0;
@@ -61,7 +62,6 @@ class PureLocalizationTrimmer : public PoseGraphTrimmer {
   ~PureLocalizationTrimmer() override {}
 
   void Trim(Trimmable* pose_graph) override;
-  void FinishTrajectory(int trajectory_id) override;
   bool IsFinished() override;
 
  private:
--- a/cartographer/mapping/pose_graph_trimmer.h
+++ b/cartographer/mapping/pose_graph_trimmer.h
@@ -45,6 +45,12 @@ class PoseGraphTrimmer {
 
   // Called once after each pose graph optimization.
   virtual void Trim(Trimmable* pose_graph) = 0;
+
+  // Gets finished trajectory
+  virtual void FinishTrajectory(int trajectory_id) = 0;
+
+  // Checks if this trimmer is terminatable state
+  virtual bool IsFinished() = 0;
 };
 
 // Keeps the last 'num_submaps_to_keep' of the trajectory with 'trajectory_id'
@@ -55,11 +61,14 @@ class PureLocalizationTrimmer : public PoseGraphTrimmer {
   ~PureLocalizationTrimmer() override {}
 
   void Trim(Trimmable* pose_graph) override;
+  void FinishTrajectory(int trajectory_id) override;
+  bool IsFinished() override;
 
  private:
   const int trajectory_id_;
-  const int num_submaps_to_keep_;
+  int num_submaps_to_keep_;
   int num_submaps_trimmed_ = 0;
+  bool finished_ = false;
 };
 
 }  // namespace mapping
--- a/cartographer/mapping/pose_graph_trimmer.h
+++ b/cartographer/mapping/pose_graph_trimmer.h
@@ -39,7 +39,6 @@ class Trimmable {
 
   // Check if the given trajectory is finished or not
   virtual bool IsFinished(int trajectory_id) const = 0;
-
 };
 
 // An interface to implement algorithms that choose how to trim the pose graph.
--- a/cartographer/mapping/pose_graph_trimmer.h
+++ b/cartographer/mapping/pose_graph_trimmer.h
@@ -37,7 +37,7 @@ class Trimmable {
   // Submaps and nodes are only marked, the numbering remains unchanged.
   virtual void MarkSubmapAsTrimmed(const SubmapId& submap_id) = 0;
 
-  // Check if the given trajectory is finished or not
+  // Check if the given trajectory is finished or not.
   virtual bool IsFinished(int trajectory_id) const = 0;
 };
 
@@ -49,7 +49,7 @@ class PoseGraphTrimmer {
   // Called once after each pose graph optimization.
   virtual void Trim(Trimmable* pose_graph) = 0;
 
-  // Checks if this trimmer is terminatable state
+  // Checks if this trimmer is terminatable state.
   virtual bool IsFinished() = 0;
 };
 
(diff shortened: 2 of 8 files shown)


wohe:
Thanks for the PR. As mentioned in our roadmap at https://github.com/googlecartographer/cartographer/projects/1 we would like to be able to remove trajectories completely. It seems this could be simplified to notify the trimmer that the whole trajectory should be removed. Then there would be no need for additional markers etc. Does that make sense? Can you modify your PR in this way?

jihoonl:
It makes sense. I will update the PR as you suggested.

jihoonl:
Hello @wohe, I am trying to implement trajectory clearing on top of #562 as you suggested and I would like to discuss the logic before actual coding. This was my current plan.
```
1. Add NotifyFinishedTrajectory() in PoseGraphTrimmer.
2. Call NotifyFinishedTrajectory() for each trimmer from SparsePoseGraph::FinishTrajectory()
3. PureLocalizationTrimmer sets num_submaps_to_keep_ to 0 if the trajectory matches.
4. Maybe.. mark the trimmer as finished. and remove it from trimmer list after optimization.
```
Do you have any cleaner logic have in mind? I don't like the `Step 4`. I have thought about some other ways but they may require major change in trimmer api and logic.

wohe:
Looking at the current flow on information it might be more consistent to let the `PoseGraphTrimmer` query the `Trimmable`, i.e. add an `IsFinished()` member function there. Then the trimmer would query this and remove all submaps if the trajectory is finished. Setting `num_submaps_to_keep_` to zero is one possible implementation.

For step 4 you could add another member function to `Trimmable()` which allows notification that the trimmer is done. What do you think?

SirVer:
@jihoonl I think this PR is actionable and in your ballpark right now. Do you have plans to work on this or should it be closed?

jihoonl:
I plan to work on this next week when I get back to the office. I close this for now and will reopen when it is ready again.

jihoonl:
@wohe Ok. Querying `Trimmable` means that it queries to PoseGraph just like it is implemented right? I will rename `finished` to `IsFinished`, and add an additional function in `Trimmable`to notify trimmer is done.

jihoonl:
Um.. actually I'm moving back to the logic I wrote 2 weeks ago. This seems better to keep trimmer flexible.
```
1. Add NotifyFinishedTrajectory() in PoseGraphTrimmer.
2. Call NotifyFinishedTrajectory() for each trimmer from PoseGraph::FinishTrajectory()
3. PureLocalizationTrimmer sets num_submaps_to_keep_ to 0 if the trajectory matches.
4. Check if trimmer is done and remove it from trimmer list after optimization.
```
I will describe a bit more after some testing.

jihoonl:
@wohe
I propose to add `FinishTrajectory` and `IsFinished` in `PoseGraphTrimmer` instead of `Trimmable` for trajectory cleaning. 

I think `FinishTrajectory` should be handled in `PoseGraphTrimmer` because handling a finished trajectory can be very different per trimmer implementation. For instance, assume that there is a trimmer that checks all trajectories to determine what to trim. If this trimmer needs to query `Trimmable` every time 
 to find out which trajectories are finished, then it would be inefficient.

`IsFinished` should be in `PoseGraphTimmer` to be called from PoseGraph. Otherwise, PoseGraph cannot call it.

Let me know if you agree with this change. Then, I will implement the same in 3D.

cartographer/mapping_2d/pose_graph.cc
```diff
@@ -369,8 +385,22 @@ void PoseGraph::WaitForAllComputations() {
 }
 
 void PoseGraph::FinishTrajectory(const int trajectory_id) {
-  // TODO(jihoonl): Add a logic to notify trimmers to finish the given
-  // trajectory.
+  AddWorkItem([this, trajectory_id]() REQUIRES(mutex_) {
+    const auto& submap_data = optimization_problem_.submap_data();
+    for (auto submap_id_data : submap_data.trajectory(trajectory_id)) {
+      submap_data_.at(submap_id_data.id).state = SubmapState::kFinished;
```
wohe: This is what I meant: Wouldn't it be more consistent with the behavior of trimming submaps if we would
1. Tell the trimmer that we are interested in finishing a trajectory, i.e. this would not be part of the trimmer interface but a feature of the localization trimmer.
2. In `Trim()` implement this by telling the `trimming_handle` we would like to remove the trajectory. Then we also would not need a marker for the finished trajectory in the `PoseGraph` object.

cartographer/mapping_2d/pose_graph.cc
```diff
@@ -369,8 +385,22 @@ void PoseGraph::WaitForAllComputations() {
...
+    }
+
+    for (auto& trimmer : trimmers_) {
+      trimmer->FinishTrajectory(trajectory_id);
+    }
+  });
+
+  // To call Trim() when there is no active trajectory
```
wohe: This does not work. It would only call `Trim()` if there is nothing queued already. This logic must be handled inside the work item.

cartographer/mapping_2d/pose_graph.cc
jihoonl: Um sorry I'm still confused what you meant by "implement this by telling the `trimming_handle` in `Trim()`. You meant add an additional variable in `TrimmingHandle` to indicate trajectories to remove? Then, query it in `Trim()`?

cschuet: I don't know when @wohe will be available again to continue the review, but we discussed this PR before he left, so I can take over.

The idea would be to add a method to the `Trimmable` interface to allow the `Trimmer` to query whether a trajectory is already finished. This method would have  to be implemented by the `TrimmingHandle` which just asks its parent `PoseGraph`. In `PoseGraph` I would implement the `FinishTrajectory` method as adding the trajectory_id to a set<int> finished_trajectories_ and implement a `PoseGraph::IsTrajectoryFinished()` method as `return finished_trajectories_.count(trajectory_id)`.

In the implementation of `PureLocalizationTrimmer::Trim(Trimmable*)` you can then check whether the trajectory the `PureLocalizationTrimmer` is responsible for is finished and if so, remove it.

cartographer/mapping_2d/pose_graph.cc
cschuet: As @wohe mentioned getting the concurrency right is a little tricky. Let's move that to another PR and just ignore for now that Trim() is not invoked.

cartographer/mapping_2d/pose_graph.cc
jihoonl: Um I thought @wohe don't want to have an additional marker(`finished_trajectories_`) in `PoseGraph` because he mentioned *we also would not need a marker for the finished trajectory in the PoseGraph object.*

I will update the PR as you explained. Let me know if anything changes.

cartographer/mapping_2d/pose_graph.cc
jihoonl: Without this change, cartographer doesn't clean up the submaps if it was the last active trajectory because no one calls `HandleWorkQueue()`. How about calling it inside of the work item?

cartographer/mapping_2d/pose_graph.cc
cschuet: It was a bit of a back and forth when I discussed with him, but IIRC we settled on caching this information in the `PoseGraph` object.

cartographer/mapping_2d/pose_graph.cc
cschuet: The situation is similar as to `PoseGraph::ComputeConstraintsForNode` and in principle you want the same logic here it you find in mapping_2d/pose_graph.cc lines lines 274 - 280. So maybe it makes sense to pull that out into a function and call it both from here and `ComputeConstraintsForNode`. But I would suggest to leave this as a TODO for this PR and solve it in a follow-up. Or write the PR first that pulls out lines 274 - 280 into it's own function, we merge it and then you call that function here.

cartographer/mapping_2d/pose_graph.cc
jihoonl: Ok.

cartographer/mapping_2d/pose_graph.cc
jihoonl: Ok. I finish this first and will make a quick follow-up

jihoonl:
Pr is updated as discussed.

cartographer/mapping/pose_graph.h
```diff
@@ -94,6 +94,9 @@ class PoseGraph {
   // Finishes the given trajectory.
   virtual void FinishTrajectory(int trajectory_id) = 0;
 
+  // Checks if the given trajectory is finished
```
cschuet: Add period at the end.

cartographer/mapping/pose_graph_trimmer.h
```diff
@@ -36,6 +36,9 @@ class Trimmable {
   // will no longer take part in scan matching, loop closure, visualization.
   // Submaps and nodes are only marked, the numbering remains unchanged.
   virtual void MarkSubmapAsTrimmed(const SubmapId& submap_id) = 0;
+
+  // Check if the given trajectory is finished or not
```
cschuet: Checks if the given trajectory is finished or not.

cartographer/mapping/pose_graph_trimmer.h
```diff
@@ -45,6 +48,9 @@ class PoseGraphTrimmer {
 
   // Called once after each pose graph optimization.
   virtual void Trim(Trimmable* pose_graph) = 0;
+
+  // Checks if this trimmer is terminatable state
```
cschuet: Checks if this trimmer is in a terminatable state.

cartographer/mapping_2d/pose_graph.cc
```diff
@@ -321,6 +321,22 @@ void PoseGraph::HandleWorkQueue() {
           trimmer->Trim(&trimming_handle);
         }
 
+        {
+          auto GetTrimmerToRemove = [&]() {
```
cschuet: Can't you use [std::remove_if](http://en.cppreference.com/w/cpp/algorithm/remove) here?

cartographer/mapping_3d/pose_graph.cc
```diff
@@ -338,6 +338,23 @@ void PoseGraph::HandleWorkQueue() {
           trimmer->Trim(&trimming_handle);
         }
 
+        {
+          auto GetTrimmerToRemove = [&]() {
```
cschuet: Can't you use [std::remove_if](http://en.cppreference.com/w/cpp/algorithm/remove) here?

cartographer/mapping_2d/pose_graph.cc
jihoonl: didn't know this. ok

jihoonl:
Done.

cartographer/mapping/pose_graph_trimmer_test.cc
```diff
@@ -37,6 +37,8 @@ class FakePoseGraph : public Trimmable {
     trimmed_submaps_.push_back(submap_id);
   }
 
+  bool IsFinished(int trajectory_id) const override { return false; }
```
cschuet: const int trajectory_id

cartographer/mapping_2d/pose_graph.cc
```diff
@@ -320,6 +320,11 @@ void PoseGraph::HandleWorkQueue() {
         for (auto& trimmer : trimmers_) {
           trimmer->Trim(&trimming_handle);
         }
+        trimmers_.erase(std::remove_if(
+            trimmers_.begin(), trimmers_.end(),
+            [](std::unique_ptr<mapping::PoseGraphTrimmer>& trimmer) {
+              return trimmer->IsFinished();
+            }));
```
cschuet: Shouldn't this be

~~~
trimmers_.erase(std::remove_if(
          trimmers_.begin(), trimmers_.end(),
          [](std::unique_ptr<mapping::PoseGraphTrimmer>& trimmer) {
            return trimmer->IsFinished();
          }),
    trimmers_.end());
~~~

cartographer/mapping_3d/pose_graph.cc
```diff
@@ -337,6 +337,11 @@ void PoseGraph::HandleWorkQueue() {
         for (auto& trimmer : trimmers_) {
           trimmer->Trim(&trimming_handle);
         }
+        trimmers_.erase(std::remove_if(
+            trimmers_.begin(), trimmers_.end(),
+            [](std::unique_ptr<mapping::PoseGraphTrimmer>& trimmer) {
+              return trimmer->IsFinished();
+            }));
```
cschuet: see above.

cschuet:
@wally-the-cartographer merge

wally-the-cartographer:
Merge requested by authorized user cschuet. Merge queue now has a length of 2.

PR #565, opened 2017-10-04
Add MutexLocker to ConnectedComponents::GetComponent and SparsePoseGr…

zjwoody:
…aph::TrimmingHandle::MarkSubmapAsTrimmed

To make them compile successfully in google.

### DIFF
--- a/cartographer/mapping/connected_components.cc
+++ b/cartographer/mapping/connected_components.cc
@@ -92,6 +92,7 @@ std::vector<std::vector<int>> ConnectedComponents::Components() {
 }
 
 std::vector<int> ConnectedComponents::GetComponent(const int trajectory_id) {
+  common::MutexLocker locker(&lock_);
   const int set_id = FindSet(trajectory_id);
   std::vector<int> trajectory_ids;
   for (const auto& entry : forest_) {
--- a/cartographer/mapping_2d/sparse_pose_graph.cc
+++ b/cartographer/mapping_2d/sparse_pose_graph.cc
@@ -683,7 +683,6 @@ int SparsePoseGraph::TrimmingHandle::num_submaps(
 
 void SparsePoseGraph::TrimmingHandle::MarkSubmapAsTrimmed(
     const mapping::SubmapId& submap_id) {
-  common::MutexLocker locker(&parent_->mutex_);
   // TODO(hrapp): We have to make sure that the trajectory has been finished
   // if we want to delete the last submaps.
   CHECK(parent_->submap_data_.at(submap_id).state == SubmapState::kFinished);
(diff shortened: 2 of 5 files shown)


cartographer/mapping/connected_components.cc
```diff
@@ -92,6 +92,7 @@ std::vector<std::vector<int>> ConnectedComponents::Components() {
 }
 
 std::vector<int> ConnectedComponents::GetComponent(const int trajectory_id) {
+  common::MutexLocker locker(&lock_);
```
cschuet: Did you verify this change? I would expect that this does not work as this function gets called from TrajectoryConnectivityState::Connect(...) <= SparsePoseGraph::UpdateTrajectoryConnectivity(...) <= SparsePoseGraph::HandleWorkQueue() which already takes the lock. So I would expect the line above to fail.

cartographer/mapping_2d/sparse_pose_graph.cc
```diff
@@ -622,6 +622,7 @@ int SparsePoseGraph::TrimmingHandle::num_submaps(
 
 void SparsePoseGraph::TrimmingHandle::MarkSubmapAsTrimmed(
     const mapping::SubmapId& submap_id) {
+  common::MutexLocker locker(&parent_->mutex_);
```
cschuet: Same here.

cartographer/mapping/connected_components.cc
zjwoody: Thanks for the review! Sorry to reply so late. I verified this change, I think they lock different lockers(their own member lock_), I have to add this because GetComponent calls the FindSet which requires a lock, otherwise the build will fail.

cartographer/mapping_2d/sparse_pose_graph.cc
zjwoody: Because it reads parent_->submap_data_ so I add a locker to lock the parent_mutex_.

wohe: Yes, but this gets called via line 351 and line 347 already locks the mutex. This looks like a deadlock with a non-recursive mutex to me. Revert?

zjwoody: I see, thanks!
I will add REQUIRES(parent_->mutex_) at .h instead.

cartographer/mapping/connected_components.cc
cschuet: Sorry for the delay. I lost sight of this.

cschuet:
@wally-the-cartographer merge

wally-the-cartographer:
Merge requested by authorized user cschuet. Merge queue now has a length of 3.

PR #567, opened 2017-10-05
Support multiple trajectories in LoadMap

ojura:
Depends on ~~#569.~~ (merged) and probably ~~#576~~ (closed, not merged) ~~#603~~ (merged, replaces #576)

Also change the prototype of AddSubmapFromProto's argument "initial_pose" to "pose". The loaded pose may be the result of prior optimisation. "initial pose" is a term I associate with an unoptimized pose.

Partially replaces #541.

Note: LoadMap will be renamed in #554 to LoadState.

### DIFF
--- a/cartographer/mapping/map_builder.cc
+++ b/cartographer/mapping/map_builder.cc
@@ -180,12 +180,12 @@ void MapBuilder::LoadMap(io::ProtoStreamReader* const reader) {
   std::map<int, int> trajectory_remapping;
   for (auto& trajectory_proto : *pose_graph.mutable_trajectory()) {
     const int new_trajectory_id = AddTrajectoryForDeserialization();
-    // Also takes care of checking for duplicates.
     CHECK(trajectory_remapping
               .emplace(trajectory_proto.trajectory_id(), new_trajectory_id)
-              .second);
-    sparse_pose_graph_->FreezeTrajectory(new_trajectory_id);
+              .second)
+        << "Duplicate trajectory ID: " << trajectory_proto.trajectory_id();
     trajectory_proto.set_trajectory_id(new_trajectory_id);
+    sparse_pose_graph_->FreezeTrajectory(new_trajectory_id);
   }
 
   // Apply the calculated remapping to constraints in the SparsePoseGraph proto
@@ -221,21 +221,19 @@ void MapBuilder::LoadMap(io::ProtoStreamReader* const reader) {
       break;
     }
     if (proto.has_node()) {
-      const int new_trajectory_id =
-          trajectory_remapping.at(proto.node().node_id().trajectory_id());
       proto.mutable_node()->mutable_node_id()->set_trajectory_id(
-          new_trajectory_id);
-      const transform::Rigid3d node_pose = node_poses.at(
-          NodeId{new_trajectory_id, proto.node().node_id().node_index()});
+          trajectory_remapping.at(proto.node().node_id().trajectory_id()));
+      const transform::Rigid3d node_pose =
+          node_poses.at(NodeId{proto.node().node_id().trajectory_id(),
+                               proto.node().node_id().node_index()});
       sparse_pose_graph_->AddNodeFromProto(node_pose, proto.node());
     }
     if (proto.has_submap()) {
-      const int new_trajectory_id =
-          trajectory_remapping.at(proto.submap().submap_id().trajectory_id());
       proto.mutable_submap()->mutable_submap_id()->set_trajectory_id(
-          new_trajectory_id);
-      const transform::Rigid3d submap_pose = submap_poses.at(SubmapId{
-          new_trajectory_id, proto.submap().submap_id().submap_index()});
+          trajectory_remapping.at(proto.submap().submap_id().trajectory_id()));
+      const transform::Rigid3d submap_pose =
+          submap_poses.at(SubmapId{proto.submap().submap_id().trajectory_id(),
+                                   proto.submap().submap_id().submap_index()});
       sparse_pose_graph_->AddSubmapFromProto(submap_pose, proto.submap());
     }
   }
--- a/cartographer/mapping/map_builder.cc
+++ b/cartographer/mapping/map_builder.cc
@@ -177,8 +177,11 @@ void MapBuilder::LoadMap(io::ProtoStreamReader* const reader) {
   proto::SparsePoseGraph pose_graph;
   CHECK(reader->ReadProto(&pose_graph));
 
-  const int map_trajectory_id = AddTrajectoryForDeserialization();
-  sparse_pose_graph_->FreezeTrajectory(map_trajectory_id);
+  for (int trajectory_id = 0; trajectory_id < pose_graph.trajectory().size();
+       ++trajectory_id) {
+    CHECK_EQ(AddTrajectoryForDeserialization(), trajectory_id);
+    sparse_pose_graph_->FreezeTrajectory(trajectory_id);
+  }
 
   MapById<SubmapId, transform::Rigid3d> submap_poses;
   for (const proto::Trajectory& trajectory_proto : pose_graph.trajectory()) {
@@ -208,15 +211,13 @@ void MapBuilder::LoadMap(io::ProtoStreamReader* const reader) {
       const transform::Rigid3d node_pose =
           node_poses.at(NodeId{proto.node().node_id().trajectory_id(),
                                proto.node().node_id().node_index()});
-      sparse_pose_graph_->AddNodeFromProto(map_trajectory_id, node_pose,
-                                           proto.node());
+      sparse_pose_graph_->AddNodeFromProto(node_pose, proto.node());
     }
     if (proto.has_submap()) {
       const transform::Rigid3d submap_pose =
           submap_poses.at(SubmapId{proto.submap().submap_id().trajectory_id(),
                                    proto.submap().submap_id().submap_index()});
-      sparse_pose_graph_->AddSubmapFromProto(map_trajectory_id, submap_pose,
-                                             proto.submap());
+      sparse_pose_graph_->AddSubmapFromProto(submap_pose, proto.submap());
     }
   }
   CHECK(reader->eof());
--- a/cartographer/mapping/map_builder.cc
+++ b/cartographer/mapping/map_builder.cc
@@ -177,10 +177,23 @@ void MapBuilder::LoadMap(io::ProtoStreamReader* const reader) {
   proto::SparsePoseGraph pose_graph;
   CHECK(reader->ReadProto(&pose_graph));
 
-  for (int trajectory_id = 0; trajectory_id < pose_graph.trajectory().size();
-       ++trajectory_id) {
-    CHECK_EQ(AddTrajectoryForDeserialization(), trajectory_id);
-    sparse_pose_graph_->FreezeTrajectory(trajectory_id);
+  std::map<int, int> trajectory_remapping;
+  for (auto& trajectory_proto : *pose_graph.mutable_trajectory()) {
+    const int new_trajectory_id = AddTrajectoryForDeserialization();
+    // Also takes care of checking for duplicates.
+    CHECK(trajectory_remapping
+              .emplace(trajectory_proto.trajectory_id(), new_trajectory_id)
+              .second);
+    sparse_pose_graph_->FreezeTrajectory(new_trajectory_id);
+    trajectory_proto.set_trajectory_id(new_trajectory_id);
+  }
+
+  // Apply the calculated remapping to constraints in the SparsePoseGraph proto
+  for (auto& constraint_proto : *pose_graph.mutable_constraint()) {
+    constraint_proto.mutable_submap_id()->set_trajectory_id(
+        trajectory_remapping.at(constraint_proto.submap_id().trajectory_id()));
+    constraint_proto.mutable_node_id()->set_trajectory_id(
+        trajectory_remapping.at(constraint_proto.node_id().trajectory_id()));
   }
 
   MapById<SubmapId, transform::Rigid3d> submap_poses;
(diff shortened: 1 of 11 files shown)


cartographer/mapping/sparse_pose_graph.h
```diff
@@ -77,10 +77,9 @@ class SparsePoseGraph {
   // Freezes a trajectory. Poses in this trajectory will not be optimized.
   virtual void FreezeTrajectory(int trajectory_id) = 0;
 
-  // Adds a 'submap' from a proto with the given 'initial_pose' to the frozen
+  // Adds a 'submap' from a proto with the given 'pose' to the frozen
```
wohe: Please update the comment.

cartographer/mapping_2d/sparse_pose_graph.cc
```diff
@@ -407,21 +407,24 @@ void SparsePoseGraph::FreezeTrajectory(const int trajectory_id) {
...
   std::shared_ptr<const Submap> submap_ptr =
       std::make_shared<const Submap>(submap.submap_2d());
-  const transform::Rigid2d initial_pose_2d = transform::Project2D(initial_pose);
+  const transform::Rigid2d pose_2d = transform::Project2D(pose);
 
   common::MutexLocker locker(&mutex_);
   trajectory_connectivity_state_.Add(trajectory_id);
   const mapping::SubmapId submap_id =
       submap_data_.Append(trajectory_id, SubmapData());
+  CHECK_EQ(submap_id, proto_submap_id);
```
wohe: This PR removes support for serialized data with trimmed submaps. Add a `TODO` that this does not support trimmed submaps yet and add output reflecting this to the `CHECK`? Similar for 3D.

cartographer/mapping/map_builder.cc
```diff
@@ -201,17 +201,20 @@ void MapBuilder::LoadMap(io::ProtoStreamReader* const reader) {
...
-      ->mutable_submaps_options()
-      ->set_resolution(0.05);
-  unused_options.mutable_trajectory_builder_3d_options();
-
-  const std::unordered_set<string> unused_sensor_ids;
-  const int map_trajectory_id =
-      AddTrajectoryBuilder(unused_sensor_ids, unused_options);
-  FinishTrajectory(map_trajectory_id);
-  sparse_pose_graph_->FreezeTrajectory(map_trajectory_id);
+  for (int trajectory_id = 0; trajectory_id < pose_graph.trajectory().size();
```
wohe: I'm wondering whether this should not be implemented as a separate function `DeserializeState()`. While this is adding support for deserializing more of the serialized state, it also no longer works with trimmed data. Before this was supported.

cartographer/mapping_2d/sparse_pose_graph.cc
ojura: I had the impression that, while serializing, IDs are remapped to look like there wasn't any trimming at all. Isn't this the case?

cartographer/mapping/map_builder.cc
ojura: I planned that when `LoadMap` learns how to load an unfrozen trajectory, it should be renamed to `LoadState`. That happens in #554. I wanted to preserve all existing functionality in `LoadState` by calling it with `load_frozen_trajectories = true`.

cartographer/mapping/sparse_pose_graph.h
ojura: Hasn't the comment been updated?

ojura: Done.

cartographer/mapping_2d/sparse_pose_graph.cc
ojura: Now I see that in `MapBuilder`, you [do not apply the same remapping](https://github.com/googlecartographer/cartographer/blob/cae9c02e0589fb3d7ae16372c9803a0f47b3829f/cartographer/mapping/map_builder.cc#L161), but you have left it as a TODO. When the TODO is completed, this `CHECK` will completely be fine, right?

cartographer/mapping_2d/sparse_pose_graph.cc
```diff
@@ -373,52 +373,55 @@ void SparsePoseGraph::FreezeTrajectory(const int trajectory_id) {
   });
 }
 
-void SparsePoseGraph::AddSubmapFromProto(const int trajectory_id,
-                                         const transform::Rigid3d& initial_pose,
+void SparsePoseGraph::AddSubmapFromProto(const transform::Rigid3d& pose,
```
wohe: Let's follow #591 and consistently name poses after the frame they relate to, i.e. here it would be `global_pose`.

cartographer/mapping/map_builder.cc
```diff
@@ -177,8 +177,11 @@ void MapBuilder::LoadMap(io::ProtoStreamReader* const reader) {
   proto::SparsePoseGraph pose_graph;
   CHECK(reader->ReadProto(&pose_graph));
 
-  const int map_trajectory_id = AddTrajectoryForDeserialization();
-  sparse_pose_graph_->FreezeTrajectory(map_trajectory_id);
+  for (int trajectory_id = 0; trajectory_id < pose_graph.trajectory().size();
+       ++trajectory_id) {
+    CHECK_EQ(AddTrajectoryForDeserialization(), trajectory_id);
```
wohe: It's a good idea to add trajectories with IDs 0, ..., n-1. The deserialized nodes and submaps have to be mapped to these trajectories then, though. Persisted trajectory IDs are `pose_graph.trajectory(trajectory_id).trajectory_id()`. One more suggestion: Make this a range-based for loop, remove this `CHECK`, and add a map from the persisted trajectory IDs to the new trajectory IDs. This way we would already support loading multiple maps.

cartographer/mapping_3d/sparse_pose_graph.cc
```diff
@@ -393,35 +393,42 @@ void SparsePoseGraph::FreezeTrajectory(const int trajectory_id) {
...
+                                             submap.submap_id().submap_index()};
+  const int trajectory_id = proto_submap_id.trajectory_id;
   std::shared_ptr<const Submap> submap_ptr =
       std::make_shared<const Submap>(submap.submap_3d());
 
   common::MutexLocker locker(&mutex_);
   AddTrajectoryIfNeeded(trajectory_id);
   const mapping::SubmapId submap_id =
       submap_data_.Append(trajectory_id, SubmapData());
+  // TODO: remove this check when the optimization problem is refactored to
```
wohe: Why is this different in 2D?

ojura: The optimization problem still keeps its trajectory nodes in nested vector, so I could not write a `InsertTrajectoryNode` method for the optimization problem (which inserts a node with an exact ID from the proto) without going in the wrong direction. The ID remapping has been removed, so this means that `AddSubmapFromProto` is actually currently broken for trimmed data until `AddTrajectoryNode` is replaced with `InsertTrajectoryNode`. So I decided to keep everything as-is in 3D for the time being.

This is why I mentioned in #549 that it would be better to first merge MapByID for the SPG3D optimization problem. Then, I could make the change for 3D as I did with 2D.

wohe: Where do you still see nested vectors? This should have been fixed by @cschuet and me in the last days.

cartographer/mapping/map_builder.cc
ojura: Oh, I did not have a remapping in mind, I just assumed that serialized trajectories would always be labeled from 0 to n-1. I am not a huge fan of remapping, because that would mean that we would have to also apply it when deserializing constraints. I would rather restore everything how it was previously.

wohe: Sure, once all data from a trajectory is gone, the trajectory ID might not be there anymore. Moreover, the more important aspect is that we want to be able to deserialize multiple sources into one instance. This means there is no way around remapping during deserialization anyway.

cartographer/mapping_3d/sparse_pose_graph.cc
ojura: Sorry, I accidentally misread this when reading the header of the 3D optimization problem. My eyes slipped to another line and saw a vector of something, so I just assumed it hasn't been refactored yet. I will update this to exactly follow 2D.

cartographer/mapping/map_builder.cc
ojura: Makes sense. I will add the remapping.

ojura: By the way: what do you think about applying the remapping to the SPG proto in-place, immediately? Or would you prefer to pass the remapping it to deserialization functions in the SPG (to be added in later PRs) which use the SPG proto?

wohe: Not sure if this helps you a lot. You'd still have to create a map to efficiently look up the new trajectory IDs and keep it around since the nodes/submaps are in separate protos.

ojura: Yes, but at least I wouldn't have to pass the map around - if I apply it the remapping to all protos in `LoadMap` (soon to become `LoadState`), I can hide the fact that a remapping was applied from the rest of Cartographer.

wohe: I misunderstood and think we agree: if you pass a proto it makes sense to change it accordingly so that the consumers of that proto are not forced to implement remapping.

ojura:
Updated, PTAL

cartographer/mapping_2d/sparse_pose_graph.cc
ojura: Done.

ojura:
Also, if there's too much stuff going on, feel free to suggest me to split the PR.

cartographer/mapping/map_builder.cc
```diff
@@ -177,8 +177,24 @@ void MapBuilder::LoadMap(io::ProtoStreamReader* const reader) {
   proto::SparsePoseGraph pose_graph;
   CHECK(reader->ReadProto(&pose_graph));
 
-  const int map_trajectory_id = AddTrajectoryForDeserialization();
-  sparse_pose_graph_->FreezeTrajectory(map_trajectory_id);
+  std::map<int, int> trajectory_remapping;
+  for (auto& trajectory_proto : *pose_graph.mutable_trajectory()) {
+    const int new_trajectory_id = AddTrajectoryForDeserialization();
+    // Also takes care of checking for duplicates.
```
wohe: Prefer code over a comment, i.e. add to the `CHECK`:
```
<< "Duplicate trajectory ID: " << trajectory_proto.trajectory_id();
```
?

cartographer/mapping/map_builder.cc
```diff
@@ -177,8 +177,24 @@ void MapBuilder::LoadMap(io::ProtoStreamReader* const reader) {
...
+    CHECK(trajectory_remapping
+              .emplace(trajectory_proto.trajectory_id(), new_trajectory_id)
+              .second);
+    sparse_pose_graph_->FreezeTrajectory(new_trajectory_id);
+    trajectory_proto.set_trajectory_id(new_trajectory_id);
+  }
+
+  // Apply the calculated remapping to constraints in the SparsePoseGraph proto
```
wohe: Remove this code? The constraints are not used in this function.

cartographer/mapping/sparse_pose_graph.h
```diff
@@ -77,16 +77,14 @@ class SparsePoseGraph {
   // Freezes a trajectory. Poses in this trajectory will not be optimized.
   virtual void FreezeTrajectory(int trajectory_id) = 0;
 
-  // Adds a 'submap' from a proto with the given 'initial_pose' to the frozen
-  // trajectory with 'trajectory_id'.
-  virtual void AddSubmapFromProto(int trajectory_id,
-                                  const transform::Rigid3d& initial_pose,
+  // Adds a 'submap' from a proto with the given 'pose' to the appropriate
```
wohe: Update comment? `'global_pose'`

cartographer/mapping/map_builder.cc
```diff
@@ -205,18 +221,22 @@ void MapBuilder::LoadMap(io::ProtoStreamReader* const reader) {
...
     if (proto.has_node()) {
-      const transform::Rigid3d node_pose =
-          node_poses.at(NodeId{proto.node().node_id().trajectory_id(),
-                               proto.node().node_id().node_index()});
-      sparse_pose_graph_->AddNodeFromProto(map_trajectory_id, node_pose,
-                                           proto.node());
+      const int new_trajectory_id =
+          trajectory_remapping.at(proto.node().node_id().trajectory_id());
+      proto.mutable_node()->mutable_node_id()->set_trajectory_id(
+          new_trajectory_id);
```
wohe: Might be more readable to inline the remapping here and leave the `node_pose` code as before?

cartographer/mapping_2d/sparse_pose_graph.cc
```diff
@@ -373,55 +373,56 @@ void SparsePoseGraph::FreezeTrajectory(const int trajectory_id) {
...
 
-void SparsePoseGraph::AddSubmapFromProto(const int trajectory_id,
-                                         const transform::Rigid3d& initial_pose,
+void SparsePoseGraph::AddSubmapFromProto(const transform::Rigid3d& global_pose,
                                          const mapping::proto::Submap& submap) {
   if (!submap.has_submap_2d()) {
     return;
   }
 
+  const mapping::SubmapId proto_submap_id = {submap.submap_id().trajectory_id(),
```
wohe: Conciseness: Just `submap_id`? It is not just in the proto, it is also used everywhere else.

cartographer/mapping/map_builder.cc
ojura: This is preparation for #544.

cartographer/mapping_2d/sparse_pose_graph.cc
ojura: Sure. It's an artifact from times when we received the actual inserted id from `NestedVectorsById::Append`.

cartographer/mapping/map_builder.cc
ojura: Done

cartographer/mapping/map_builder.cc
ojura: Done

ojura:
Updated.

cartographer/mapping/map_builder.cc
wohe: Then let's do this change there. It is difficult to discuss this change without context and it is orthogonal to the other content of this PR.

cartographer/mapping_2d/sparse_pose_graph/optimization_problem.h
```diff
@@ -65,12 +65,18 @@ class OptimizationProblem {
   void AddImuData(int trajectory_id, const sensor::ImuData& imu_data);
   void AddOdometerData(int trajectory_id,
                        const sensor::OdometryData& odometry_data);
+  void InsertTrajectoryNode(const mapping::NodeId& node_id, common::Time time,
```
wohe: Consistency: `Insert...` after `Add...`, same in the `.cc` file to match 3D?

ojura: Removed and reordered.

ojura:
Ok, should be ready. I also rebased on top of master.

wohe:
Thanks for the PR @ojura!

ojura:
No problem, that you for putting up with my PR spam :-)

PR #568, opened 2017-10-05
Refactor node initial pose (move to constant data).

ojura:
Partially replaces #541.

PR #569, opened 2017-10-05
Deserialize trajectory nodes

ojura:
Partially replaces #541. Depends on ~~#568~~ (merged), and ~~also on #567 in a way (`AddNodeFromProto` does not take a `trajectory_id`, following how `AddSubmapFromProto` is changed in #567~~).

### DIFF
--- a/cartographer/mapping/map_builder.cc
+++ b/cartographer/mapping/map_builder.cc
@@ -218,6 +218,14 @@ void MapBuilder::LoadMap(io::ProtoStreamReader* const reader) {
     if (!reader->ReadProto(&proto)) {
       break;
     }
+    if (proto.has_node()) {
+      const auto& pose_graph_node =
+          pose_graph.trajectory(proto.node().node_id().trajectory_id())
+              .node(proto.node().node_id().node_index());
+      const transform::Rigid3d pose =
+          transform::ToRigid3(pose_graph_node.pose());
+      sparse_pose_graph_->AddNodeFromProto(pose, proto.node());
+    }
     if (proto.has_submap()) {
       const transform::Rigid3d submap_pose = transform::ToRigid3(
           pose_graph.trajectory(proto.submap().submap_id().trajectory_id())
--- a/cartographer/mapping/map_builder.cc
+++ b/cartographer/mapping/map_builder.cc
@@ -224,7 +224,8 @@ void MapBuilder::LoadMap(io::ProtoStreamReader* const reader) {
               .node(proto.node().node_id().node_index());
       const transform::Rigid3d pose =
           transform::ToRigid3(pose_graph_node.pose());
-      sparse_pose_graph_->AddNodeFromProto(pose, proto.node());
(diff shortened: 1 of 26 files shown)


cartographer/mapping/map_builder.cc
```diff
@@ -218,6 +218,14 @@ void MapBuilder::LoadMap(io::ProtoStreamReader* const reader) {
     if (!reader->ReadProto(&proto)) {
       break;
     }
+    if (proto.has_node()) {
+      const auto& pose_graph_node =
+          pose_graph.trajectory(proto.node().node_id().trajectory_id())
+              .node(proto.node().node_id().node_index());
+      const transform::Rigid3d pose =
+          transform::ToRigid3(pose_graph_node.pose());
+      sparse_pose_graph_->AddNodeFromProto(pose, proto.node());
```
wohe: To get this merged, let's pass the `map_trajectory_id` similar to the handling of submaps. Map loading right now creates a single trajectory to contain the frozen data. This is unproblematic since no optimization variables are added for the frozen data.

To handle multiple trajectories with trimmed data, we have to change the serialization first. Since we now also want to support trimming in the middle, we have to change the serialization to maintain information about gaps, e.g. having explicit node/submap indices instead of implicit from the repeated field index.

ojura: I understand that we can't use the full node/submap ID from the proto, because the node/submap index is not remapped to account for trimmed nodes/submaps. But I am confused about why can't we use just the trajectory ID from the proto? AFAIK these are not affected by trimming. I don't see an advantage to passing `map_trajectory_id` via the function call, same with `AddSubmapFromProto`.

And that `CHECK_EQ(node_id, proto_node_id)` which breaks loading trimmed trajectories can be disabled, as I suggested in #567, or softened just to compare the trajectory IDs.

wohe: We can't just use the ID from the proto because we only create one frozen trajectory here which has the ID `map_trajectory_id`. If we merge this PR now, we might add nodes to trajectories which do not exist, i.e. this is a bug. If you pass the `map_trajectory_id`, we this PR adds support for loading maps including the node data to use it for loop closures.

ojura: Yes, that's why imagined that this gets merged after #567. But sure, if you want to merge this first, I can pass `map_trajectory_id` for now.

wohe: Yes, this is preferable so that the code base is in a good state at each commit if possible.

cartographer/mapping/sparse_pose_graph.h
```diff
@@ -83,6 +83,10 @@ class SparsePoseGraph {
                                   const transform::Rigid3d& initial_pose,
                                   const proto::Submap& submap) = 0;
 
+  virtual void AddNodeFromProto(int trajectory_id,
```
wohe: Add a comment what this does, similar to `AddSubmapFromProto()`?

ojura:
Done, PTAL

ojura:
Done

Issue #508, opened 2017-09-07
Range data is no longer kept around


ojura:
In recent PRs (#503, #488), keeping range data in memory (in trajectory nodes) has been removed from libcartographer. Users of libcartographer who used this data (e.g for realtime map rebuilding as discussed in https://github.com/googlecartographer/cartographer_ros/issues/475, or to implement something like #415) are now out of luck.

Would you consider bringing this back, for example as a configurable option (either runtime or compile time)? This way, if you require the range data, you can sacrifice some memory and keep it around; otherwise (and by default), behavior is like in the present implementation, with reduced memory usage. This would be the best of both worlds. Then, for example, the realtime map building functionality would depend on this option being enabled; otherwise, you have to use the assets writer, which would retrieve the range data from the bag.

wohe:
I think this discussion is already spread out over too many issues. I can just reiterate:

- Memory consumption dropped significantly since we no longer have to keep range data around.
- Real-time rebuilding of submaps for SLAM is not something I see us adding in favor of improved CPU usage and memory consumption; improving submap quality by improving the performance of local SLAM seems like the better approach to me.
- If you are unsatisfied with the current performance of the assets writer, I'd prefer to concentrate on getting this fixed, e.g. as discussed in googlecartographer/cartographer_ros#475.

ojura:
Perhaps I focused too much on the map rebuilding thing here -- that is not the only reason why I would like to keep the range data available. I agree that all discussion regarding map rebuilding belongs in other issues.

I wanted to focus in this issue that, as a user of libcartographer, I had been using the range data, and I would like to continue to do so. Of course, I can keep a fork where I retain this. I thought that allowing the library user to decide for themselves whether to keep the range data in memory would be an acceptable compromise which could end up in upstream, thus negating the need for a fork.

wohe:
I think I'm missing some context here. Of course you should not have to fork. If you're using range data outside of SLAM why is it not easily possible to keep it around in code separate from the SLAM library? That would seem to achieve better separation of concerns.

ojura:
> I think I'm missing some context here.

Sure. For example, among other uses, we are using the point clouds for visualization purposes, and having them stored inside trajectory nodes enabled us to visualize what the rangefinder device had seen at a particular point on the trajectory. 

[...]

PR #574, opened 2017-10-06
Add local SLAM result callback.

ojura:
Depends on ~~#619~~ (merged) and ~~#617~~ (merged).
Related to #508.

Also, if cartographer_ros is going to use this, and we wish to serialize the saved range data, that will have to be handled in cartographer_ros, right?

### DIFF
--- a/cartographer/mapping/global_trajectory_builder.h
+++ b/cartographer/mapping/global_trajectory_builder.h
@@ -29,12 +29,14 @@ template <typename LocalTrajectoryBuilder,
 class GlobalTrajectoryBuilder
     : public mapping::GlobalTrajectoryBuilderInterface {
  public:
-  GlobalTrajectoryBuilder(const LocalTrajectoryBuilderOptions& options,
-                          const int trajectory_id,
-                          SparsePoseGraph* const sparse_pose_graph)
+  GlobalTrajectoryBuilder(
+      const LocalTrajectoryBuilderOptions& options, const int trajectory_id,
+      SparsePoseGraph* const sparse_pose_graph,
+      const AccumulatedRangeDataCallback& accumulated_range_data_callback)
       : trajectory_id_(trajectory_id),
         sparse_pose_graph_(sparse_pose_graph),
-        local_trajectory_builder_(options) {}
+        local_trajectory_builder_(options),
+        accumulated_range_data_callback_(accumulated_range_data_callback) {}
   ~GlobalTrajectoryBuilder() override {}
 
   GlobalTrajectoryBuilder(const GlobalTrajectoryBuilder&) = delete;
@@ -62,6 +64,11 @@ class GlobalTrajectoryBuilder
               matching_result->insertion_result->insertion_submaps));
       CHECK_EQ(node_id->trajectory_id, trajectory_id_);
     }
+    if (accumulated_range_data_callback_) {
+      accumulated_range_data_callback_(
+          trajectory_id_, matching_result->time, matching_result->local_pose,
+          std::move(matching_result->range_data_in_local), std::move(node_id));
+    }
   }
 
   void AddSensorData(const sensor::ImuData& imu_data) override {
@@ -85,6 +92,7 @@ class GlobalTrajectoryBuilder
   const int trajectory_id_;
   SparsePoseGraph* const sparse_pose_graph_;
   LocalTrajectoryBuilder local_trajectory_builder_;
+  AccumulatedRangeDataCallback accumulated_range_data_callback_;
 };
 
 }  // namespace mapping
--- a/cartographer/mapping/global_trajectory_builder.h
+++ b/cartographer/mapping/global_trajectory_builder.h
@@ -32,11 +32,11 @@ class GlobalTrajectoryBuilder
   GlobalTrajectoryBuilder(
       const LocalTrajectoryBuilderOptions& options, const int trajectory_id,
       SparsePoseGraph* const sparse_pose_graph,
-      const AccumulatedRangeDataCallback& accumulated_range_data_callback)
+      const LocalSlamResultCallback& local_slam_result_callback)
       : trajectory_id_(trajectory_id),
         sparse_pose_graph_(sparse_pose_graph),
         local_trajectory_builder_(options),
-        accumulated_range_data_callback_(accumulated_range_data_callback) {}
+        local_slam_result_callback_(local_slam_result_callback) {}
   ~GlobalTrajectoryBuilder() override {}
 
   GlobalTrajectoryBuilder(const GlobalTrajectoryBuilder&) = delete;
@@ -64,8 +64,8 @@ class GlobalTrajectoryBuilder
               matching_result->insertion_result->insertion_submaps));
       CHECK_EQ(node_id->trajectory_id, trajectory_id_);
     }
-    if (accumulated_range_data_callback_) {
-      accumulated_range_data_callback_(
+    if (local_slam_result_callback_) {
+      local_slam_result_callback_(
           trajectory_id_, matching_result->time, matching_result->local_pose,
           std::move(matching_result->range_data_in_local), std::move(node_id));
     }
@@ -92,7 +92,7 @@ class GlobalTrajectoryBuilder
   const int trajectory_id_;
   SparsePoseGraph* const sparse_pose_graph_;
   LocalTrajectoryBuilder local_trajectory_builder_;
-  AccumulatedRangeDataCallback accumulated_range_data_callback_;
+  LocalSlamResultCallback local_slam_result_callback_;
 };
 
 }  // namespace mapping
--- a/cartographer/mapping/global_trajectory_builder_interface.h
+++ b/cartographer/mapping/global_trajectory_builder_interface.h
@@ -62,7 +62,7 @@ class GlobalTrajectoryBuilderInterface {
   // A callback which is called after local SLAM processes an accumulated
   // 'sensor::RangeData'. If the data was inserted into a submap, reports the
   // assigned 'NodeId', otherwise 'nullptr' if the data was filtered out.
-  using AccumulatedRangeDataCallback =
+  using LocalSlamResultCallback =
       std::function<void(int /* trajectory ID */, common::Time,
                          transform::Rigid3d /* local pose estimate */,
                          sensor::RangeData /* in local frame */,
--- a/cartographer/mapping/global_trajectory_builder_interface.h
+++ b/cartographer/mapping/global_trajectory_builder_interface.h
@@ -58,6 +58,15 @@ class GlobalTrajectoryBuilderInterface {
   virtual void AddSensorData(const sensor::OdometryData& odometry_data) = 0;
   virtual void AddSensorData(
       const sensor::FixedFramePoseData& fixed_frame_pose) = 0;
+
+  // A callback which is called after local SLAM processes an accumulated
+  // 'sensor::RangeData'. If the data was inserted into a submap, reports the
+  // assigned 'NodeId', otherwise 'nullptr' if the data was filtered out.
+  using AccumulatedRangeDataCallback =
+      std::function<void(int /* trajectory ID */, common::Time,
+                         transform::Rigid3d /* local pose estimate */,
+                         sensor::RangeData /* in local frame */,
+                         std::unique_ptr<const mapping::NodeId>)>;
 };
 
 }  // namespace mapping
--- a/cartographer/mapping/map_builder.cc
+++ b/cartographer/mapping/map_builder.cc
@@ -55,10 +55,10 @@ proto::MapBuilderOptions CreateMapBuilderOptions(
 
 MapBuilder::MapBuilder(
     const proto::MapBuilderOptions& options,
-    const AccumulatedRangeDataCallback& accumulated_range_data_callback)
+    const LocalSlamResultCallback& local_slam_result_callback)
     : options_(options),
       thread_pool_(options.num_background_threads()),
-      accumulated_range_data_callback_(accumulated_range_data_callback) {
+      local_slam_result_callback_(local_slam_result_callback) {
   if (options.use_trajectory_builder_2d()) {
     sparse_pose_graph_2d_ = common::make_unique<mapping_2d::SparsePoseGraph>(
         options_.sparse_pose_graph_options(), &thread_pool_);
@@ -88,7 +88,7 @@ int MapBuilder::AddTrajectoryBuilder(
                 mapping_3d::SparsePoseGraph>>(
                 trajectory_options.trajectory_builder_3d_options(),
                 trajectory_id, sparse_pose_graph_3d_.get(),
-                accumulated_range_data_callback_)));
+                local_slam_result_callback_)));
   } else {
     CHECK(trajectory_options.has_trajectory_builder_2d_options());
     trajectory_builders_.push_back(
@@ -100,7 +100,7 @@ int MapBuilder::AddTrajectoryBuilder(
                 mapping_2d::SparsePoseGraph>>(
                 trajectory_options.trajectory_builder_2d_options(),
                 trajectory_id, sparse_pose_graph_2d_.get(),
-                accumulated_range_data_callback_)));
+                local_slam_result_callback_)));
   }
   if (trajectory_options.pure_localization()) {
     constexpr int kSubmapsToKeep = 3;
--- a/cartographer/mapping/map_builder.cc
+++ b/cartographer/mapping/map_builder.cc
@@ -53,8 +53,12 @@ proto::MapBuilderOptions CreateMapBuilderOptions(
   return options;
 }
 
-MapBuilder::MapBuilder(const proto::MapBuilderOptions& options)
-    : options_(options), thread_pool_(options.num_background_threads()) {
+MapBuilder::MapBuilder(
+    const proto::MapBuilderOptions& options,
+    const AccumulatedRangeDataCallback& accumulated_range_data_callback)
+    : options_(options),
+      thread_pool_(options.num_background_threads()),
+      accumulated_range_data_callback_(accumulated_range_data_callback) {
   if (options.use_trajectory_builder_2d()) {
     sparse_pose_graph_2d_ = common::make_unique<mapping_2d::SparsePoseGraph>(
         options_.sparse_pose_graph_options(), &thread_pool_);
@@ -83,7 +87,8 @@ int MapBuilder::AddTrajectoryBuilder(
                 mapping_3d::proto::LocalTrajectoryBuilderOptions,
                 mapping_3d::SparsePoseGraph>>(
                 trajectory_options.trajectory_builder_3d_options(),
-                trajectory_id, sparse_pose_graph_3d_.get())));
+                trajectory_id, sparse_pose_graph_3d_.get(),
+                accumulated_range_data_callback_)));
   } else {
     CHECK(trajectory_options.has_trajectory_builder_2d_options());
     trajectory_builders_.push_back(
@@ -94,7 +99,8 @@ int MapBuilder::AddTrajectoryBuilder(
                 mapping_2d::proto::LocalTrajectoryBuilderOptions,
                 mapping_2d::SparsePoseGraph>>(
                 trajectory_options.trajectory_builder_2d_options(),
-                trajectory_id, sparse_pose_graph_2d_.get())));
+                trajectory_id, sparse_pose_graph_2d_.get(),
+                accumulated_range_data_callback_)));
   }
   if (trajectory_options.pure_localization()) {
     constexpr int kSubmapsToKeep = 3;
--- a/cartographer/mapping/map_builder.h
+++ b/cartographer/mapping/map_builder.h
@@ -50,7 +50,7 @@ proto::MapBuilderOptions CreateMapBuilderOptions(
 class MapBuilder {
  public:
   using LocalSlamResultCallback =
-      mapping::GlobalTrajectoryBuilderInterface::LocalSlamResultCallback;
+      GlobalTrajectoryBuilderInterface::LocalSlamResultCallback;
 
   MapBuilder(
       const proto::MapBuilderOptions& options,
--- a/cartographer/mapping/map_builder.h
+++ b/cartographer/mapping/map_builder.h
@@ -52,9 +52,8 @@ class MapBuilder {
   using LocalSlamResultCallback =
       GlobalTrajectoryBuilderInterface::LocalSlamResultCallback;
 
-  MapBuilder(
-      const proto::MapBuilderOptions& options,
-      const LocalSlamResultCallback& local_slam_result_callback);
+  MapBuilder(const proto::MapBuilderOptions& options,
+             const LocalSlamResultCallback& local_slam_result_callback);
   ~MapBuilder();
 
   MapBuilder(const MapBuilder&) = delete;
--- a/cartographer/mapping/map_builder.h
+++ b/cartographer/mapping/map_builder.h
@@ -49,7 +49,12 @@ proto::MapBuilderOptions CreateMapBuilderOptions(
 // and a SparsePoseGraph for loop closure.
 class MapBuilder {
  public:
-  MapBuilder(const proto::MapBuilderOptions& options);
+  using AccumulatedRangeDataCallback =
+      mapping::GlobalTrajectoryBuilderInterface::AccumulatedRangeDataCallback;
+
+  MapBuilder(
+      const proto::MapBuilderOptions& options,
+      const AccumulatedRangeDataCallback& accumulated_range_data_callback);
   ~MapBuilder();
 
   MapBuilder(const MapBuilder&) = delete;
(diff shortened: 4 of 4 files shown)


cartographer/mapping_2d/local_trajectory_builder.cc
```diff
@@ -186,7 +186,7 @@ LocalTrajectoryBuilder::AddAccumulatedRangeData(
               {},  // 'low_resolution_point_cloud' is only used in 3D.
               {},  // 'rotational_scan_matcher_histogram' is only used in 3D.
               pose_estimate}),
-      std::move(insertion_submaps)});
+      std::move(insertion_submaps), gravity_aligned_range_data});
```
wohe: The idea was that there is no need for two ways to expose this data, i.e. that this replaces `last_pose_estimate_`. Wouldn't it make sense to push the `scan_match_callback_` to the local trajectory builder and call it from here, i.e. no need to move the data out of local trajectory builder except through the call to the callback.

ojura: As I mentioned in #508, my goal is to preserve the association of range data with the node ID (this was previously available because the range data was part of the trajectory node). We don't know the node ID yet in the local trajectory builder.

wohe: I see. Then maybe keeping the `last_pose_estimate_` and replacing it by the callback in the global trajectory builder is the way to go. Then we would not have a `bool` but rather an optional `NodeId` depending on whether a node was inserted?

ojura: How would you make the NodeId optional? By using a smart pointer (unique)?

wohe: Yes, `std::unique_ptr<>` sounds good to me.

ojura: Updated.

cartographer/mapping/map_builder.cc
```diff
@@ -52,8 +52,12 @@ proto::MapBuilderOptions CreateMapBuilderOptions(
   return options;
 }
 
-MapBuilder::MapBuilder(const proto::MapBuilderOptions& options)
-    : options_(options), thread_pool_(options.num_background_threads()) {
+MapBuilder::MapBuilder(
+    const proto::MapBuilderOptions& options,
+    std::unique_ptr<InsertedRangeDataCallback> inserted_range_data_callback)
```
wohe: Why the `unique_ptr`? Seems unnecessary to me.

cartographer/mapping/global_trajectory_builder.h
```diff
@@ -29,12 +29,14 @@ template <typename LocalTrajectoryBuilder,
 class GlobalTrajectoryBuilder
     : public mapping::GlobalTrajectoryBuilderInterface {
  public:
-  GlobalTrajectoryBuilder(const LocalTrajectoryBuilderOptions& options,
-                          const int trajectory_id,
-                          SparsePoseGraph* const sparse_pose_graph)
+  GlobalTrajectoryBuilder(
+      const LocalTrajectoryBuilderOptions& options, const int trajectory_id,
+      SparsePoseGraph* const sparse_pose_graph,
+      InsertedRangeDataCallback* const inserted_range_data_callback)
```
wohe: Passing a pointer seems like premature optimization. Pass a copy?

cartographer/mapping_2d/local_trajectory_builder.h
```diff
@@ -43,6 +43,7 @@ class LocalTrajectoryBuilder {
   struct InsertionResult {
     std::shared_ptr<const mapping::TrajectoryNode::Data> constant_data;
     std::vector<std::shared_ptr<const Submap>> insertion_submaps;
+    sensor::RangeData inserted_range_data;
```
wohe: This should not be added here. We want all the data from `last_pose_estimate_` to be made available via the callback, even if there is not `InsertionResult`. This is redundant then. Moreover, we should then remove `pose_estimate()` from the global trajectory builder interface.

ojura: Ooops. I completely forgot that I can't pass the range data if the insertion result was a null pointer.

cartographer/mapping/map_builder.cc
ojura: The callback is optional, right? If you don't want to give one to libcartographer, you pass a null pointer during construction of `MapBuilder`.

cartographer/mapping_2d/local_trajectory_builder.h
ojura: Ooops. I forgot what I was doing here last week.

cartographer/mapping/global_trajectory_builder.h
ojura: Same as below - what if we did not provide a callback, which is covered by passing a null pointer?

cartographer/mapping_2d/local_trajectory_builder.h
ojura: About the contents of pose_estimate: I want the original range data just like was previously stored inside a trajectory node, not a transformed point cloud. Can I change contents of  pose_estimate struct to reflect this?

ojura: One additional thing: sometimes `nullptr` is returned because there isn't enough accumulated data. How will the global trajectory builder know not to invoke the callback, because `last_pose_estimate_` hasn't been updated in the meantime and it would not make sense to invoke the callback with the same data? Maybe by comparing `time` from the `PoseEstimate`?

cartographer/mapping/global_trajectory_builder.h
wohe: You can also compare a `std::function<>` as a bool. It compares false if it is default constructed. `if (the_function) { ... }`.

cartographer/mapping_2d/local_trajectory_builder.h
wohe: True. Maybe we should then get rid of the `last_post_estimate_` completely and return this information instead, possible containing an insertion result, i.e. either nullptr, the `MatchingResult` not containing an `InsertionResult`, or the `MatchingResult` including an `InsertionResult`? Then it would also be clearer that these are results. `pose_estimate()` seems weird since it would only be called after adding some range data anyway.

ojura: Ok. I will get on it. To clarify: `AddRangeData` should return a `std::unique_ptr<MatchingResult>` (which is `nullptr` if accumulation is incomplete); if `AddAccumulatedRangeData` is called, the returned `MatchingResult` contains the accumulated range data, the pose estimate (plus the gravity alignment quaternion for 2D) and time, and a `std::unique_ptr<InsertionResult>`: `nullptr` if motion filter has eliminated the accumulated range data, or the `InsertionResult` as before. `NodeId` remains the second argument of the callback.

wohe: Why the gravity alignment? I thought of the contents of `PoseEstimate` which contain the full 3D pose in 2D and 3D. Other than that: Yes.

ojura: If we pass `gravity_aligned_range_data` to the callback, along with `pose_estimate`, we might need  `gravity_alignment.inverse()` to correctly interpret the `gravity_aligned_range_data` with respect to some internal parts of Cartographer. E.g. `gravity_aligned_range_data` is inserted into a submap like this: 
```
submap.InsertRangeData(
      TransformRangeData(gravity_aligned_range_data,
                         transform::Embed3D(pose_estimate_2d.cast<float>()))
```
where the relationship between `pose_estimate_2d` and `pose_estimate` is the following:
`pose_estimate = transform::Embed3D(pose_estimate_2d) * gravity_alignment;`

So, you would need to multiply `pose_estimate` with `gravity_alignment.inverse()` in order to be able to exactly reconstruct how the range data was (or would be) inserted into a submap.

This is not a problem with scans which are inserted into the SPG, because `gravity_alignment` is stored into the corresponding node's `constant_data,` and you can retrieve it from there. But you would not be able to do this for scans which had been discarded by the motion filter, because there would be no corresponding trajectory node.

I am not saying that I need this, I am just being cautious. We can drop it, if you prefer.

cartographer/mapping_2d/local_trajectory_builder.h
```diff
@@ -53,21 +57,25 @@ class LocalTrajectoryBuilder {
...
   void AddImuData(const sensor::ImuData& imu_data);
   void AddOdometerData(const sensor::OdometryData& odometry_data);
 
  private:
-  std::unique_ptr<InsertionResult> AddAccumulatedRangeData(
+  std::unique_ptr<MatchingResult> AddAccumulatedRangeData(
       common::Time time, const sensor::RangeData& range_data);
   sensor::RangeData TransformAndFilterRangeData(
       const transform::Rigid3f& gravity_alignment,
       const sensor::RangeData& range_data) const;
```
ojura: A question. Why in the world is this a `const` method?

cartographer/mapping_3d/local_trajectory_builder_test.cc
```diff
@@ -251,12 +251,13 @@ class LocalTrajectoryBuilderTest : public ::testing::Test {
...
-              node.time,
-              sensor::TimedRangeData{
-                  range_data.origin, range_data.returns, {}}) != nullptr) {
-        const auto pose_estimate = local_trajectory_builder_->pose_estimate();
-        EXPECT_THAT(pose_estimate.pose, transform::IsNearly(node.pose, 1e-1));
+      std::unique_ptr<LocalTrajectoryBuilder::MatchingResult> matching_result(
+          local_trajectory_builder_->AddRangeData(
+              node.time, sensor::TimedRangeData{
+                             range_data.origin, range_data.returns, {}}));
+      if (matching_result != nullptr) {
```
ojura: This test now also performs the expect check for measurements eliminated by the motion filter (these previously had a `nullptr` for the insertion result and were not checked).

ojura:
@wohe PTAL at the new implementation, and please have a look at [the comment which has been buried in a review above](https://github.com/googlecartographer/cartographer/pull/574#discussion_r146586429).

cartographer/mapping_2d/local_trajectory_builder.h
```diff
@@ -41,10 +41,23 @@ namespace mapping_2d {
...
+  // constant node data (although the node containing this data may not exist
+  // if the motion filter blocked insertion), and an 'InsertionResult' pointer,
+  // which is null if the motion filter has blocked the accumulated range data
+  // from insertion; otherwise, it points to an 'InsertionResult' instance.
   struct InsertionResult {
-    std::shared_ptr<const mapping::TrajectoryNode::Data> constant_data;
     std::vector<std::shared_ptr<const Submap>> insertion_submaps;
   };
+  struct MatchingResult {
+    std::shared_ptr<const sensor::RangeData> range_data;
```
wohe: Why the pointer? Seems unnecessary to me.

cartographer/mapping_2d/local_trajectory_builder.cc
```diff
@@ -148,38 +149,19 @@ LocalTrajectoryBuilder::AddAccumulatedRangeData(
...
-  // InsertRangeData() since the queried values are valid for next insertion.
-  std::vector<std::shared_ptr<const Submap>> insertion_submaps;
-  for (const std::shared_ptr<Submap>& submap : active_submaps_.submaps()) {
-    insertion_submaps.push_back(submap);
-  }
-  active_submaps_.InsertRangeData(
-      TransformRangeData(gravity_aligned_range_data,
-                         transform::Embed3D(pose_estimate_2d.cast<float>())));
-
   sensor::AdaptiveVoxelFilter adaptive_voxel_filter(
```
wohe: This seems unnecessary effort if the data is not inserted into the pose graph. Rather duplicate time/pose for performance and revert the change to the `InsertionResult`?

cartographer/mapping/collated_trajectory_builder.cc
```diff
@@ -46,10 +46,6 @@ CollatedTrajectoryBuilder::CollatedTrajectoryBuilder(
 
 CollatedTrajectoryBuilder::~CollatedTrajectoryBuilder() {}
 
-const PoseEstimate& CollatedTrajectoryBuilder::pose_estimate() const {
```
wohe: This PR already is too large. Move the deletion of `pose_estimate()` to a follow up PR. This also allows us to change cartographer_ros without breaking it in between.

cartographer/mapping_2d/local_trajectory_builder.h
ojura: @wohe before we go on addressing other things you mentioned, pleave have a look at what I wrote above.

cartographer/mapping_2d/local_trajectory_builder.h
ojura: Sorry, I thought the `const` was for `AddAccumulatedRangeData`.

cartographer/mapping_2d/local_trajectory_builder.h
ojura: I wanted to ensure that unnecessary copying is avoided when transferring ownership of `range_data` to the library user. Although a unique pointer would be more appropriate since `range_data` is not used in rest of Cartographer, unlike `constant_data`. You think that we don't need pointers at all? I have always felt more comfortable when handling structures containing lots of data using pointers rather than relying on RVO or using move semantics.

cartographer/mapping_2d/local_trajectory_builder.h
wohe: 1. We never stored unfiltered data anywhere. Can you explain what you mean by that?
2. Passing the range data in different frames depending on whether it is 2D or 3D is a bad idea as it means that code using the data has to implement separate code path to handle both. There is also no benefit when changing this. The user can gets both the range data and the pose and can decide to transform it as the user wishes.

cartographer/mapping_2d/local_trajectory_builder.h
wohe: Move semantics should achieve what you want. Otherwise: Do not optimize prematurely.

cartographer/mapping_2d/local_trajectory_builder.h
wohe: 1. That's fine, keeping the range data not just the point cloud sounds like a good idea. But why not transform? I was thinking of the voxel filtering which happens before, i.e. the range data is a subset of the input point cloud.
2. Why not do as before:
```
TransformRangeData(gravity_aligned_range_data,
                         transform::Embed3D(pose_estimate_2d.cast<float>()))
```
? Then the range data is in the local map frame. If you are later interested in the range data in tracking frame, you can multiply by `pose_estimate.inverse()`.

ojura: Okay, we can do that. I did not like that at first because it was not following the way how the filtered range data was stored in the node constant data (not transformed to the local frame).

wohe: As I said the important part is to be consistent across 2D and 3D. The other option is obviously to transform the 2D range data to be in tracking frame.

ojura: This would correspond to `gravity_aligned_range_data * pose_estimate_2d gravity_alignment.inverse() * pose_estimate_2d.inverse()`, right?

```
range_data_in_local = gravity_aligned_range_data * pose_estimate_2d

pose_estimate = pose_estimate_2d * gravity_alignment
pose_estimate.inverse() = 
    gravity_alignment_inverse() * pose_estimate_2d.inverse()

range_data_in_tracking = range_data_in_local * pose_estimate.inverse() = 
    gravity_aligned_range_data * pose_estimate_2d gravity_alignment.inverse() 
    * pose_estimate_2d.inverse()
```

And it would make sense to apply the same to filtered `RangeData` stored in node `constant_data`, right?

ojura:
@wohe just a ping about whether we should store the range data in tracking or in local frame, as discussed [above](https://github.com/googlecartographer/cartographer/pull/574#discussion_r147421295).

cartographer/mapping/global_trajectory_builder_interface.h
```diff
@@ -58,6 +58,17 @@ class GlobalTrajectoryBuilderInterface {
   virtual void AddSensorData(const sensor::OdometryData& odometry_data) = 0;
   virtual void AddSensorData(
       const sensor::FixedFramePoseData& fixed_frame_pose) = 0;
+
+  // An optional callback which is called when local SLAM finishes processing
```
wohe: Comments should be kept short and concise. There is the danger that when code changes in the future, comments go out of sync. How about this:
```
  // A callback which is called when local SLAM processed an accumulated
  // 'sensor::RangeData'. If the data was inserted into a submap,
  // reports the assigned 'NodeId', otherwise 'nullptr' if the data was filtered
  // out.
```

cartographer/mapping_3d/local_trajectory_builder.cc
```diff
@@ -224,7 +226,7 @@ LocalTrajectoryBuilder::InsertIntoSubmap(
               high_resolution_point_cloud,
               low_resolution_point_cloud,
               rotational_scan_matcher_histogram,
-              pose_observation}),
+              pose_estimate}),
```
wohe: For the future: Changes like these can easily be done in a separate PR. This one is already large enough.

cartographer/mapping_3d/local_trajectory_builder.cc
```diff
@@ -195,13 +198,13 @@ const mapping::PoseEstimate& LocalTrajectoryBuilder::pose_estimate() const {
 
 std::unique_ptr<LocalTrajectoryBuilder::InsertionResult>
 LocalTrajectoryBuilder::InsertIntoSubmap(
-    const common::Time time, const sensor::RangeData& range_data_in_tracking,
+    const common::Time time, const sensor::RangeData& filtered_range_data,
```
wohe: Revert? That it is filtered seems unimportant to understand what this function does, but that it still has to be transformed is important.

cartographer/mapping_2d/local_trajectory_builder.h
wohe: From your explanation it sounds you should keep it like it is. If there is not a good reason why something should be changed, we should not change it. This feels like bikeshedding otherwise.

Regarding the transform: Points get multiplied to the very right, not the left. Your last equality is wrong. `pose_estimated_2d` is from gravity-aligned to local map frame. `gravity_alignment` from tracking to gravity-aligned frame, so:

1. `gravity_aligned_range_data` transformed by `pose_estimate_2d` is in local map frame.
2. `gravity_aligned_range_data` transformed by `gravity_alignment.inverse()` is in tracking frame.

Also: Please do not change `constant_data`. It contains the data in the suitable format for loop closure detection and is furthermore completely unrelated to this PR.

ojura: Okay. I am going to introduce `range_data_in_local`, pass it to `InsertIntoSubmap` by const reference, and then `move` it into `MatchingResult`.

cartographer/mapping_3d/local_trajectory_builder.cc
ojura: I found it very unredable that the unfiltered variable in `AddAccumulatedRangeData` is also called `range_data_in_tracking`; and then `filtered_range_data` is passed again as `range_data_in_tracking` in `InsertIntoSubmap`. Maybe `filtered_range_data_in_tracking`?

wohe: I'm not feeling strongly about this. I personally would always chose conciseness if in doubt, but that is just me. That not every name is instantly understood by everyone is unavoidable. More of a question of how big the problem is here.

ojura: I think it's a subtle detail which should be made clearer, that filtered scans are inserted into submaps. If you don't mind, I'd like to keep it.

ojura:
Updated, PTAL. If you want, I can split the minor renames in 3D into a separate PR.

wohe:
Thanks a lot for your quick responses. To allow this PR to be merged, please only address review comments in this PR. It is a good idea to split off all the orthogonal changes, but even better would be to postpone unrelated change to future PRs. I would love to merge more of your code quicker, but I feel this PR is a moving target and is already too large. A good explanation of my view on a successful review process: https://www.bignerdranch.com/blog/growing-a-code-review-culture/

ojura:
I will split up this PR.

ojura:
Ok, I have split it up. This PR depends on #619.

cartographer/mapping/global_trajectory_builder_interface.h
```diff
@@ -58,6 +58,15 @@ class GlobalTrajectoryBuilderInterface {
...
       const sensor::FixedFramePoseData& fixed_frame_pose) = 0;
+
+  // A callback which is called after local SLAM processes an accumulated
+  // 'sensor::RangeData'. If the data was inserted into a submap, reports the
+  // assigned 'NodeId', otherwise 'nullptr' if the data was filtered out.
+  using AccumulatedRangeDataCallback =
+      std::function<void(int /* trajectory ID */, common::Time,
+                         transform::Rigid3d /* local pose estimate */,
+                         sensor::RangeData /* in local frame */,
+                         std::unique_ptr<const mapping::NodeId>)>;
```
cschuet: @ojura @wohe I am currently working on the design for cloud-based SLAM and for that I need to forward all the information necessary to formulate the global SLAM problem in the cloud. Part of that is the InsertionResult. So I am wondering whether we can extend this callback to also include the InsertionResult.

ojura: Sure, why not.

cschuet: e.g.

`std::function<void(int /* trajectory ID */, MatchingResult)>`

cschuet: Let's see what Wolfgang thinks. I realize I am somewhat late to the party :)

ojura: No, you're just in time :) Fell free to submit a PR over at [larics/cartographer](https://github.com/larics/cartographer) for the `range-data-callback` branch.

ojura: @cshuet On the other hand, there doesn't exist a single `MatchingResult` (or `InsertionResult`) class. There are two separate ones, one for 2D and one for 3D local trajectory builder. `GlobalTrajectoryBuilder` is templated. Perhaps changing that belongs in another Cartographer PR.

cschuet:
@ojura I was thinking the same. Go ahead with yours. I will expand as
necessary in a separate PR.

cartographer/mapping/global_trajectory_builder_interface.h
```diff
@@ -58,6 +58,15 @@ class GlobalTrajectoryBuilderInterface {
   virtual void AddSensorData(const sensor::OdometryData& odometry_data) = 0;
   virtual void AddSensorData(
       const sensor::FixedFramePoseData& fixed_frame_pose) = 0;
+
+  // A callback which is called after local SLAM processes an accumulated
+  // 'sensor::RangeData'. If the data was inserted into a submap, reports the
+  // assigned 'NodeId', otherwise 'nullptr' if the data was filtered out.
+  using AccumulatedRangeDataCallback =
```
wohe: The range data is only part of what this callback gets as information. `LocalSlamResultCallback` or `LocalSlamCallback` maybe? This would also better match your comment. @cschuet Opinion?

cschuet: I would prefer `LocalSlamResultCallback` but both are fine for me.

ojura: The "range data" in the callback name refers not only to the contained range data that is passed to the callback, but to the fact that the callback has been called right after this data has been processed (the callback is called from `AddRangefinderData()` after all). The callback is not called after e.g. adding IMU or odometry data, so I thought this name would emphasize this.

But the question is - is that important? The user probably just cares about the pose estimate the most, and probably does not care about the fact that it is updated only after range data is added - it's an internal libcartographer detail. In that case I agree with both of your suggestions. 

On the other hand, if we do want to hint that this is closely related to processing range data, I suggest we go something with "range data" in the name.

wohe: Let's call it `LocalSlamResultCallback` then.

ojura:
@wohe Renamed.

cartographer/mapping/map_builder.h
```diff
@@ -49,12 +49,12 @@ proto::MapBuilderOptions CreateMapBuilderOptions(
 // and a SparsePoseGraph for loop closure.
 class MapBuilder {
  public:
-  using AccumulatedRangeDataCallback =
-      mapping::GlobalTrajectoryBuilderInterface::AccumulatedRangeDataCallback;
+  using LocalSlamResultCallback =
+      mapping::GlobalTrajectoryBuilderInterface::LocalSlamResultCallback;
```
wohe: Can't we drop the `mapping::` prefix?

ojura:
Updated

wohe:
@wally-the-cartographer merge

wally-the-cartographer:

PR #552, opened 2017-09-29
Deserialize trajectory connectivity

ojura:

PR #575, opened 2017-10-06
Fix crashing bug in localization.

wohe:
Now that we deserialize trajectory nodes, we need to make sure
that the global matcher sampler is added for the map trajectory.


ojura:
Oops. I actually handled this in #552. That's the curse of splitting everything in many PRs.

PR #576, opened 2017-10-06
Apply SPG serialization ID remapping in MapBuilder::SerializeState.

ojura:
Related to #567. @wohe will this alleviate your concerns about that [`CHECK`](https://github.com/googlecartographer/cartographer/pull/567#discussion_r142923769) in #567 breaking loading of states with trimmed data?

I am not sure if extracting the remapping by passing a pointer to `SPG::ToProto()` is the most elegant thing to do, I am open to suggestions.

### DIFF
--- a/cartographer/mapping/map_builder.cc
+++ b/cartographer/mapping/map_builder.cc
@@ -135,7 +135,7 @@ string MapBuilder::SubmapToProto(const mapping::SubmapId& submap_id,
   }
 
   const auto submap_data = sparse_pose_graph_->GetSubmapData(submap_id);
-  if (submap_data.submap == nullptr) {
+  if (submap_data.trimmed()) {
     return "Requested submap " + std::to_string(submap_id.submap_index) +
            " from trajectory " + std::to_string(submap_id.trajectory_id) +
            " but it has been trimmed.";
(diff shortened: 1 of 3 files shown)


wohe:
Thanks for the PR @ojura, but I think we should implement this differently. The current serialization with the remapping was suitable when trimming the start of a trajectory for the localization use case. Given that we now also want to trim in the middle of a trajectory, the serialization should follow more closely the C++ implementation, i.e. serialize an index with each data without remapping. We did #577 and #578 to show where we would like to end up and will continue to refactor the code to keep data in this way.

ojura:
Okay, but until trimming in the middle gets implemented, can we stick with the current serialization remapping so all these serialization PRs can get merged? Maintaining them all against the upstream is a bit tiring.

wohe:
The problem with this PR as it is is that it moves us away from supporting trimming in the middle of a trajectory. It should rather move us towards it or at least not away from it. This also does not seem hard to do, i.e. remove the remapping and serialize the node and submap indices. If you'd prefer us to write a PR in this direction, just let me know.

ojura:
Please do write the PR, it will be much appreciated. I will be away for a day or two, so I will update the rest of the PRs when I get back. Closing this.

wohe:
@cschuet Do you want to look into implementing this?

ojura:
Hi, I'm back. Have you managed to prepare a PR? I can now start looking into this, but if you're already on it, I can hold off so we avoid double work.

cschuet:
#592 #593 #596 #597 landed yesterday. I have another CL that does what #597 does for submaps for nodes. But @wohe is out today, so this won't land until tomorrow.

PR #578, opened 2017-10-11
Introduces mapping::MapById in the 2D pose graph for submaps.

wohe:
PAIR=cschuet

### DIFF
--- a/cartographer/mapping/id.h
+++ b/cartographer/mapping/id.h
@@ -227,7 +227,7 @@ class MapById {
     trajectory.data_.erase(it);
   }
 
-  bool contains(const IdType& id) const {
+  bool Contains(const IdType& id) const {
     return trajectories_.count(id.trajectory_id) != 0 &&
            trajectories_.at(id.trajectory_id).data_.count(GetIndex(id)) != 0;
   }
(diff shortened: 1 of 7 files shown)


cartographer/mapping/id.h
```diff
@@ -217,6 +226,11 @@ class MapById {
     trajectory.data_.erase(it);
   }
 
+  bool contains(const IdType& id) const {
```
ojura: I may be nitpicking, but perhaps `size_t count(const IdType& id)` would be more consistent with other STL containers used in Cartographer (if you wanted to be consistent). I actually added it myself when I was updating another PR:

https://github.com/googlecartographer/cartographer/pull/481/commits/91c5c25a808c0bbd7d990d066974652d3ee7fe93#diff-617a4ae9be5a1b7b4c512ed01c3ad6bdR228

On the other hand, `MapLimits` has a [`Contains()`](https://github.com/googlecartographer/cartographer/blob/b0b4f30007e8472d0b4942e18bf5400e99e166dc/cartographer/mapping_2d/map_limits.h#L79) method (should that be lowercased, since it's inlined)?

SirVer: Since the documentation states `like std::map` I agree with @ojura that `count()` is more consistent.

I think MapLimits is separate design question, since it does not model a container.

wohe: Maybe I should change the comment to "reminiscent" then? It has many obvious differences compared to `std::map`.

wohe:
PTAL

PR #583, opened 2017-10-13
Introduce proxy object for range based loops over nodes of trajectory.

cschuet:


cartographer/mapping/id.h
```diff
@@ -113,6 +113,19 @@ class NestedVectorsById {
   std::vector<std::vector<ValueType>> data_;
 };
 
+template <typename IteratorType>
+class Range {
+ public:
+  Range(IteratorType begin, IteratorType end) : begin_(begin), end_(end) {}
+
+  IteratorType begin() { return begin_; }
```
wohe: Make the member functions `const`?

cartographer/mapping/id.h
```diff
@@ -113,6 +113,19 @@ class NestedVectorsById {
   std::vector<std::vector<ValueType>> data_;
 };
 
+template <typename IteratorType>
+class Range {
+ public:
+  Range(IteratorType begin, IteratorType end) : begin_(begin), end_(end) {}
```
wohe: Arguments as `const&`?

PR #584, opened 2017-10-13
Introduces mapping::MapById for submaps in 3D.

wohe:


cartographer/mapping_3d/sparse_pose_graph/optimization_problem.cc
```diff
@@ -150,43 +144,37 @@ void OptimizationProblem::Solve(const std::vector<Constraint>& constraints,
...
-            C_submaps[trajectory_id].at(submap_index).translation());
-      }
+  for (const auto& submap_id_data : submap_data_) {
+    const bool frozen =
+        frozen_trajectories.count(submap_id_data.id.trajectory_id) != 0;
+    if (first_submap) {
+      first_submap = false;
+      // Fix the first submap of the first trajectory except for allowing
+      // gravity alignment.
+      C_submaps.Insert(
```
cschuet: Why not add an Emplace() to MapById?

wohe: Unfortunately it seems not trivial. With C++11 forwarding a tuple with the arguments to a constructor is complicated, it gets easier with C++14 and easy to do in C++17. Changing the `CeresPose` to a single argument being a tuple seems also less than ideal. Leaving it for now as discussed.

PR #585, opened 2017-10-16
Introduce proxy object for iteration over trajectory ids.

cschuet:


cartographer/mapping/id.h
```diff
@@ -214,6 +214,45 @@ class MapById {
...
+
+    explicit ConstTrajectoryIterator(
+        typename std::map<int, MapByIndex>::const_iterator current_trajectory)
+        : current_trajectory_(current_trajectory) {}
+
+    int operator*() const {
+      return current_trajectory_->first;
+    }
+
+    int* operator->() const {
```
wohe: Remove as discussed since there are no members?

cartographer/mapping/id.h
```diff
@@ -214,6 +214,45 @@ class MapById {
     typename std::map<int, DataType>::const_iterator current_data_;
   };
 
+  class ConstTrajectoryIterator {
+   public:
+    using iterator_category = std::bidirectional_iterator_tag;
+    using value_type = int;
+    using difference_type = int64;
```
wohe: Add `reference` and `pointer`?

cartographer/mapping/id.h
```diff
@@ -281,6 +320,13 @@ class MapById {
                                 EndOfTrajectory(trajectory_id));
   }
 
+  // Returns Range object for range-based loops over the trajectory ids.
```
wohe: nit: `IDs`.

cartographer/mapping/id_test.cc
```diff
@@ -79,6 +79,45 @@ TEST(IdTest, MapByIdTrajectoryRange) {
   EXPECT_TRUE(expected_data.empty());
 }
 
+TEST(IdTest, MapByIdTrajectoryIdRange) {
+  MapById<NodeId, int> map_by_id;
+  map_by_id.Append(7, 2);
```
wohe: Since we reuse the same `MapById` many times now: Extract a test fixture?

cartographer/mapping/id_test.cc
```diff
@@ -79,6 +79,45 @@ TEST(IdTest, MapByIdTrajectoryRange) {
...
+  map_by_id.Append(42, 3);
+  map_by_id.Append(0, 0);
+  map_by_id.Append(0, 1);
+
+  std::deque<int> expected_data = {0, 7, 42};
+  for (int trajectory_id : map_by_id.trajectory_ids()) {
```
wohe: `const int`?

cartographer/mapping/id.h

cartographer/mapping/id.h

cartographer/mapping/id.h

cartographer/mapping/id_test.cc
cschuet: Done

cartographer/mapping/id_test.cc
cschuet: Done

PR #586, opened 2017-10-16
Add FindChecked to MapById.

cschuet:


cartographer/mapping/id.h
```diff
@@ -158,6 +158,16 @@ class MapById {
       }
     }
 
+    explicit ConstIterator(const MapById& map_by_id, const NodeId& id)
```
wohe: `NodeId` -> `IdType`
Also: Add a test using a `SubmapId`?

cartographer/mapping/id.h
```diff
@@ -225,6 +235,11 @@ class MapById {
     return IdType{trajectory_id, index};
   }
 
+  // Returs an iterator to the element at id or throws an exception.
```
wohe: typo: `Returs` -> `Returns`
`id` -> `'id'`
We use a `CHECK`, so: `or throws an exception` -> `which must exist`.

cartographer/mapping/id.h

cartographer/mapping/id.h

PR #590, opened 2017-10-16
Pose renames.

cschuet:
Renames:
 - initial_pose to local_pose
 - optimized_pose to global_pose

### DIFF
--- a/cartographer/mapping/sparse_pose_graph.h
+++ b/cartographer/mapping/sparse_pose_graph.h
@@ -77,10 +77,10 @@ class SparsePoseGraph {
   // Freezes a trajectory. Poses in this trajectory will not be optimized.
   virtual void FreezeTrajectory(int trajectory_id) = 0;
 
-  // Adds a 'submap' from a proto with the given 'initial_pose' to the frozen
+  // Adds a 'submap' from a proto with the given 'local_pose' to the frozen
   // trajectory with 'trajectory_id'.
   virtual void AddSubmapFromProto(int trajectory_id,
-                                  const transform::Rigid3d& initial_pose,
+                                  const transform::Rigid3d& local_pose,
                                   const proto::Submap& submap) = 0;
 
   // Adds a 'node' from a proto with the given 'pose' to the frozen trajectory
--- a/cartographer/mapping/trajectory_node.h
+++ b/cartographer/mapping/trajectory_node.h
@@ -47,7 +47,7 @@ struct TrajectoryNode {
     Eigen::VectorXf rotational_scan_matcher_histogram;
 
     // The initial unoptimized node pose.
-    transform::Rigid3d initial_pose;
+    transform::Rigid3d local_pose;
   };
 
   common::Time time() const { return constant_data->time; }
--- a/cartographer/mapping_2d/scan_matching/ceres_scan_matcher.cc
+++ b/cartographer/mapping_2d/scan_matching/ceres_scan_matcher.cc
@@ -60,14 +60,14 @@ CeresScanMatcher::CeresScanMatcher(
 CeresScanMatcher::~CeresScanMatcher() {}
 
 void CeresScanMatcher::Match(const transform::Rigid2d& previous_pose,
-                             const transform::Rigid2d& initial_pose_estimate,
+                             const transform::Rigid2d& local_pose_estimate,
                              const sensor::PointCloud& point_cloud,
                              const ProbabilityGrid& probability_grid,
                              transform::Rigid2d* const pose_estimate,
                              ceres::Solver::Summary* const summary) const {
-  double ceres_pose_estimate[3] = {initial_pose_estimate.translation().x(),
-                                   initial_pose_estimate.translation().y(),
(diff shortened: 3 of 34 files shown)


cartographer/mapping/sparse_pose_graph.h
```diff
@@ -77,10 +77,10 @@ class SparsePoseGraph {
   // Freezes a trajectory. Poses in this trajectory will not be optimized.
   virtual void FreezeTrajectory(int trajectory_id) = 0;
 
-  // Adds a 'submap' from a proto with the given 'initial_pose' to the frozen
+  // Adds a 'submap' from a proto with the given 'local_pose' to the frozen
```
wohe: Rename this to `pose` or `global_pose`?

cartographer/mapping/trajectory_node.h
```diff
@@ -47,7 +47,7 @@ struct TrajectoryNode {
     Eigen::VectorXf rotational_scan_matcher_histogram;
 
     // The initial unoptimized node pose.
```
wohe: Change the comment: The pose in local map frame?

cartographer/mapping_2d/scan_matching/ceres_scan_matcher.cc
```diff
@@ -60,14 +60,14 @@ CeresScanMatcher::CeresScanMatcher(
 CeresScanMatcher::~CeresScanMatcher() {}
 
 void CeresScanMatcher::Match(const transform::Rigid2d& previous_pose,
-                             const transform::Rigid2d& initial_pose_estimate,
+                             const transform::Rigid2d& local_pose_estimate,
```
wohe: Let's drop the `local_` prefix here. It depends on the frame in which the `probability_grid` is expressed, right now the local SLAM frame in 2D, but the submap frame in 3D and we might want to change it in the future. Maybe we retain `initial_` here, but now in the meaning of starting point? Similar in the other scan matchers.

cartographer/mapping_2d/sparse_pose_graph.cc
```diff
@@ -222,15 +222,15 @@ void SparsePoseGraph::ComputeConstraintsForScan(
   const mapping::SubmapId matching_id = submap_ids.front();
   const auto& constant_data = trajectory_nodes_.at(node_id).constant_data;
   const transform::Rigid2d pose = transform::Project2D(
-      constant_data->initial_pose *
+      constant_data->local_pose *
       transform::Rigid3d::Rotation(constant_data->gravity_alignment.inverse()));
-  const transform::Rigid2d optimized_pose =
+  const transform::Rigid2d global_pose =
       optimization_problem_.submap_data().at(matching_id).pose *
       sparse_pose_graph::ComputeSubmapPose(*insertion_submaps.front())
```
wohe: We should probably also rename `ComputeSubmapPose()`. Maybe `ComputeLocalSubmapPose()`?

cartographer/mapping_2d/sparse_pose_graph/optimization_problem.cc
```diff
@@ -80,10 +80,10 @@ void OptimizationProblem::AddOdometerData(
 
 void OptimizationProblem::AddTrajectoryNode(
     const int trajectory_id, const common::Time time,
-    const transform::Rigid2d& initial_pose, const transform::Rigid2d& pose,
+    const transform::Rigid2d& local_pose, const transform::Rigid2d& pose,
```
wohe: `pose` -> `global_pose`?

cartographer/mapping_2d/sparse_pose_graph/constraint_builder.cc
```diff
@@ -162,14 +162,14 @@ void ConstraintBuilder::ComputeConstraint(
     const mapping::TrajectoryNode::Data* const constant_data,
     const transform::Rigid2d& initial_relative_pose,
     std::unique_ptr<ConstraintBuilder::Constraint>* constraint) {
-  const transform::Rigid2d initial_pose =
+  const transform::Rigid2d local_pose =
```
wohe: I wonder what a good naming is here. This an `pose_estimate` are in local map frame. One is the starting point, the other the result of the scan matchers.

cartographer/mapping_3d/local_trajectory_builder.cc
```diff
@@ -133,9 +133,9 @@ LocalTrajectoryBuilder::AddAccumulatedRangeData(
       adaptive_voxel_filter.Filter(filtered_range_data.returns);
   if (options_.use_online_correlative_scan_matching()) {
     // We take a copy since we use 'initial_ceres_pose' as an output argument.
-    const transform::Rigid3d initial_pose = initial_ceres_pose;
+    const transform::Rigid3d local_pose = initial_ceres_pose;
```
wohe: In 3D, we do scan-to-submap matching in the submap frame, so this is not a local pose, but something different. `initial_pose_in_submap` maybe similar to `pose_observation_in_submap` below?

Issue #594, opened 2017-10-17
Trim odometry data from optimization_problem


jihoonl:
TrimTrajectoryNode [2d](https://github.com/googlecartographer/cartographer/blob/master/cartographer/mapping_2d/sparse_pose_graph/optimization_problem.cc#L89), [3d](https://github.com/googlecartographer/cartographer/blob/master/cartographer/mapping_3d/sparse_pose_graph/optimization_problem.cc#L92) in optimization problem should also trim obsolte odometry data.

wohe:
We added some projects to better communicate our short term roadmap. See https://github.com/googlecartographer/cartographer/projects/1. Our intention is to address trimming arbitrary parts of the odometry data.

wohe:
Fixed by #655.

PR #595, opened 2017-10-17
 Add PopFronts() in TransformInterpolationBuffer

jihoonl:
Towards #594


wohe:
Thanks for the PR, but we would like to address our intended roadmap directly, i.e. move to a data structure allowing trimming the middle. @cschuet will work on this next week.

PR #596, opened 2017-10-17
Remove trimmed() from TrajectoryNode

cschuet:
With the switch to MapById this is now obsolete.

### DIFF
--- a/cartographer/mapping/sparse_pose_graph.cc
+++ b/cartographer/mapping/sparse_pose_graph.cc
@@ -79,7 +79,7 @@ proto::SparsePoseGraph SparsePoseGraph::ToProto() {
     for (size_t old_node_index = 0;
          old_node_index != single_trajectory_nodes.size(); ++old_node_index) {
       const auto& node = single_trajectory_nodes[old_node_index];
-      if (node.constant_data) {
+      if (node.constant_data != nullptr) {
         node_id_remapping[NodeId{static_cast<int>(trajectory_id),
                                  static_cast<int>(old_node_index)}] =
             NodeId{static_cast<int>(trajectory_id),
--- a/cartographer/mapping/sparse_pose_graph.cc
+++ b/cartographer/mapping/sparse_pose_graph.cc
@@ -79,16 +79,14 @@ proto::SparsePoseGraph SparsePoseGraph::ToProto() {
     for (size_t old_node_index = 0;
          old_node_index != single_trajectory_nodes.size(); ++old_node_index) {
       const auto& node = single_trajectory_nodes[old_node_index];
-      if (!node.trimmed()) {
-        node_id_remapping[NodeId{static_cast<int>(trajectory_id),
-                                 static_cast<int>(old_node_index)}] =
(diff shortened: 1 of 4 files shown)


cartographer/mapping_2d/sparse_pose_graph.cc
```diff
@@ -651,7 +650,6 @@ void SparsePoseGraph::TrimmingHandle::MarkSubmapAsTrimmed(
 
   // Mark the 'nodes_to_remove' as trimmed and remove their data.
   for (const mapping::NodeId& node_id : nodes_to_remove) {
-    CHECK(!parent_->trajectory_nodes_.at(node_id).trimmed());
     parent_->trajectory_nodes_.at(node_id).constant_data.reset();
```
wohe: You also have to change this code then to use `Trim()` instead of resetting the `constant_data`, right?

cartographer/mapping/sparse_pose_graph.cc
```diff
@@ -79,16 +79,14 @@ proto::SparsePoseGraph SparsePoseGraph::ToProto() {
     for (size_t old_node_index = 0;
          old_node_index != single_trajectory_nodes.size(); ++old_node_index) {
       const auto& node = single_trajectory_nodes[old_node_index];
-      if (!node.trimmed()) {
```
wohe: This is not possible until `GetTrajectoryNodes()` returns a `MapById<>` instead of `vector<vector<>>`?

cschuet: Ok. I will clean this up then when I change to MapById.

cartographer/mapping_2d/sparse_pose_graph.cc
cschuet: Done.

cartographer/mapping/sparse_pose_graph.cc
```diff
@@ -79,7 +79,7 @@ proto::SparsePoseGraph SparsePoseGraph::ToProto() {
...
+      if (node.constant_data) {
```
wohe: Consistency: Be explicit `!= nullptr`

cartographer/mapping_2d/sparse_pose_graph.cc
```diff
@@ -649,10 +648,9 @@ void SparsePoseGraph::TrimmingHandle::MarkSubmapAsTrimmed(
   parent_->constraint_builder_.DeleteScanMatcher(submap_id);
   parent_->optimization_problem_.TrimSubmap(submap_id);
 
-  // Mark the 'nodes_to_remove' as trimmed and remove their data.
+  // Remove the 'nodes_to_remove' from the SPG and the optimization problem.
```
wohe: Style: Eschew abbreviations: pose graph.

cschuet: Done

cartographer/mapping/sparse_pose_graph.cc
cschuet: Done

PR #601, opened 2017-10-18
Add time to points in PointCloud.

### LINKED ISSUE #573: Add timing offsets to range data.
SLAM quality can be improved by applying the transformation most correct in time to each point in range data. For good performance, we do not want to pass each point individually timed through the pipeline, but instead pass range data where each point can have a time offset.

Initial implementation:
1. Add a 4th component to range data representing a time offset in seconds, similar to #571.
2. Handle this data in voxel filtering and accumulation in the local trajectory builders.

Later it might make sense to also change accumulation to be based on time instead of number of range data, or to change the local trajectory builder to take time offsets into account during scan matching (without the need for an unwarping step beforehand). Also, the origin might be interpreted to be at time offset zero, and transformed accordingly for improved quality ray casting.

### LINKED ISSUE #600: Add time component to points in PointCloud.
Points are not scanned at the same time, but continuously.
In order to improve local quality, we would like to build the local trajectory from timed points.
This requires that PointCloud (and RangeData) store the relative timing of points.
Let's make PointCloud store points as Vector4f and put the relative time in the fourth component.

gaschler:
Adds a type TimedPointCloud that holds 4-dimensional vectors where measurement time of individual points are stored in the fourth entry.
Uses TimedPointCloud for incoming measurements so measurement time is visible to LocalTrajectoryBuilder.

Fixes #573.

### DIFF
--- a/cartographer/sensor/point_cloud.cc
+++ b/cartographer/sensor/point_cloud.cc
@@ -32,9 +32,33 @@ PointCloud TransformPointCloud(const PointCloud& point_cloud,
   return result;
 }
 
+TimedPointCloud TransformPointCloud(const TimedPointCloud& point_cloud,
+                                    const transform::Rigid3f& transform) {
+  TimedPointCloud result;
+  result.reserve(point_cloud.size());
+  for (const Eigen::Vector4f& point : point_cloud) {
+    Eigen::Vector4f result_point;
+    result_point.head<3>() = transform * point.head<3>();
+    result_point[3] = point[3];
+    result.emplace_back(result_point);
+  }
+  return result;
+}
+
 PointCloud Crop(const PointCloud& point_cloud, const float min_z,
                 const float max_z) {
   PointCloud cropped_point_cloud;
+  for (const auto& point : point_cloud) {
+    if (min_z <= point.z() && point.z() <= max_z) {
+      cropped_point_cloud.push_back(point);
+    }
+  }
+  return cropped_point_cloud;
+}
+
+TimedPointCloud Crop(const TimedPointCloud& point_cloud, const float min_z,
+                     const float max_z) {
+  TimedPointCloud cropped_point_cloud;
   for (const auto& point : point_cloud) {
     if (min_z <= point.z() && point.z() <= max_z) {
       cropped_point_cloud.push_back(point);
--- a/cartographer/sensor/point_cloud.cc
+++ b/cartographer/sensor/point_cloud.cc
@@ -32,8 +32,8 @@ PointCloud TransformPointCloud(const PointCloud& point_cloud,
   return result;
 }
 
-TimedPointCloud TransformPointCloud(const TimedPointCloud& point_cloud,
-                                    const transform::Rigid3f& transform) {
+TimedPointCloud TransformTimedPointCloud(const TimedPointCloud& point_cloud,
+                                         const transform::Rigid3f& transform) {
   TimedPointCloud result;
   result.reserve(point_cloud.size());
   for (const Eigen::Vector4f& point : point_cloud) {
@@ -45,8 +45,8 @@ TimedPointCloud TransformPointCloud(const TimedPointCloud& point_cloud,
   return result;
 }
 
-PointCloud Crop(const PointCloud& point_cloud, const float min_z,
-                const float max_z) {
+PointCloud CropPointCloud(const PointCloud& point_cloud, const float min_z,
+                          const float max_z) {
   PointCloud cropped_point_cloud;
   for (const auto& point : point_cloud) {
     if (min_z <= point.z() && point.z() <= max_z) {
@@ -56,8 +56,8 @@ PointCloud Crop(const PointCloud& point_cloud, const float min_z,
   return cropped_point_cloud;
 }
 
-TimedPointCloud Crop(const TimedPointCloud& point_cloud, const float min_z,
-                     const float max_z) {
+TimedPointCloud CropTimedPointCloud(const TimedPointCloud& point_cloud,
+                                    const float min_z, const float max_z) {
   TimedPointCloud cropped_point_cloud;
   for (const auto& point : point_cloud) {
     if (min_z <= point.z() && point.z() <= max_z) {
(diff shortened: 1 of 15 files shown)


gaschler:
Only merge this PR when the respective ROS component change googlecartographer/cartographer_ros#549 is ready.

cartographer/io/hybrid_grid_points_processor.cc
```diff
@@ -39,8 +39,12 @@ HybridGridPointsProcessor::FromDictionary(
 }
 
 void HybridGridPointsProcessor::Process(std::unique_ptr<PointsBatch> batch) {
-  range_data_inserter_.Insert({batch->origin, batch->points, {}},
-                              &hybrid_grid_);
+  // TODO(gaschler): Move this to a helper function.
+  sensor::PointCloud points_time(batch->points.size(), Eigen::Vector4f::Zero());
```
SirVer: RangeDataInserter should only deal with a PointCloud, not with a TimedPointCloud anymore.

cartographer/sensor/point_cloud.h
```diff
@@ -27,7 +27,14 @@
...
 
-typedef std::vector<Eigen::Vector3f> PointCloud;
+// Stores 3D positions of points with their measurement time
+// in the fourth entry.
+// Measurement time should be increasing and relative to the most
+// current measurement. The most current time entry is 0.f.
+// If timing is not available, all fourth entries are 0.f.
+// For 2D points, the third entry is 0.f and the fourth entry is time.
+// TODO(gaschler): Define a type PositionTime?
+typedef std::vector<Eigen::Vector4f> PointCloud;
```
SirVer: As discussed, rename to TimedPointCloud. Also beef up the documentation: What is the unit of the fourth component?

We still retain PointCloud - with `origin` being the origin of the first point and the rest of the poinst being unwarped as well as possible.

cartographer/sensor/point_cloud.h
```diff
@@ -27,7 +27,14 @@
...
 
 struct PointCloudWithIntensities {
```
SirVer: Remove 'offset_seconds' and make `points` a TimedPointCLoud

gaschler: Done

cartographer/sensor/point_cloud.h
gaschler: Done, added documentation.
Done, added PointCloud again.

cartographer/io/hybrid_grid_points_processor.cc
gaschler: Done

gaschler:
Thanks for your comments. PTAL at this alternative solution. (It is possible to split it up into two commits.)

cartographer/sensor/point_cloud.h
```diff
@@ -27,26 +27,41 @@
 namespace cartographer {
 namespace sensor {
 
+// Stores 3D positions of points.
+// For 2D points, the third entry is 0.f.
+// Used for unwarped point clouds but not for sensor measurements.
 typedef std::vector<Eigen::Vector3f> PointCloud;
 
+// Stores 3D positions of points with their measurement time
```
SirVer: Reflow comment to 80 columns.

cartographer/sensor/point_cloud.h
```diff
@@ -27,26 +27,41 @@
...
+// in the fourth entry.
+// Time should be measured in seconds, increasing and relative to the most
```
SirVer: I do not understand this comment. The most current, i.e. closest to now measurement is the last which does not have dt = 0'. How about:

Time is in seconds, increasing and relative to the moment 'points[0]' was acquired.

cartographer/sensor/range_data.cc
```diff
@@ -31,11 +31,27 @@ RangeData TransformRangeData(const RangeData& range_data,
   };
 }
 
+TimedRangeData TransformRangeData(const TimedRangeData& range_data,
```
SirVer: consistency: TransformTimedRangeData

SirVer: consistency: TransformTimedRangeData

cartographer/sensor/range_data.h
```diff
@@ -35,12 +35,31 @@ struct RangeData {
   PointCloud misses;
 };
 
+// Rays begin at 'origin'. 'returns' are the points where obstructions were
```
SirVer: Simplify comment: 'Like 'RangeData', but with 'TimedPointClouds'.

cartographer/sensor/point_cloud.cc
```diff
@@ -32,6 +32,19 @@ PointCloud TransformPointCloud(const PointCloud& point_cloud,
   return result;
 }
 
+TimedPointCloud TransformPointCloud(const TimedPointCloud& point_cloud,
```
SirVer: Consistency: TransformTimedPointCloud

cartographer/sensor/point_cloud.cc
```diff
@@ -43,5 +56,16 @@ PointCloud Crop(const PointCloud& point_cloud, const float min_z,
   return cropped_point_cloud;
 }
 
+TimedPointCloud Crop(const TimedPointCloud& point_cloud, const float min_z,
```
SirVer: Rename Crop to CropPointCloud and rename to CropTimedPointCloud.

cartographer/sensor/range_data.cc
```diff
@@ -31,11 +31,27 @@ RangeData TransformRangeData(const RangeData& range_data,
...
+  };
+}
+
 RangeData CropRangeData(const RangeData& range_data, const float min_z,
                         const float max_z) {
   return RangeData{range_data.origin, Crop(range_data.returns, min_z, max_z),
                    Crop(range_data.misses, min_z, max_z)};
 }
 
+TimedRangeData CropRangeData(const TimedRangeData& range_data,
```
SirVer: consistency: CropTimedRangeData.

gaschler:
All done, please take another look.

PR #604, opened 2017-10-19
Remove unused concept of CompressedRangeData.

SirVer:


ojura:
This might still be useful for users of #574.

wohe:
@ojura Can you give context? What do you want to use it for? Note that we still have `CompressedPointCloud` and do not intend to remove that.

ojura:
Ok, good. I imagined that if you want to store the `sensor::RangeData` you received through the callback, you might want to call `sensor::Compress` on it.

PR #605, opened 2017-10-19
Add map trajectory without a trajectory builder.

wohe:


ojura:
Please take a look at my response in #549. I am using the map resolution settings (which is contained in the trajectory builder) for realtime map generation using the old assets writer code. So I would prefer if we had this information available when the trajectory is restored.

wohe:
@ojura This is not removing any information. Before we had to create some fake options to make some `CHECK`s not fail.

PR #610, opened 2017-10-23
Adds lower_bound() to MapById.

cschuet:


cartographer/mapping/id.h
```diff
@@ -322,6 +323,43 @@ class MapById {
 
   bool empty() const { return begin() == end(); }
 
+  ConstIterator lower_bound(int trajectory_id, const common::Time& time) {
```
SirVer: documentation. This is not an obvious function - especially the interplay between the two parameters. While you are on it: document behavior of this function, `BeginOfTrajectory` and `EndOfTrajectory` for non-existing trajectories.

cartographer/mapping/id_test.cc
```diff
@@ -2,7 +2,7 @@
  * Copyright 2017 The Cartographer Authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
- * you may not use this file except in compliance with the License.
+* you may not use this file except in compliance with the License.
```
SirVer: revert line

cartographer/mapping/id.h
```diff
@@ -322,6 +323,43 @@ class MapById {
...
+    if (SizeOfTrajectoryOrZero(trajectory_id) == 0) {
+      return EndOfTrajectory(trajectory_id);
+    }
+
+    const std::map<int, DataType>& trajectory =
+        trajectories_.at(trajectory_id).data_;
+    if (std::prev(trajectory.end())->second.time() < time) {
```
SirVer: CHECK(!trajectory.empty())? Or is this an invariant that you guarantee?

cartographer/mapping/id.h
```diff
@@ -322,6 +323,43 @@ class MapById {
...
+      return EndOfTrajectory(trajectory_id);
+    } else if (!(trajectory.begin()->second.time() < time)) {
```
SirVer: Should this not be <=?

cartographer/mapping/id.h
```diff
@@ -322,6 +323,43 @@ class MapById {
...
+      return BeginOfTrajectory(trajectory_id);
+    }
+
+    int L = trajectory.begin()->first;
```
SirVer: style: `left`, `right` and `middle/pivot`

cartographer/mapping/id.h
```diff
@@ -322,6 +323,43 @@ class MapById {
...
+                               IdType{trajectory_id, lower_bound_m->first});
+        } else {
+          CHECK(M < R);
+          R = M;
+        }
+      } else {
+        CHECK(lower_bound_m->first > L);
+        L = lower_bound_m->first;
+      }
+      M = L + (R - L) / 2;
```
SirVer: move to beginning of loop, drop initialization outside and make const?

cartographer/mapping/id.h
```diff
@@ -322,6 +323,43 @@ class MapById {
...
+      return BeginOfTrajectory(trajectory_id);
+    }
+
+    int L = trajectory.begin()->first;
+    int R = std::prev(trajectory.end())->first;
+    int M = L + (R - L) / 2;
+
+    while (true) {
+      auto lower_bound_m = trajectory.lower_bound(M);
+      if (!(lower_bound_m->second.time() < time)) {
```
SirVer: again, should this not be <=. In either case, remove ! and swap bodies.

cartographer/mapping/id.h
SirVer: Also, this does not follow the semantic set by the STL: 

> Returns an iterator pointing to the first element in the range [first, last) that is not less than (i.e. greater or equal to) value, or last if no such element is found.

cartographer/mapping/id_test.cc
```diff
@@ -145,6 +160,32 @@ TEST(IdTest, FindSubmapId) {
...
+  map_by_id.Append(0, Data(1));
+  map_by_id.Append(0, Data(2));
+  map_by_id.Append(0, Data(4));
+  map_by_id.Append(0, Data(5));
+  CHECK(map_by_id.lower_bound(0, CreateTime(3)) ==
+        (MapById<SubmapId, Data>::ConstIterator(map_by_id, SubmapId{0, 2})));
+  CHECK(map_by_id.lower_bound(0, CreateTime(2)) ==
+        (MapById<SubmapId, Data>::ConstIterator(map_by_id, SubmapId{0, 1})));
+  CHECK(map_by_id.lower_bound(0, CreateTime(4)) ==
+        (MapById<SubmapId, Data>::ConstIterator(map_by_id, SubmapId{0, 2})));
```
SirVer: CHECK (0, -1000) and (0, 1000).

CHECK negative values for time.
CHECK huge container?

cartographer/mapping/id.h
cschuet: DOne

cschuet: It adds the idea that you have data grouped by trajectories and also uses the time() property for ordering rather than the keys. Should we rename it?

cartographer/mapping/id.h
cschuet: This is guaranteed by 'SizeOfTrajectoryOrZero(trajectory_id) == 0'.

cartographer/mapping/id.h
cschuet: No, as discussed.

cartographer/mapping/id.h

cartographer/mapping/id.h
cschuet: Nope.

cartographer/mapping/id.h
cschuet: Ok

cartographer/mapping/id_test.cc
cschuet: Done

cartographer/mapping/id_test.cc
wohe: Write the simple O(n) solution and test with lots of random data?

cartographer/mapping/id.h
```diff
@@ -322,6 +323,45 @@ class MapById {
...
+      return BeginOfTrajectory(trajectory_id);
+    }
+
+    int left = trajectory.begin()->first;
+    int right = std::prev(trajectory.end())->first;
+    while (true) {
+      const int middle = left + (right - left) / 2;
+      auto lower_bound_middle = trajectory.lower_bound(middle);
+      if (!(lower_bound_middle->second.time() < time)) {
+        if (std::prev(lower_bound_middle)->second.time() < time) {
```
wohe: Can we remove this extra case? That is, change `left` and `right` to mean either left or right but maybe also at the solution. Since the interesting case for binary search is the condition that excludes a potential candidate, then do
```
if (std::prev(lower_bound_middle)->second.time() < time) {
  left = lower_bound_middle->first;
} else {
  right = std::prev(lower_bound_middle)->first;
}
```
and once `left == right` we are done.

wohe: As discussed offline this is of course not correct. `middle` equals `left` in the interesting case, so it should be:
```
if (lower_bound_middle->second.time() < time) {
  left = std::next(lower_bound_middle)->first;
} else {
  right = lower_bound_middle->first;
}
```

cartographer/mapping/id_test.cc
cschuet: Done

cartographer/mapping/id_test.cc
```diff
@@ -16,16 +16,33 @@
...
 
+#include "cartographer/common/time.h"
 #include "gtest/gtest.h"
 
 namespace cartographer {
 namespace mapping {
 namespace {
 
+common::Time CreateTime(int secs) {
+  return common::Time(std::chrono::seconds(secs));
```
wohe: `common::Time(common::FromSeconds(milliseconds))`?

cartographer/mapping/id_test.cc
```diff
@@ -16,16 +16,33 @@
...
 #include <utility>
 
+#include "cartographer/common/time.h"
 #include "gtest/gtest.h"
 
 namespace cartographer {
 namespace mapping {
 namespace {
 
+common::Time CreateTime(int secs) {
```
wohe: `const int milliseconds`?

cartographer/mapping/id_test.cc
```diff
@@ -16,16 +16,33 @@
...
+}
+
+struct Data {
+  Data(int secs) : time_(CreateTime(secs)) {}
+
+  const common::Time& time() const {
+    return time_;
+  }
+
+  const common::Time time_;
```
wohe: `time` or this should be a class.

cartographer/mapping/id_test.cc
```diff
@@ -145,6 +162,65 @@ TEST(IdTest, FindSubmapId) {
...
+  CHECK(map_by_id.lower_bound(0, CreateTime(3)) ==
+        (MapById<SubmapId, Data>::ConstIterator(map_by_id, SubmapId{0, 2})));
+  CHECK(map_by_id.lower_bound(0, CreateTime(2)) ==
+        (MapById<SubmapId, Data>::ConstIterator(map_by_id, SubmapId{0, 1})));
+  CHECK(map_by_id.lower_bound(0, CreateTime(4)) ==
+        (MapById<SubmapId, Data>::ConstIterator(map_by_id, SubmapId{0, 2})));
+}
+
+TEST(IdTest, LowerBoundFuzz) {
+  constexpr int MAX_T_INCREMENT = 20;
```
wohe: Style: `kMaxTIncrement`.

cartographer/mapping/id_test.cc
```diff
@@ -145,6 +162,65 @@ TEST(IdTest, FindSubmapId) {
...
+  constexpr int MAX_N_NODES = 20;
+  constexpr int N_TESTS = 100;
+
+  std::random_device rd;
```
wohe: Unnecessary? Remove.

cartographer/mapping/id_test.cc
cschuet: Done

cartographer/mapping/id_test.cc
cschuet: Done

cartographer/mapping/id_test.cc
cschuet: Done

cartographer/mapping/id_test.cc
cschuet: Done

cartographer/mapping/id_test.cc
cschuet: Done

cartographer/mapping/id_test.cc
```diff
@@ -189,24 +191,23 @@ TEST(IdTest, LowerBound) {
 }
 
 TEST(IdTest, LowerBoundFuzz) {
-  constexpr int MAX_T_INCREMENT = 20;
-  constexpr int MAX_N_NODES = 20;
-  constexpr int N_TESTS = 100;
+  constexpr int kMaxTimeIncrement = 20;
+  constexpr int kMaxNumberOfNodes = 20;
```
wohe: Style: `kNumNodes`. `num` is a "widespread convention". Similar below.

cartographer/mapping/id_test.cc
```diff
@@ -145,6 +164,64 @@ TEST(IdTest, FindSubmapId) {
...
+  std::uniform_int_distribution<int> dt_dist(1, kMaxTimeIncrement);
+  std::uniform_int_distribution<int> N_dist(1, kMaxNumberOfNodes);
+
+  for (int i = 0; i < kNumberOfTests; ++i) {
+    const int N = N_dist(rng);
+    int t = 0;
+    MapById<SubmapId, Data> map_by_id;
+    for (int j = 0; j < N; ++j) {
+      t = t + dt_dist(rng);
+      map_by_id.Append(0, Data(t));
```
wohe: `0` -> `kTrajectoryId`?

Issue #602, opened 2017-10-18
SPG Submap serialization - extrapolated poses are serialized the same way as optimized poses


ojura:
Submaps for which optimised poses are not available are treated the same way in serialization after `GetSubmapDataUnderLock`  computes an extrapolated pose. During deserialization, in `AddSubmapFromProto`, these extrapolated poses will get injected into `optimized_submap_transforms_` with the rest of submap poses which really _were_ optimized poses. It would be more faithful with respect to the previous state to restore only optimized submap poses, while we should still be able to extrapolate the rest, just like in the original state.

cschuet:
What's your specific concern here? It seems to me that storing the extrapolated submap transforms wrongfully in `optimized_submap_transforms_` has neither a user-visible nor an algorithmic impact: after the first run of the global optimization problem Cartographer will be in exactly the same state as if we had remembered which ones were extrapolated.

SirVer:
@cschuet Serializing implies to me that writing and loading state is equivalent to just keep the system running. This is not quite correct in this case.

cschuet:
@SirVer Ok I see. No, I don't think this is what we are going for, e.g. we are not planning to support continuing to map a deserialized trajectory. I guess what I am asking is: What would you do with that additional information if we stored it?

SirVer:
Mhh, true. 

> What would you do with that additional information if we stored it?

What difference does it make right now in the online system if a pose shows up in `optimized_submap_transforms_` or not?

ojura:
That `optimized_submap_transforms_` does not influence the algorithm execution might be true, but someone might do something at some point that causes different behaviour to be exhibited. I think that it's a fragile assumption. (And there is some code in `RunOptimization()` which uses `optimized_submap_transforms_` in such a way that I would not put my hand in fire that it works exactly the same after deserialization).

cschuet:
> What difference does it make right now in the online system if a pose shows up in `optimized_submap_transforms_` or not?
Seems to me like the main purpose is to be able to provide our best guess at the global submap position for visualization. `SparsePoseGraph::GetAllSubmapData()` is called by cartographer_ros's `MapBuilderBridge`.

Ok we can keep this issue open. So the fix would be to add a `SubmapId` to our proto to remember the last optimized submap and at restoration time only copy those into `optimized_submap_transforms_` that have a SubmapId <= that number.

ojura:
[...]

PR #614, opened 2017-10-30
Do not insert extrapolated submap transforms into global_submap_poses.

ojura:
Related to #602.

### DIFF
--- a/cartographer/mapping/map_builder.cc
+++ b/cartographer/mapping/map_builder.cc
@@ -171,6 +171,7 @@ void MapBuilder::SerializeState(io::ProtoStreamWriter* const writer) {
       submap_proto->mutable_submap_id()->set_submap_index(
           submap_id_data.id.submap_index);
       submap_id_data.data.submap->ToProto(submap_proto);
+      submap_proto->set_pose_type(ToProto(submap_id_data.data.pose_type));
       writer->WriteProto(proto);
     }
   }
--- a/cartographer/mapping/pose_graph.cc
+++ b/cartographer/mapping/pose_graph.cc
@@ -48,6 +48,29 @@ PoseGraph::Constraint::Tag FromProto(
   LOG(FATAL) << "Unsupported tag.";
 }
 
+proto::SubmapPoseType ToProto(const PoseGraph::SubmapData::PoseType pose_type) {
+  switch (pose_type) {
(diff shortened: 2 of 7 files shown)


cartographer/mapping_2d/sparse_pose_graph.cc
```diff
@@ -383,19 +384,29 @@ void SparsePoseGraph::AddSubmapFromProto(const transform::Rigid3d& global_pose,
...
+  AddWorkItem([this, submap_id, global_pose_2d,
+               is_global_pose_available]() REQUIRES(mutex_) {
     CHECK_EQ(frozen_trajectories_.count(submap_id.trajectory_id), 1);
     submap_data_.at(submap_id).state = SubmapState::kFinished;
-    optimization_problem_.InsertSubmap(submap_id, global_pose_2d);
+    if (is_global_pose_available) {
+      // Probably incorrect not to do this when a global pose is not available.
+      // Perhaps we should follow GrowSubmapTransformsAsNeeded in that case?
+      // Although we don't have the previous submap available currently, so we
+      // can't compute the transform like in GrowSubmapTransformsAsNeeded...
```
ojura: See comment in code.

ojura:
The biggest problem with this PR is that when deserializing, we can't really emulate what happens with submaps that have extrapolated positions, as noted in that code comment. 

When a new submap is created, `InitializeGlobalSubmapPoses` (ex `GrowSubmapTransformsAsNeeded`) extrapolates a pose for the submap based on the last submap's local pose and [calls `optimization_problem_.AddSubmap` with the extrapolated pose:

https://github.com/googlecartographer/cartographer/blob/a4c0e4754e99e461e2497b05f234839b181fc3cc/cartographer/mapping_2d/sparse_pose_graph.cc#L83-L92

 However, in `AddSubmapFromProto` we don't have access to the previous submap, so we can't emulate what `InitializeGlobalSubmapPoses` does when calling `optimization_problem_.InsertSubmap`.

ojura:
Okay, I reworked this. The rationale is simple: if a submap pose was not in `global_submap_poses` previously, it should not be there after deserialization either.

If you are absolutely certain that this is not important and won't affect anything, we can close this PR and #602.

wohe:
Yes, I think it is not important and we can close this PR.

PR #615, opened 2017-10-30
Pull out time-based transformation interpolation.

cschuet:


cartographer/transform/timestamped_transform.cc
```diff
@@ -0,0 +1,47 @@
...
+ */
+
+#include "cartographer/transform/timestamped_transform.h"
+
+namespace cartographer {
+namespace transform {
+
+TimestampedTransform Interpolate(const TimestampedTransform& start,
+                                 const TimestampedTransform& end,
+                                 const common::Time& time) {
```
wohe: Consistency: Pass `common::Time` by value, it is just an `int64`.

cartographer/transform/timestamped_transform.cc
```diff
@@ -0,0 +1,47 @@
...
+  CHECK(start.time <= time && time <= end.time) << "Can't interpolate to time "
```
wohe: Use `CHECK_LE` and `CHECK_GE`? This is preferable since in the case of a crash we will get good debug output, i.e. the values of `start.time` or `end.time`.

cartographer/transform/timestamped_transform.cc
```diff
@@ -0,0 +1,47 @@
...
+                                                << time;
+
+  if (end.time == time) {
+    return end;
+  }
+  if (start.time == time) {
+    return start;
```
wohe: Unnecessary special cases which slow down the common case?

cartographer/transform/transform_interpolation_buffer.cc
```diff
@@ -59,23 +59,7 @@ transform::Rigid3d TransformInterpolationBuffer::Lookup(
         return timestamped_transform.time < time;
       });
   const auto end = start;
-  if (end->time == time) {
-    return end->transform;
-  }
-  --start;
```
wohe: We should keep this code, right? Add a unit test? I think this is for the case that `timestamped_transforms_.size() == 1`.

cartographer/transform/timestamped_transform.cc
cschuet: done

cartographer/transform/timestamped_transform.cc
cschuet: done

cartographer/transform/timestamped_transform.cc
cschuet: done

cartographer/transform/transform_interpolation_buffer.cc
cschuet: done

PR #616, opened 2017-10-30
Reuse previous result in extrapolator.

gaschler:
Avoids extrapolating from the last known pose buts reuses an
intermediate result from a previous ExtrapolatePose call if
time is in the future.

### DIFF
--- a/cartographer/mapping/pose_extrapolator.cc
+++ b/cartographer/mapping/pose_extrapolator.cc
@@ -75,7 +75,7 @@ void PoseExtrapolator::AddPose(const common::Time time,
   AdvanceImuTracker(time, imu_tracker_.get());
   TrimImuData();
   TrimOdometryData();
-  extrapolating_imu_tracker_.reset(new ImuTracker(*imu_tracker_));
+  extrapolating_imu_tracker_ = common::make_unique<ImuTracker>(*imu_tracker_);
 }
 
 void PoseExtrapolator::AddImuData(const sensor::ImuData& imu_data) {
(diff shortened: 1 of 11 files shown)


cartographer/mapping/pose_extrapolator.cc
```diff
@@ -75,6 +75,7 @@ void PoseExtrapolator::AddPose(const common::Time time,
   AdvanceImuTracker(time, imu_tracker_.get());
   TrimImuData();
   TrimOdometryData();
+  extrapolating_imu_tracker_.reset(new ImuTracker(*imu_tracker_));
```
wohe: Consistency: Favor `make_unique` over `new`, i.e.:
```
extrapolating_imu_tracker_ = common::make_unique<ImuTracker>(*imu_tracker_);
```

cartographer/mapping/pose_extrapolator.cc
```diff
@@ -208,10 +209,12 @@ void PoseExtrapolator::AdvanceImuTracker(const common::Time time,
 
 Eigen::Quaterniond PoseExtrapolator::ExtrapolateRotation(
     const common::Time time) {
-  ImuTracker imu_tracker = *imu_tracker_;
-  AdvanceImuTracker(time, &imu_tracker);
+  if (time < extrapolating_imu_tracker_->time()) {
+    extrapolating_imu_tracker_.reset(new ImuTracker(*imu_tracker_));
```
wohe: Did a `CHECK` not work? It would seem better to constrain usage if possible. It should be easy to ensure that we can extrapolate at the caller, and doing so has a performance benefit.

gaschler:
Note: Only merge when https://github.com/googlecartographer/cartographer_ros/pull/569 is ready.

wohe:
This is not reviewable. It is mostly #569 with one function added, right? Prefer to keep it one PR in this case, or even better: make it orthogonal.

These PR break cartographer_ros which is avoidable here: add the new function `GetLastExtrapolatedTime()` first, then change cartographer_ros, then add the caching.

gaschler:
Abandoned in favor of https://github.com/googlecartographer/cartographer/pull/623

PR #618, opened 2017-10-30
Refactoring of LocalTrajectoryBuilder::AddAccumulatedRangeData

ojura:
In preparation for #574:

- introduce `InsertIntoSubmap` for 2D
- clarify some variable names in 3D
- move `rotational_scan_matcher_histogram` calculation
  to `InsertIntoSubmap` for 3D (so it's not performed unncecesarily,
   if the motion filter is going to discard the data)
- refactor last version of range data before insertion into
  `range_data_in_local` (`filtered_range_data_in_local` for 3D), so it can be used for `InsertIntoSubmap` and then moved into `MatchingResult` in #619


ojura:
@wohe Updated

PR #620, opened 2017-10-30
Remove PoseEstimate.

ojura:
Depends on #574 and switching `cartographer_ros` to use the new callback API for handling the last scan and pose estimate.


BrannonKing:
Can you explain to me about the purpose of this change? I'm still interested in scoring a pose estimate. I can see how to get the covariance out of the Ceres scan matcher. It seems that this change would make it easier to aggregate that covariance across the submaps.

ojura:
The new range data callback API is a substitute for `PoseEstimate`. Instead of polling the Cartographer library for updated pose estimates, you pass a callback which gets called whenever the pose estimate is updated; that happens after scan matching (and possibly inserting) new range data from the sensor queue.

SirVer:
As stated in OP, this does depend on the base PR being merged and the update to `cartographer_ros` - this is a beautiful change once it can land. For now it is not reviewable though, since the diff is mixed with the base branch and cartographer_ros is not updated.

Closing for the moment. @ojura, please just reopen or resend once we are at the point.

ojura:
Sure.  But still, can't you just review the last commit, `6fec906`? All the above commits are from #574.

ojura:
@SirVer Please reopen.

SirVer:
@ojura What is the point? Reviewing a moving target doesn't help you, since as long as the underlying code changes, your code is never final. It doesn't save any time. Let's review once we can merge.

PR #623, opened 2017-11-02
Extrapolator reuses result using additional IMU trackers.

gaschler:
Avoids extrapolating from the last known pose buts reuses an
intermediate result from a previous ExtrapolatePose call if
time is in the future. This makes the extrapolator stricter that time 
arguments must be monotonously increasing when calling the
Extrapolate methods.

### DIFF
--- a/cartographer/mapping/pose_extrapolator.cc
+++ b/cartographer/mapping/pose_extrapolator.cc
@@ -75,6 +75,8 @@ void PoseExtrapolator::AddPose(const common::Time time,
   AdvanceImuTracker(time, imu_tracker_.get());
   TrimImuData();
   TrimOdometryData();
+  odometry_imu_tracker_ = common::make_unique<ImuTracker>(*imu_tracker_);
+  pose_imu_tracker_ = common::make_unique<ImuTracker>(*imu_tracker_);
 }
 
 void PoseExtrapolator::AddImuData(const sensor::ImuData& imu_data) {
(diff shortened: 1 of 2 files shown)


cartographer/mapping/pose_extrapolator.cc
```diff
@@ -75,6 +75,8 @@ void PoseExtrapolator::AddPose(const common::Time time,
   AdvanceImuTracker(time, imu_tracker_.get());
   TrimImuData();
   TrimOdometryData();
+  odometry_imu_tracker_ = common::make_unique<ImuTracker>(*imu_tracker_);
+  pose_imu_tracker_ = common::make_unique<ImuTracker>(*imu_tracker_);
```
wohe: I don't understand the naming. `extrapolation_imu_tracker_` maybe?

gaschler: ExtrapolatePose uses pose_imu_tracker_, AddOdometryData uses odometry_imu_tracker_.
Both extrapolate starting from imu_tracker_, the last known pose.
Do you suggest to rename pose_imu_tracker_ to extrapolation_imu_tracker_? That's fine with me.

wohe: Yes, that was my suggestion. I find 'pose' not understandable. There are poses everywhere: some are added as input (this is the most obvious use of 'pose' related to the public interface of this class, but it has nothing to do with them), there are poses each IMU tracker provides, poses in odometry. That it is the IMU tracker used for the extrapolated pose does not seem obvious to me. This is why I suggested 'extrapolation'. That we extrapolate the rotation to be able to understand odometry seems like an implementation detail, but the `extrapolation_imu_tracker_` would be used for the extrapolated pose. Does that make sense? I'm open to better suggestions.

gaschler: Done

PR #627, opened 2017-11-03
Generalize IntegrateImu().

wohe:
This is in preparation of changing the data structure
for IMU data away from a deque. Needed for localization
and life-long mapping.


cartographer/mapping_3d/imu_integration.h
```diff
@@ -36,25 +36,19 @@ struct IntegrateImuResult {
...
 };
 
-// Returns velocity delta in map frame.
-IntegrateImuResult<double> IntegrateImu(
-    const std::deque<sensor::ImuData>& imu_data, const common::Time start_time,
-    const common::Time end_time,
-    std::deque<sensor::ImuData>::const_iterator* it);
-
-template <typename T>
+template <typename T, typename RangeType, typename IteratorType>
```
gaschler: I find
T -> NumberType
RangeType -> ContainerType
easier to read.

cartographer/mapping_3d/imu_integration.h
```diff
@@ -36,25 +36,19 @@ struct IntegrateImuResult {
...
 IntegrateImuResult<T> IntegrateImu(
-    const std::deque<sensor::ImuData>& imu_data,
+    const RangeType& imu_data,
     const Eigen::Transform<T, 3, Eigen::Affine>&
         linear_acceleration_calibration,
     const Eigen::Transform<T, 3, Eigen::Affine>& angular_velocity_calibration,
     const common::Time start_time, const common::Time end_time,
-    std::deque<sensor::ImuData>::const_iterator* it) {
+    IteratorType* const it) {
```
gaschler: Have you considered the alternative "ContainerType::const_iterator*? This would require less abstraction.

wohe: This change is to support ranges, e.g. `mapping::Range<>`, so `const_iterator` is not available.

cartographer/mapping_3d/imu_integration.h
wohe: Not feeling strongly but `T` is here to support Ceres jets, i.e. not just a number but including derivative information. This is consistent with `common/math.h` and `transform/transform.h`. I'm open to better suggestions.

`RangeType` is chosen because in the follow up this will not be a container.

PR #628, opened 2017-11-03
Adds a 'LandmarkData' struct and proto.

pifon2a:


cartographer/sensor/proto/sensor.proto
```diff
@@ -44,3 +44,15 @@ message FixedFramePoseData {
   optional int64 timestamp = 1;
   optional transform.proto.Rigid3d pose = 2;
 }
+
+// Proto representation of ::cartographer::sensor::LandmarkData.
+message LandmarkData {
+  message Landmark {
+    optional string id = 1;
```
wohe: Should we use `bytes` instead of `string` to not force UTF-8?

pifon2a: Done. Thanks!

PR #631, opened 2017-11-06
Introduce sensor::MapByTime<>.

wohe:
This data structure is used for keeping IMU data. This
allows trimming IMU data in the middle which is needed for
life-long mapping.


ojura:
This API is not very serialization friendly. How can I iterate through all trajectories, and through all per-trajectory data in order to serialize everything?

wohe:
Iterating through all per-trajectory data is already supported: `trajectory()`. For iterating over all trajectories please take a look e.g. at `MapById<>::trajectory_ids()`. Not sure what you mean by serialization friendly, but having these nested loops would not be much different from what we had before.

ojura:
`MapById<>::trajectory_ids()`

Okay, thanks. I was referring to the fact that we would have to look up the trajectory IDs externally. Perhaps `MapByTime` should also get `trajectory_ids()`?

wohe:
Of course. We just did not need it so far.

PR #632, opened 2017-11-06
Add 3D localization support.

wohe:
This adds information about which nodes are contained in which
deserialized submaps. This is necessary for the rotational scan
matcher.

Also skips frozen trajectories when tying together nodes of a
trajectory with IMU data which is unavailable in this case.

### DIFF
--- a/cartographer/mapping/map_builder.cc
+++ b/cartographer/mapping/map_builder.cc
@@ -229,6 +229,22 @@ void MapBuilder::LoadMap(io::ProtoStreamReader* const reader) {
       sparse_pose_graph_->AddSubmapFromProto(submap_pose, proto.submap());
     }
   }
+
+  // Add information about which nodes belong to which submap.
+  for (const proto::SparsePoseGraph::Constraint& constraint_proto :
+       pose_graph.constraint()) {
+    if (constraint_proto.tag() !=
+        mapping::proto::SparsePoseGraph::Constraint::INTRA_SUBMAP) {
+      continue;
+    }
+    const NodeId node_id{
+        trajectory_remapping.at(constraint_proto.node_id().trajectory_id()),
+        constraint_proto.node_id().node_index()};
+    const SubmapId submap_id{
+        trajectory_remapping.at(constraint_proto.submap_id().trajectory_id()),
+        constraint_proto.submap_id().submap_index()};
+    sparse_pose_graph_->AddNodeToSubmap(node_id, submap_id);
+  }
   CHECK(reader->eof());
 }
 
--- a/cartographer/mapping/sparse_pose_graph.h
+++ b/cartographer/mapping/sparse_pose_graph.h
@@ -88,7 +88,7 @@ class SparsePoseGraph {
                                 const proto::Node& node) = 0;
 
   // Adds information that 'node_id' was inserted into 'submap_id'. The submap
-  // and has to be deserialized first.
+  // has to be deserialized first.
   virtual void AddNodeToSubmap(const NodeId& node_id,
                                const SubmapId& submap_id) = 0;
 
--- a/cartographer/mapping/sparse_pose_graph.h
+++ b/cartographer/mapping/sparse_pose_graph.h
@@ -87,6 +87,11 @@ class SparsePoseGraph {
   virtual void AddNodeFromProto(const transform::Rigid3d& global_pose,
                                 const proto::Node& node) = 0;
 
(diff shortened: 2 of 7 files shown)


wohe:
Related to googlecartographer/cartographer_ros#518.

cartographer/mapping/sparse_pose_graph.h
```diff
@@ -87,6 +87,11 @@ class SparsePoseGraph {
   virtual void AddNodeFromProto(const transform::Rigid3d& global_pose,
                                 const proto::Node& node) = 0;
 
+  // Adds information that 'node_id' was inserted into 'submap_id'. The submap
+  // and has to be deserialized first.
```
ojura: Missed a word here.

wohe: Thanks. Fixed.

cartographer/mapping_2d/sparse_pose_graph.cc
```diff
@@ -427,6 +427,14 @@ void SparsePoseGraph::AddNodeFromProto(const transform::Rigid3d& global_pose,
   });
 }
 
+void SparsePoseGraph::AddNodeToSubmap(const mapping::NodeId& node_id,
+                                      const mapping::SubmapId& submap_id) {
+  common::MutexLocker locker(&mutex_);
+  AddWorkItem([this, node_id, submap_id]() REQUIRES(mutex_) {
+    submap_data_.at(submap_id).node_ids.insert(node_id);
```
ojura: Just to make you aware, this has already been [implemented in `AddSerializedConstraints`](https://github.com/googlecartographer/cartographer/commit/7964211fef53b13831a09baaa022fdce80f6c19e#diff-6d03d3eb04a3406b2be0eda40f99e843R454), although `AddSerializedConstraints` is not used yet anywhere.

It's planned that `AddSerializedConstraints` will be called from `LoadMap` only if `load_frozen = false`, as you have suggested earlier. However, you want to call `AddNodeToSubmap` unconditionally. Do you have thoughts on this?

cartographer/mapping/map_builder.cc
```diff
@@ -229,6 +229,22 @@ void MapBuilder::LoadMap(io::ProtoStreamReader* const reader) {
...
+      continue;
+    }
+    const NodeId node_id{
+        trajectory_remapping.at(constraint_proto.node_id().trajectory_id()),
+        constraint_proto.node_id().node_index()};
+    const SubmapId submap_id{
+        trajectory_remapping.at(constraint_proto.submap_id().trajectory_id()),
+        constraint_proto.submap_id().submap_index()};
+    sparse_pose_graph_->AddNodeToSubmap(node_id, submap_id);
+  }
```
ojura: Maybe you could [rewrite the SPG proto](https://github.com/googlecartographer/cartographer/pull/544/commits/8df868c9a29bc84a025efa181619256747bb0842#diff-e09271f09acc58d86ed1eb0e57a2b72cR191) instead because we'll soon be using it for full deserealization?

cartographer/mapping_2d/sparse_pose_graph.cc
wohe: This is best discussed in the context of the PR that makes use of `AddSerializedConstraints()`. I'm not too sure there is much to deduplicate though. For loading a frozen map (and not continuing SLAM) we do not want to add constraints, but the information which nodes are part of which submap is still needed.

cartographer/mapping_3d/sparse_pose_graph/optimization_problem.cc
```diff
@@ -216,6 +216,11 @@ void OptimizationProblem::Solve(const std::vector<Constraint>& constraints,
     for (auto node_it = node_data_.begin(); node_it != node_data_.end();) {
       const int trajectory_id = node_it->id.trajectory_id;
       const auto trajectory_end = node_data_.EndOfTrajectory(trajectory_id);
+      if (frozen_trajectories.count(trajectory_id) != 0) {
+        // We skip frozen trajectories.
+        node_it = trajectory_end;
+        continue;
```
cschuet: Why not 'break'?

wohe: We only want to skip this trajectory because it is frozen, but the next trajectory might need poses tied together by IMU again. In pure localization this is currently skipping trajectory 0 containing the map, continuing with trajectory 1 which is not frozen.

PR #633, opened 2017-11-06
Range loop over trajectories for MapByTime

ojura:
More or less copy pasted from `MapById`.

### DIFF
--- a/cartographer/sensor/map_by_time.h
+++ b/cartographer/sensor/map_by_time.h
@@ -140,7 +140,7 @@ class MapByTime {
     using iterator_category = std::bidirectional_iterator_tag;
     using value_type = int;
     using difference_type = int64;
-    using pointer = int*;
+    using pointer = const int*;
     using reference = const int&;
 
     explicit ConstTrajectoryIterator(
--- a/cartographer/sensor/map_by_time.h
+++ b/cartographer/sensor/map_by_time.h
@@ -34,9 +34,6 @@ namespace sensor {
 // 'DataType' must contain a 'time' member of type common::Time.
 template <typename DataType>
 class MapByTime {
- private:
-  using TimeIndexedMap = std::map<common::Time, DataType>;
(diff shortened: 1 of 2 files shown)


cartographer/sensor/map_by_time.h
```diff
@@ -34,6 +34,9 @@ namespace sensor {
 // 'DataType' must contain a 'time' member of type common::Time.
 template <typename DataType>
 class MapByTime {
+ private:
+  using TimeIndexedMap = std::map<common::Time, DataType>;
```
wohe: No need for this and no better readability? Remove?

cartographer/sensor/map_by_time.h
```diff
@@ -130,7 +132,44 @@ class MapByTime {
...
-    typename std::map<common::Time, DataType>::const_iterator iterator_;
+    typename TimeIndexedMap::const_iterator iterator_;
+  };
+
+  class ConstTrajectoryIterator {
+   public:
+    using iterator_category = std::bidirectional_iterator_tag;
+    using value_type = int;
+    using difference_type = int64;
+    using pointer = int*;
```
wohe: Shouldn't this be `const int*`?

cartographer/sensor/map_by_time.h
ojura: Okay. I thought that this might make sense since `std::map<common::Time, DataType>` is used 5 times.

cartographer/sensor/map_by_time.h
ojura: Do you want to make the change in `MapById` as well?

ojura: Assuming you do. Updated.

ojura: Wait, shouldn't it be `int * const`? A pointer to const, not a const pointer?

PR #635, opened 2017-11-07
Prepare for proto3 switch by requiring proto3.

SirVer:
Also change Dockerfiles to build proto3 from HEAD.

PAIR=@cschuet


SirVer:
Please ignore, playing around with mergebot.

PR #636, opened 2017-11-07
Unwarp by point in LocalTrajectoryBuilder.

gaschler:


gaschler:
This changes depends on https://github.com/googlecartographer/cartographer_ros/pull/576

gaschler:
I did some measurements and the computation time of 3D SLAM was ~40% longer, so do not merge the current implementation yet.

SirVer:
Oh right, I forgot. Should this PR then be closed till it is ready?

gaschler:
Closing because the current implementation is not efficient enough in 3D.

gaschler:
With some coarse filtering before unwarping in the 3D pipeline, this feature is now a better compromise between correct point cloud unwarping and computation time. In the worst case, computation time is ~20% longer for offline SLAM on the data of two high-density 3D laser scanners.

cschuet:
@wohe Do you feel the performance impact is acceptable? Should we go ahead with this feature?

cschuet:
@wohe How about disabling unwarping if all points in the batch have the same timestamp?

gaschler:
I implemented the TODO in ExtrapolatePose to cache the result there, which reduces the most common case (points without time deltas) to 100-105% computation time on frondend-heavy tasks (short offline mapping of high-density 3D data).
@cschuet Could you please take another look?

gaschler:
@wally-the-cartographer merge

wally-the-cartographer:

wohe:
@gaschler Should the comment at https://github.com/googlecartographer/cartographer/blob/0156e6b8ce3132786fb49c5a19cf0a9b95455842/cartographer/sensor/point_cloud.h#L36 be updated or do I misunderstand this change?

gaschler:
@wohe Thanks for pointing out this wrong comment, sent you a PR.

PR #637, opened 2017-11-07
move string and size specific int/uint typedefs into cartographer namespace

catskul:
move string and size specific int/uint typedefs into cartographer namespace as minimal change necessary to keep global namespace clean

Alternate, more complete pull request forthcoming.


SirVer:
Thanks!

catskul:
Any desire for a pull request of other option of replacing string  with std::string all over the place, or is this sufficient?

SirVer:
@catskul Removing this using and moving to std::string everywhere is the cleanest approach. A PR would be very welcome.

catskul:
will.

PR #606, opened 2017-10-20
Initial pose implementation.

cschuet:
Fixes #469.

PR #642, opened 2017-11-08
Add support for loading initial trajectory pose

damienrg:
#606 add initial trajectory pose to TrajectoryBuilderOptions proto but it was not possible to load it.

### DIFF
--- a/cartographer/mapping_3d/sparse_pose_graph.h
+++ b/cartographer/mapping_3d/sparse_pose_graph.h
@@ -32,7 +32,6 @@
 #include "cartographer/common/mutex.h"
 #include "cartographer/common/thread_pool.h"
 #include "cartographer/common/time.h"
-#include "cartographer/mapping/initial_trajectory_pose.h"
 #include "cartographer/mapping/pose_graph_trimmer.h"
 #include "cartographer/mapping/sparse_pose_graph.h"
 #include "cartographer/mapping/trajectory_connectivity_state.h"
@@ -105,9 +104,9 @@ class SparsePoseGraph : public mapping::SparsePoseGraph {
   mapping::MapById<mapping::NodeId, mapping::TrajectoryNode>
   GetTrajectoryNodes() override EXCLUDES(mutex_);
   std::vector<Constraint> constraints() override EXCLUDES(mutex_);
-  void SetInitialTrajectoryPose(int from_trajectory_id, int to_trajectory_id,
-                                const transform::Rigid3d& pose,
-                                const common::Time time) override
+  void SetInitialTrajectoryPose(
+      int from_trajectory_id,
+      const mapping::InitialTrajectoryPose& initial_pose) override
       EXCLUDES(mutex_);
   transform::Rigid3d GetInterpolatedGlobalTrajectoryPose(
       int trajectory_id, const common::Time time) const REQUIRES(mutex_);
--- a/cartographer/mapping_3d/sparse_pose_graph.h
+++ b/cartographer/mapping_3d/sparse_pose_graph.h
@@ -32,6 +32,7 @@
 #include "cartographer/common/mutex.h"
 #include "cartographer/common/thread_pool.h"
 #include "cartographer/common/time.h"
+#include "cartographer/mapping/initial_trajectory_pose.h"
 #include "cartographer/mapping/pose_graph_trimmer.h"
 #include "cartographer/mapping/sparse_pose_graph.h"
 #include "cartographer/mapping/trajectory_connectivity_state.h"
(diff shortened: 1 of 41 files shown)


damienrg:
Was Cartographer Open House recorded? If yes is it possible to download it as I think there are valid arguments for going in the command-line flag in cartographer_start_trajectory though I am not sure if it is the best way for now.

SirVer:
No, unfortunately we do not report the open houses. But @cschuet should upload the slides soonish. What other ideas do you have for providing the initial pose?

cschuet:
Slides are up, but there is nothing on there that reflects them.

How about you write a short RFC, Damien, where you explain how think it
should be done and then we can discuss on there?

cartographer/mapping_3d/sparse_pose_graph.h
```diff
@@ -104,9 +104,9 @@ class SparsePoseGraph : public mapping::SparsePoseGraph {
   mapping::MapById<mapping::NodeId, mapping::TrajectoryNode>
   GetTrajectoryNodes() override EXCLUDES(mutex_);
   std::vector<Constraint> constraints() override EXCLUDES(mutex_);
-  void SetInitialTrajectoryPose(int from_trajectory_id, int to_trajectory_id,
-                                const transform::Rigid3d& pose,
-                                const common::Time time) override
+  void SetInitialTrajectoryPose(
```
SirVer: I am not sure why this is required. Adding the initial pose through the TrajectoryBuilderOptions should be sufficient, right?

configuration_files/trajectory_builder.lua
```diff
@@ -12,11 +12,13 @@
...
 
+include "initial_trajectory_pose.lua"
 include "trajectory_builder_2d.lua"
 include "trajectory_builder_3d.lua"
 
 TRAJECTORY_BUILDER = {
   trajectory_builder_2d = TRAJECTORY_BUILDER_2D,
   trajectory_builder_3d = TRAJECTORY_BUILDER_3D,
   pure_localization = false,
+  initial_trajectory_pose = INITIAL_TRAJECTORY_POSE,
```
SirVer: I am not sure this is a good approach. While it is symmetric to how pure_localization is handled, this is quite different in the use case.

Pure localization always goes along with different settings, for example the loop closure will be run more often usually and maybe less aggressive settings for search windows. 

Initial pose however is very dependant on the data set in question: once you have one run finished, you want to start a second and relate the initial pose to this particular state. It sounds more like a flag than a configuration paramater to me. Therefore I feel that a commandline flag would be more appropriate for it.

Wdyt?

cartographer/mapping_3d/sparse_pose_graph.h
damienrg: TrajectoryBuilderOptions is not available in sparse_pose_graph. Maybe you mean that I should do this https://github.com/googlecartographer/cartographer/pull/606#discussion_r149653974.

configuration_files/trajectory_builder.lua
damienrg: Yes it is cumbersome to do it this way and as you mentioned it was to be consistent with pure localization. Can you elaborate a bit about command line flag? Is it a flag that should be passed to start_trajectory_main in cartographer_ros? If yes, what flag do you have in mind, something like:
```
-initial_pose "{ to_trajectory_id: 1, timestamp: now, translation : { 0, 0, 0}, rotation : { ... }}"
```

SirVer:
> Is it a flag that should be passed to start_trajectory_main in cartographer_ros?

Yes, I think so. That seems to be the correct place to me - after all that is the binary you need to run to start a new trajectory. And the initial pose is a one-time option, so commandline seems right. I just wonder about how it is provided.

> initial_pose "{ to_trajectory_id: 1, timestamp: now, translation : { 0, 0, 0}, rotation : { ... }}"

I think a good way would be to write this as a lua transform, like in [transform.lua](https://github.com/googlecartographer/cartographer_ros/blob/master/cartographer_ros/configuration_files/transform.lua). I.e. in the general case:

    start_trajectory_main --initial_pose '{ translation = { 0., 0., 0. }, rotation = { 0., -math.pi / 2., 0., } }' 

This can be parsed as `LuaParameterDictionary("return " + FLAGS_initial_pose`.

cschuet:
We also need `to_trajectory_id` and `timestamp` so this would look like e.g.

    start_trajectory_main --initial_pose '{to_trajectory_id = 0, timestamp = 1234, relative_pose = { translation = { 0., 0., 0. }, rotation = { 0., -math.pi / 2., 0., } } }'

damienrg:
Closing this and I will open a PR for cartographer_ros.

jihoonl:
@damienrg How's your progress? I would like to take over if you are stuck by something. I got a working version internally.

PR #648, opened 2017-11-10
CMake bug fixes and performance improvements.

### LINKED ISSUE #341: Dockerfiles install Ceres from master instead of a specific version

### LINKED ISSUE #629: FindLuaGoogle.cmake accepts Lua 5.1 though cartographer requires >= 5.2

SirVer:
- Pin versions of Ceres and protobuf (Fixes #341).
- Require Lua >= 5.2 (Fixes #629).
- No longer run ceres tests in CI. We implicitly test Ceres through Cartographer's tests and this saves ~4 minutes on CI.

### DIFF
--- a/scripts/install_ceres.sh
+++ b/scripts/install_ceres.sh
@@ -17,9 +17,12 @@
 set -o errexit
 set -o verbose
 
+VERSION="1.13.0"
+
 # Build and install Ceres.
 git clone https://ceres-solver.googlesource.com/ceres-solver
 cd ceres-solver
+git checkout tags/${VERSION}
 mkdir build
 cd build
 cmake .. -G Ninja
--- a/scripts/install_proto3.sh
+++ b/scripts/install_proto3.sh
@@ -17,9 +17,12 @@
 set -o errexit
 set -o verbose
 
+VERSION="v3.4.1"
+
 # Build and install proto3.
 git clone https://github.com/google/protobuf.git
 cd protobuf
+git checkout tags/${VERSION}
 mkdir build
 cd build
 cmake -G Ninja \
--- a/cmake/modules/FindLuaGoogle.cmake
+++ b/cmake/modules/FindLuaGoogle.cmake
@@ -70,7 +70,7 @@ unset(_lua_append_versions)
 
 # this is a function only to have all the variables inside go away automatically
 function(_lua_set_version_vars)
-    set(LUA_VERSIONS5 5.3 5.2 5.1 5.0)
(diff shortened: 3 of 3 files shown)


ojura:
> - No longer run ceres tests in CI. We implicitly test Ceres through Cartographer's tests and this saves ~4 minutes on CI.
Wait, where has this been changed in this PR?

scripts/install_ceres.sh
```diff
@@ -17,9 +17,12 @@
 set -o errexit
 set -o verbose
 
+VERSION="1.13.0"
+
 # Build and install Ceres.
 git clone https://ceres-solver.googlesource.com/ceres-solver
 cd ceres-solver
+git checkout tags/${VERSION}
```
damienrg: Clone directly the tag to save time if they are huge differences from master and tag?

damienrg: Clone directly the tag to save time if they are huge differences from master and tag?

scripts/install_proto3.sh
```diff
@@ -17,9 +17,12 @@
 set -o errexit
 set -o verbose
 
+VERSION="v3.4.1"
+
 # Build and install proto3.
 git clone https://github.com/google/protobuf.git
 cd protobuf
+git checkout tags/${VERSION}
```
damienrg: Same here.

scripts/install_ceres.sh
SirVer: The only way to do this I know is to use `git clone --branch` which also accepts tags. But if there is a tag and a branch with the same name, this will give us the wrong thing, so I had to use this. 

https://stackoverflow.com/questions/791959/download-a-specific-tag-with-git

SirVer:
@ojura Ups, thanks for catching this merge-o.

@wally-the-cartographer merge

wally-the-cartographer:

scripts/install_ceres.sh
damienrg: Yes I was thinking to `git clone --branch` and indeed it does not work in this case.

damienrg: Using git fetch you can do something like this:
```
mkdir ceres-solver
cd ceres-solver
git init
git remote add origin https://ceres-solver.googlesource.com/ceres-solver
git fetch --depth 1 origin "tags/${VERSION}"
git checkout -b "${VERSION}" FETCH_HEAD
```

SirVer: @damienrg With the short history of Cartographer, this barely saves space. I think this is not worth the extra lines for now.

PR #661, opened 2017-11-13
Omit odometry penalty within frozen trajectory.

gaschler:
In the case of pure localization, this reduces the size of the optimization problem by a large factor.


wohe:
@wally-the-cartographer merge

wally-the-cartographer:

PR #664, opened 2017-11-13
Check if filtered point clouds are empty.

zjwoody:
The point cloud will be filtered by MaxRangeFilter before doing scan matching. For some special cases, the filtered point cloud maybe empty, it will cause crash:
real_time_correlative_scan_matcher.cc:159 Check failed: candidate.score > 0.f (nan vs. 0) 
This PR fixed this bug, check the point cloud size before scan matching, return false for empty point cloud.

### DIFF
--- a/cartographer/mapping_2d/local_trajectory_builder.cc
+++ b/cartographer/mapping_2d/local_trajectory_builder.cc
@@ -48,7 +48,7 @@ sensor::RangeData LocalTrajectoryBuilder::TransformAndFilterRangeData(
       sensor::VoxelFiltered(cropped.misses, options_.voxel_filter_size())};
 }
 
-void LocalTrajectoryBuilder::ScanMatch(
+bool LocalTrajectoryBuilder::ScanMatch(
     const common::Time time, const transform::Rigid2d& pose_prediction,
     const sensor::RangeData& gravity_aligned_range_data,
     transform::Rigid2d* const pose_observation) {
@@ -61,6 +61,10 @@ void LocalTrajectoryBuilder::ScanMatch(
       options_.adaptive_voxel_filter_options());
   const sensor::PointCloud filtered_gravity_aligned_point_cloud =
       adaptive_voxel_filter.Filter(gravity_aligned_range_data.returns);
+  if (filtered_gravity_aligned_point_cloud.empty()) {
+    LOG(WARNING) << "The point cloud is empty after filtering.";
+    return false;
+  }
   if (options_.use_online_correlative_scan_matching()) {
     real_time_correlative_scan_matcher_.Match(
         pose_prediction, filtered_gravity_aligned_point_cloud,
@@ -71,6 +75,7 @@ void LocalTrajectoryBuilder::ScanMatch(
   ceres_scan_matcher_.Match(
       pose_prediction, initial_ceres_pose, filtered_gravity_aligned_point_cloud,
       matching_submap->probability_grid(), pose_observation, &summary);
+  return true;
 }
 
 std::unique_ptr<LocalTrajectoryBuilder::MatchingResult>
@@ -150,8 +155,11 @@ LocalTrajectoryBuilder::AddAccumulatedRangeData(
 
   // local map frame <- gravity-aligned frame
   transform::Rigid2d pose_estimate_2d;
-  ScanMatch(time, pose_prediction, gravity_aligned_range_data,
-            &pose_estimate_2d);
+  if (!ScanMatch(time, pose_prediction, gravity_aligned_range_data,
+                 &pose_estimate_2d)) {
+    LOG(WARNING) << "Scan matching failed.";
+    return nullptr;
+  }
   const transform::Rigid3d pose_estimate =
       transform::Embed3D(pose_estimate_2d) * gravity_alignment;
   extrapolator_->AddPose(time, pose_estimate);
--- a/cartographer/mapping_2d/local_trajectory_builder.cc
+++ b/cartographer/mapping_2d/local_trajectory_builder.cc
@@ -69,8 +69,7 @@ std::unique_ptr<transform::Rigid2d> LocalTrajectoryBuilder::ScanMatch(
         matching_submap->probability_grid(), &initial_ceres_pose);
   }
 
-  std::unique_ptr<transform::Rigid2d> pose_observation =
-      common::make_unique<transform::Rigid2d>();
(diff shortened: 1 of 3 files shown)


SirVer:
I am confused a bit... How can voxel filtering make a non-empty point cloud empty? Should not at least one point remain.

gaschler:
LocalTrajectoryBuilder::ScanMatch uses AdaptiveVoxelFilter to filter points further than options_.max_range.
If we want to return failure, I'd rather return a unique_ptr than bool for consistency.

cartographer/mapping_2d/local_trajectory_builder.cc
```diff
@@ -48,7 +48,7 @@ sensor::RangeData LocalTrajectoryBuilder::TransformAndFilterRangeData(
       sensor::VoxelFiltered(cropped.misses, options_.voxel_filter_size())};
 }
 
-void LocalTrajectoryBuilder::ScanMatch(
+bool LocalTrajectoryBuilder::ScanMatch(
```
wohe: As @gaschler pointed out, changing the function to return a `std::unique_ptr<transform::Rigid2d>` would be better.

cartographer/mapping_2d/local_trajectory_builder.cc
```diff
@@ -61,6 +61,10 @@ void LocalTrajectoryBuilder::ScanMatch(
       options_.adaptive_voxel_filter_options());
   const sensor::PointCloud filtered_gravity_aligned_point_cloud =
       adaptive_voxel_filter.Filter(gravity_aligned_range_data.returns);
+  if (filtered_gravity_aligned_point_cloud.empty()) {
+    LOG(WARNING) << "The point cloud is empty after filtering.";
```
gaschler: No warning here.

cartographer/mapping_2d/local_trajectory_builder.cc
```diff
@@ -61,6 +61,10 @@ void LocalTrajectoryBuilder::ScanMatch(
...
+    return false;
```
gaschler: return nullptr

cartographer/mapping_2d/local_trajectory_builder.cc
```diff
@@ -61,6 +61,10 @@ void LocalTrajectoryBuilder::ScanMatch(
       options_.adaptive_voxel_filter_options());
   const sensor::PointCloud filtered_gravity_aligned_point_cloud =
       adaptive_voxel_filter.Filter(gravity_aligned_range_data.returns);
+  if (filtered_gravity_aligned_point_cloud.empty()) {
```
wohe: Doesn't the same bug also exist in 3D?

cartographer/mapping_2d/local_trajectory_builder.cc
gaschler: Please return `unique_ptr<transform::Rigid2d>`, it is more consistent with FastCorrelativeScanMatcher.

cartographer/mapping_2d/local_trajectory_builder.h
```diff
@@ -84,8 +84,9 @@ class LocalTrajectoryBuilder {
       const Eigen::Quaterniond& gravity_alignment);
 
   // Scan matches 'gravity_aligned_range_data' and fill in the
-  // 'pose_observation' with the result.
-  void ScanMatch(common::Time time, const transform::Rigid2d& pose_prediction,
+  // 'pose_observation' with the result. Return false if failed to match the
```
gaschler: "and returns observed pose, or nullptr on failure."

cartographer/mapping_2d/local_trajectory_builder.cc
zjwoody: Yes, I'll fix it in 3D. Thanks.

cartographer/mapping_2d/local_trajectory_builder.h
zjwoody: Done.

cartographer/mapping_2d/local_trajectory_builder.cc
zjwoody: Done.

cartographer/mapping_2d/local_trajectory_builder.cc
zjwoody: Done.

cartographer/mapping_2d/local_trajectory_builder.cc
zjwoody: Done.

zjwoody: Done.

cartographer/mapping_2d/local_trajectory_builder.h
```diff
@@ -84,10 +84,11 @@ class LocalTrajectoryBuilder {
       const Eigen::Quaterniond& gravity_alignment);
 
   // Scan matches 'gravity_aligned_range_data' and fill in the
-  // 'pose_observation' with the result.
-  void ScanMatch(common::Time time, const transform::Rigid2d& pose_prediction,
-                 const sensor::RangeData& gravity_aligned_range_data,
-                 transform::Rigid2d* pose_observation);
+  // 'pose_observation' with the result. Return the observed pose, or nullptr on
```
gaschler: Replace "and fill in ... with the result." by "and returns"

cartographer/mapping_2d/local_trajectory_builder.cc
```diff
@@ -61,16 +60,22 @@ void LocalTrajectoryBuilder::ScanMatch(
...
+    return nullptr;
+  }
   if (options_.use_online_correlative_scan_matching()) {
     real_time_correlative_scan_matcher_.Match(
         pose_prediction, filtered_gravity_aligned_point_cloud,
         matching_submap->probability_grid(), &initial_ceres_pose);
   }
 
+  std::unique_ptr<transform::Rigid2d> pose_observation =
```
gaschler: You could use auto here if you like.

zjwoody: Done.

cartographer/mapping_2d/local_trajectory_builder.h
zjwoody: Done.

gaschler:
@wally-the-cartographer merge

wally-the-cartographer:

PR #666, opened 2017-11-14
Serialize odometry data

ojura:
Replaces #550.

### DIFF
--- a/cartographer/mapping/map_builder.cc
+++ b/cartographer/mapping/map_builder.cc
@@ -197,15 +197,18 @@ void MapBuilder::SerializeState(io::ProtoStreamWriter* const writer) {
   {
     const auto all_odometry_data = sparse_pose_graph_->GetOdometryData();
     for (const int trajectory_id : all_odometry_data.trajectory_ids()) {
-      for (const auto& odometry_data : all_odometry_data.trajectory(trajectory_id)) {
+      for (const auto& odometry_data :
+           all_odometry_data.trajectory(trajectory_id)) {
         proto::SerializedData proto;
         auto* const odometry_data_proto = proto.mutable_odometry_data();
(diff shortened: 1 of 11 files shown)


ojura:
By the way, this PR is almost line-for-line identical to #548, `/s/IMU/Odometry`. And a similar one will be  forthcoming for fixed frame pose data in 3D. Perhaps this could be somehow simplified? Introducing all these getters for the optimization problem and the pose graph is noise, IMHO.

It would be nice if we had all these `MapByTime` objects in a single structure in the optimization problem, and if we could just ask this structure to serialize itself from `MapBuilder`.

cartographer/mapping/map_builder.cc
```diff
@@ -193,7 +193,19 @@ void MapBuilder::SerializeState(io::ProtoStreamWriter* const writer) {
       }
     }
   }
-  // TODO(whess): Serialize additional sensor data: odometry.
```
wohe: Retain the `TODO`? There is still fixed frame pose data missing.

cartographer/mapping/proto/serialization.proto
```diff
@@ -37,9 +37,14 @@ message ImuData {
...
+message OdometryData {
+  int32 trajectory_id = 1;
+  sensor.proto.OdometryData odometry_data = 2;
+}
+
 message SerializedData {
   Submap submap = 1;
   Node node = 2;
   ImuData imu_data = 3;
-  // TODO(whess): Add odometry.
```
wohe: Similar here: retain for fixed frame pose data?

ojura:
Updated. I also forgot to expose `AddOdometryData` in the SPG base class in `mapping/sparse_pose_graph.h`, so I added it.

wohe:
@wally-the-cartographer merge

wally-the-cartographer:

PR #668, opened 2017-11-14
Reduce transforms in LocalTrajectoryBuilder.

gaschler:
This speeds up AddRangeData because we avoid one copy and loop over all points.


cschuet:
@wally-the-cartographer merge

wally-the-cartographer:

PR #670, opened 2017-11-14
Remove PoseEstimate.

ojura:
Replaces #620.

Depends on switching `cartographer_ros` to use the new callback API for handling the last scan and pose estimate (googlecartographer/cartographer_ros#594).


ojura:
@wohe yanking all of this code sure does feel good, but I wonder why we haven't considered packing it all in a struct (`PoseEstimate`?) and passing it to the callback, instead of having a bunch of arguments in the callback like we do now.

@cshuet mentioned that he would like to pass the whole `MatchingResult`, so there will definitely be more discussion about how the callback looks.

wohe:
Seems unrelated to this PR. @ojura, if you resolve the conflicts, I'm happy to review this.

Regarding the struct for passing the information: I fail to see the benefit that would provide. We were using a struct here since `pose_estimate()` can only have a single return value. In the callback we can have multiple function arguments, so there is no need for a struct.

ojura:
It was kind of relevant because I was thinking that we could reuse the `PoseEstimate` struct for that.

One benefit of using a struct would be more stable and robust API. If we decide to pass something else via the callback, the library users wouldn't have to change their code if we were using a struct. Otherwise, they need to change the callback prototype in their code to follow the API change. 

Also, suppose that we change the interpretation of data, but not the type: e.g. the frame of range data; if we used a struct, we would rename the struct member (from `range_data_in_local` to `range_data_in_tracking`), and the library users would be alerted to this since the name changed and their code wouldn't compile. Currently, the callback arguments must only match by type and not by name, so this type of error wouldn't be caught during compiling.

wohe:
@wally-the-cartographer merge

wally-the-cartographer:

ojura:
@wohe would love to hear your thoughts (in case you missed the message above).

wohe:
As I said I would always use arguments as the default. Sometimes there is a good reason for a struct, one of them you mentioned. The situation here is the same as for other functions. You are right that you have to be careful when refactoring since argument names are not part of the interface, but this is true for every function that is part of the public API.

MichaelGrupp:
I noticed that `pose_estimate.h` still exists as an empty file after this commit, is this intentional?

See [here](https://github.com/googlecartographer/cartographer/blob/31f28b509746a1ea9dec417187acc26e7ca5f5ff/cartographer/mapping/pose_estimate.h)

gaschler:
@MichaelGrupp Good catch! The follow-up PR to remove it seems to be missing, feel free to send one.

MichaelGrupp:
Done - see the new PR.

PR #673, opened 2017-11-15
replace implicit use of cartographer::string with explicit use of std::string

### LINKED ISSUE #622: Remove using std::string and switch to using std::string directly everywhere.
https://github.com/googlecartographer/cartographer/blob/e10650910e157bd68c55f4d1bc6603f25f89d293/cartographer/common/port.h#L37

Should probably not be in the header file, or at least should be within one of the owned namespaces.

Fixing may cause knock on build failures for any include graph descendant.

catskul:
Fixes #622.


SirVer:
@catskul Wow, that was a ton of work! Thanks for doing this cleanup. I found one more naked string and fixed it.

@wally-the-cartographer merge

wally-the-cartographer:

catskul:
@SirVer very welcome.

PR #680, opened 2017-11-15
Rename proto::SparsePoseGraph.

wohe:
[RFC=0001](https://github.com/googlecartographer/rfcs/blob/master/text/0001-renaming-sparse-pose-graph.md)

### DIFF
--- a/cartographer/mapping/pose_graph.h
+++ b/cartographer/mapping/pose_graph.h
@@ -26,8 +26,8 @@
 #include "cartographer/common/lua_parameter_dictionary.h"
 #include "cartographer/mapping/id.h"
 #include "cartographer/mapping/pose_graph_trimmer.h"
+#include "cartographer/mapping/proto/pose_graph.pb.h"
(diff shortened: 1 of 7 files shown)


cartographer/mapping/pose_graph.h
```diff
@@ -26,8 +26,8 @@
 #include "cartographer/common/lua_parameter_dictionary.h"
 #include "cartographer/mapping/id.h"
 #include "cartographer/mapping/pose_graph_trimmer.h"
+#include "cartographer/mapping/proto/pose_graph.pb.h"
```
gaschler: This change will also be necessary in cartographer_ros/cartographer_ros/assets_writer_main.cc and perhaps others on the ros side.

wohe: Of course. But it makes no sense to send the follow up PR until this is in. And it only needs changes to `assets_writer_main.cc`, no others.

wohe:
@wally-the-cartographer merge

wally-the-cartographer:

PR #681, opened 2017-11-15
Filter early in AddRangeData.

gaschler:
Filter earlier in 3D.
Combine two transforms to one in 2D.

### DIFF
--- a/cartographer/mapping_2d/local_trajectory_builder.cc
+++ b/cartographer/mapping_2d/local_trajectory_builder.cc
@@ -37,11 +37,11 @@ LocalTrajectoryBuilder::LocalTrajectoryBuilder(
 LocalTrajectoryBuilder::~LocalTrajectoryBuilder() {}
 
 sensor::RangeData LocalTrajectoryBuilder::TransformAndFilterRangeData(
-    const transform::Rigid3f& gravity_alignment,
+    const transform::Rigid3f& transform,
     const sensor::RangeData& range_data) const {
-  const sensor::RangeData cropped = sensor::CropRangeData(
-      sensor::TransformRangeData(range_data, gravity_alignment),
-      options_.min_z(), options_.max_z());
+  const sensor::RangeData cropped =
+      sensor::CropRangeData(sensor::TransformRangeData(range_data, transform),
+                            options_.min_z(), options_.max_z());
   return sensor::RangeData{
       cropped.origin,
       sensor::VoxelFiltered(cropped.returns, options_.voxel_filter_size()),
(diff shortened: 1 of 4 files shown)


cartographer/mapping_2d/local_trajectory_builder.cc
```diff
@@ -37,11 +37,11 @@ LocalTrajectoryBuilder::LocalTrajectoryBuilder(
 LocalTrajectoryBuilder::~LocalTrajectoryBuilder() {}
 
 sensor::RangeData LocalTrajectoryBuilder::TransformAndFilterRangeData(
-    const transform::Rigid3f& gravity_alignment,
+    const transform::Rigid3f& transform,
```
wohe: Choose better names? This is not an arbitrary transform but needed so that min and max z make sense. `gravity_alignment` was also used in other places as well, so that helped the reader to understand that it is this transform. Please also see https://google-cartographer.readthedocs.io/en/latest/terminology.html for the frame it related to.

cartographer/mapping_2d/local_trajectory_builder.cc
```diff
@@ -121,26 +121,29 @@ LocalTrajectoryBuilder::AddRangeData(const common::Time time,
 
   if (num_accumulated_ >= options_.scans_per_accumulation()) {
     num_accumulated_ = 0;
-    return AddAccumulatedRangeData(
-        time, sensor::TransformRangeData(accumulated_range_data_,
-                                         tracking_delta.inverse()));
+    // Transforms 'accumulated_range_data_' into a frame where gravity direction
```
wohe: Since we now have `terminology.rst`: Prefer to make the code self describing? How about renaming `TransformAndFilterRangeData()` to reflect that it is into the gravity aligned frame? `TransformToGravityAlignedFrameAndFilter()`?

cartographer/mapping_2d/local_trajectory_builder.cc
```diff
@@ -121,26 +121,29 @@ LocalTrajectoryBuilder::AddRangeData(const common::Time time,
...
+    // is approximately +z.
+    const transform::Rigid3d gravity_alignment = transform::Rigid3d::Rotation(
+        extrapolator_->EstimateGravityOrientation(time));
+    const sensor::RangeData gravity_aligned_range_data =
+        TransformAndFilterRangeData(
+            gravity_alignment.cast<float>() * tracking_delta.inverse(),
+            accumulated_range_data_);
+    if (gravity_aligned_range_data.returns.empty()) {
```
wohe: Move this check back to `AddAccumulatedRangeData()` and you can get rid of the temporary. What do you think?

gaschler:
Thanks. Done with the changes, please take another look.

cartographer/mapping_2d/local_trajectory_builder.cc
```diff
@@ -121,26 +122,29 @@ LocalTrajectoryBuilder::AddRangeData(const common::Time time,
 
   if (num_accumulated_ >= options_.scans_per_accumulation()) {
     num_accumulated_ = 0;
+    const transform::Rigid3d gravity_alignment = transform::Rigid3d::Rotation(
+        extrapolator_->EstimateGravityOrientation(time));
     return AddAccumulatedRangeData(
-        time, sensor::TransformRangeData(accumulated_range_data_,
-                                         tracking_delta.inverse()));
+        time,
+        // Transforms 'accumulated_range_data_' into a frame where gravity
```
wohe: Remove the comment? The function name now already better reflects this and the comment misses the part that the call also filters the range data.

cartographer/mapping_2d/local_trajectory_builder.cc
```diff
@@ -36,7 +36,8 @@ LocalTrajectoryBuilder::LocalTrajectoryBuilder(
 
 LocalTrajectoryBuilder::~LocalTrajectoryBuilder() {}
 
-sensor::RangeData LocalTrajectoryBuilder::TransformAndFilterRangeData(
+sensor::RangeData
+LocalTrajectoryBuilder::TransformToGravityAlignedFrameAndFilter(
     const transform::Rigid3f& gravity_alignment,
```
wohe: I wonder whether we should not rename this argument still. It is a transform from the tracking frame in which the range data was accumulated and to gravity aligned frame at a different time. Not sure what a good name is though. Ideas?

cartographer/mapping_2d/local_trajectory_builder.cc
```diff
@@ -121,26 +122,29 @@ LocalTrajectoryBuilder::AddRangeData(const common::Time time,
...
-      extrapolator_->EstimateGravityOrientation(time));
-  const sensor::RangeData gravity_aligned_range_data =
-      TransformAndFilterRangeData(gravity_alignment.cast<float>(), range_data);
+    const common::Time time,
+    const sensor::RangeData& gravity_aligned_range_data) {
   if (gravity_aligned_range_data.returns.empty()) {
     LOG(WARNING) << "Dropped empty horizontal range data.";
     return nullptr;
   }
+  const transform::Rigid3d gravity_alignment = transform::Rigid3d::Rotation(
```
wohe: Can we pass this as an argument? It is closely tied to the `time` and `gravity_aligned_range_data` we take as arguments.

cartographer/mapping_2d/local_trajectory_builder.cc
gaschler: The Rigid3f argument transforms from first_pose_estimate_ to the gravity-aligned frame. Let's call it first_tracking_frame_to_gravity_aligned_frame. What do you think?

wohe: I would prefer a shorter name especially since your suggested name while very long does not convey the most important aspect: As pointed out in the terminology documentation there are separate gravity-aligned frames for different nodes. The one we use here is related to the tracking frame at a later time than the first tracking frame. Maybe `transform_to_gravity_aligned_frame`?

gaschler:
Done, PTAL.

wohe:
@wally-the-cartographer merge

wally-the-cartographer:

PR #684, opened 2017-11-16
Refactor image related code.

SirVer:
- Take immutable parameter as const ref instead of pointer.
- make kCairoFormat a global constant, since it is defined in many places in cartographer and cartographer_ros, but we rely on it actually always being the same value.


SirVer:
Followup in cartographer_ros will be send once this is in.

SirVer:
@wally-the-cartographer merge

wally-the-cartographer:

PR #686, opened 2017-11-16
Rename options for consistency.

wohe:
"scan" is better named "node" if it refers to a node in global SLAM.

"penalty factor" is renamed "weight" for consistency with other similar options.

"scans_per_accumulation" is renamed "num_accumulated_range_data" to match the code and since the accumulated data is not called "scan".


wohe:
@wally-the-cartographer merge

wally-the-cartographer:

PR #689, opened 2017-11-17
Serialize fixed frame pose data.

ojura:
Similar to #666 and #548.


cartographer/mapping_2d/pose_graph.h
```diff
@@ -83,7 +83,7 @@ class PoseGraph : public mapping::PoseGraph {
       EXCLUDES(mutex_);
   void AddFixedFramePoseData(
       int trajectory_id,
-      const sensor::FixedFramePoseData& fixed_frame_pose_data);
+      const sensor::FixedFramePoseData& fixed_frame_pose_data) override;
```
wohe: Shouldn't this be annotated `EXLUDES(mutex_)` here as well?

ojura: Currently it's just a `LOG(FATAL)`, no mutex locking.

cartographer/mapping_2d/pose_graph.h
```diff
@@ -109,8 +109,8 @@ class PoseGraph : public mapping::PoseGraph {
   sensor::MapByTime<sensor::ImuData> GetImuData() override EXCLUDES(mutex_);
   sensor::MapByTime<sensor::OdometryData> GetOdometryData() override
       EXCLUDES(mutex_);
-  sensor::MapByTime<sensor::FixedFramePoseData> GetFixedFramePoseData() override
-      EXCLUDES(mutex_);
+  sensor::MapByTime<sensor::FixedFramePoseData> GetFixedFramePoseData()
+      override;
```
ojura: I also removed here, since it's not implemented yet.

cartographer/mapping_2d/pose_graph.h
wohe: I understand that it is not implemented, but `EXCLUDES(mutex_)` is part of the interface, i.e. you annotate that whoever calls this function must not have locked `mutex_`. It does not claim that it would be locked in the function. There is not reason to change the interface just because the function is currently not implemented.

ojura:
@wohe I updated the interface with mutex annotations as suggested.

ojura:
@wohe ping. When this is merged, I can finally open the PR for loading unfrozen trajectories.

cartographer/mapping_2d/pose_graph.cc
```diff
@@ -555,6 +555,11 @@ sensor::MapByTime<sensor::OdometryData> PoseGraph::GetOdometryData() {
   return optimization_problem_.odometry_data();
 }
 
+sensor::MapByTime<sensor::FixedFramePoseData>
+PoseGraph::GetFixedFramePoseData() {
+  return {};  // Not implemented yet in 2D.
```
gaschler: Well, this is not ideal because it does not warn anybody, however, any warning would break the serialization function.
I guess 2D and 3D do behave differently, and this is a compromise so we can easily serialize both.

cschuet:
@wally-the-cartographer merge

wally-the-cartographer:

cartographer/mapping_2d/pose_graph.cc
ojura: Technically, it's not wrong - 2D _really has no FixedFramePoses_, so it returns an empty vector :-)

PR #693, opened 2017-11-22
ImuTracker uses conjugate.

gaschler:
Normalized quaternions can be inverted with conjugate, which is faster.
The motivation for optimizing ImuTracker::Advance is that
per-point unwarping will call this for every point.


gaschler:
@wally-the-cartographer merge

wally-the-cartographer:

PR #694, opened 2017-11-22
ExtrapolatePose uses fewer transforms.

gaschler:
ExtrapolatePose avoids conversion to Rigid3d and computes translation
and rotation directly, which is faster.
Per-point unwarping will call this function for every point, so we
optimize it.


gaschler:
@wally-the-cartographer merge

wally-the-cartographer:

PR #697, opened 2017-11-23
Fix debug build

ojura:
Similar to #418.


gaschler:
Is this change enough fix the GCC debug build?
A quick grep shows 8 places where `LOG(FATAL);` is used.
(In case you are about the change all 8 places, I'd rather not add a comment everywhere.)

ojura:
No, it's not required everywhere. See discussion in #418 (only if nontrivial destructor runs after `LOG(FATAL`).

gaschler:
Thanks, this will also fix my local valgrind runs.

@wally-the-cartographer merge

wally-the-cartographer:
Merge requested by authorized user gaschler. Merge queue now has a length of 3.

PR #702, opened 2017-11-23
CeresScanMatcher target_translation

gaschler:
For its translation cost function, CeresScanMatcher now takes
a target translation as an argument instead of a pose, which 
was confusing.


cartographer/mapping_3d/local_trajectory_builder.cc
```diff
@@ -158,7 +158,7 @@ LocalTrajectoryBuilder::AddAccumulatedRangeData(
     return nullptr;
   }
   ceres_scan_matcher_->Match(
-      matching_submap->local_pose().inverse() * pose_prediction,
+      (matching_submap->local_pose().inverse() * pose_prediction).translation(),
```
cschuet: Can we write this more efficiently now?

gaschler: Done

gaschler: Actually, I reverted the more efficient computation because it was less readable.

gaschler:
@wally-the-cartographer merge

wally-the-cartographer:

PR #707, opened 2017-11-24
Make the MotionFilter internal.

wohe:
Also moves it from mapping_3d to mapping since it is used
in both 2D and 3D SLAM.

[RFC=0003](https://github.com/googlecartographer/rfcs/blob/master/text/0003-internal-headers.md)


wohe:
@wally-the-cartographer merge

wally-the-cartographer:

PR #713, opened 2017-11-28
Implement end-to-end client streaming RPC.

cschuet:


cartographer_grpc/framework/rpc.cc
```diff
@@ -53,23 +74,53 @@ ::grpc::internal::ServerAsyncStreamingInterface* Rpc::streaming_interface() {
...
+  switch (rpc_handler_info_.rpc_type) {
+    case ::grpc::internal::RpcMethod::CLIENT_STREAMING:
+      return server_async_reader_.get();
+    default:
+      LOG(FATAL) << "RPC type not implemented.";
+  }
+  LOG(FATAL) << "Never reached.";
+}
+
+Rpc::RpcState* Rpc::SetRpcStatePending(State state, bool pending) {
```
gaschler: Unusual to have a Set* method return a state. Perhaps there should be a separate get method.

cartographer_grpc/framework/rpc.h
```diff
@@ -39,17 +40,27 @@ class Rpc {
     const State state;
     Service* service;
     Rpc* rpc;
+    bool pending;
```
gaschler: a comment would be good

cartographer_grpc/framework/service.cc
```diff
@@ -54,13 +54,16 @@ void Service::StartServing(
 void Service::StopServing() { shutting_down_ = true; }
 
 void Service::HandleEvent(Rpc::State state, Rpc* rpc, bool ok) {
+  rpc->SetRpcStatePending(state, false);
```
gaschler: Here, the state appears to be more like an event type. Rename?

cartographer_grpc/framework/service.cc
```diff
@@ -91,24 +101,56 @@ void Service::HandleNewConnection(Rpc* rpc, bool ok) {
...
   rpc->server_context()->AsyncNotifyWhenDone(
-      rpc->GetRpcState(Rpc::State::DONE));
+      rpc->SetRpcStatePending(Rpc::State::DONE, true));
   switch (rpc->rpc_type()) {
     case ::grpc::internal::RpcMethod::CLIENT_STREAMING:
-      RequestAsyncClientStreaming(method_index, rpc->server_context(),
-                                  rpc->streaming_interface(), completion_queue,
-                                  completion_queue,
-                                  rpc->GetRpcState(Rpc::State::NEW_CONNECTION));
+      RequestAsyncClientStreaming(
```
gaschler: Service needs several of rpc's attributes. Consider alternative designs such as making Rpc a small struct without logic and doing the work in Service.

cartographer_grpc/framework/rpc.h
```diff
@@ -39,17 +40,27 @@ class Rpc {
...
   };
 
   Rpc(int method_index, ::grpc::ServerCompletionQueue* server_completion_queue,
       const RpcHandlerInfo& rpc_handler_info, Service* service);
 
+  void OnRequest();
```
gaschler: The public interface of Rpc seems larger than necessary. Perhaps some of this should happen in Service?

cartographer_grpc/framework/rpc.h
```diff
@@ -39,17 +40,27 @@ class Rpc {
...
   int method_index() const { return method_index_; }
   ::grpc::ServerCompletionQueue* server_completion_queue();
   ::grpc::internal::RpcMethod::RpcType rpc_type() const;
   ::grpc::ServerContext* server_context() { return &server_context_; }
   ::grpc::internal::ServerAsyncStreamingInterface* streaming_interface();
-  RpcState* GetRpcState(State state);
+  ::grpc::internal::AsyncReaderInterface<::google::protobuf::Message>*
+  async_reader_interface();
+
+  RpcState* SetRpcStatePending(State state, bool pending);
```
gaschler: Rpc is already the name of the class so it can be removed from the method name, also in the next line.

cartographer_grpc/framework/server_test.cc
```diff
@@ -28,17 +28,67 @@ namespace framework {
...
+    request.set_input(i);
+    CHECK(writer->Write(request));
+  }
+  writer->WritesDone();
+  grpc::Status status = writer->Finish();
+  CHECK(status.ok());
+  CHECK_EQ(result.output(), 3);
+
+  server_->Shutdown();
 }
```
gaschler: maybe add a todo for tests that interrupt the connection or shutdown early?

cartographer_grpc/framework/rpc_handler.h
```diff
@@ -41,6 +45,9 @@ struct RpcHandlerInfo {
   const grpc::internal::RpcMethod::RpcType rpc_type;
 };
 
+void SendInternal(std::unique_ptr<::google::protobuf::Message> message,
```
gaschler: Can you explain the benefit of this indirection?

cschuet:
PTAL

cartographer_grpc/framework/rpc.h
```diff
@@ -37,32 +38,35 @@ class Rpc {
   enum class State { NEW_CONNECTION = 0, READ, WRITE, DONE };
   struct RpcState {
     const State state;
-    Service* service;
     Rpc* rpc;
+    bool pending;
```
gaschler: Add comment to explain

cartographer_grpc/framework/server.cc
```diff
@@ -66,7 +66,8 @@ void Server::RunCompletionQueue(
   void* tag;
   while (completion_queue->Next(&tag, &ok)) {
     auto* rpc_state = static_cast<Rpc::RpcState*>(tag);
```
gaschler: Again, shouldn't RpcState be renamed to RpcEvent?

cartographer_grpc/framework/server_test.cc
```diff
@@ -28,17 +28,65 @@ namespace framework {
...
+};
+
+const std::string kServerAddress = "localhost:50051";
+const std::size_t kNumThreads = 1;
+
+class ServerTest : public ::testing::Test {
+ protected:
+  void SetUp() override {
+    Server::Builder server_builder;
+    server_builder.SetServerAddress(kServerAddress);
```
gaschler: I think these tests become flaky if run in parallel because of the shared port.
If this is the case, add a todo.

cartographer_grpc/framework/service.cc
```diff
@@ -80,33 +84,42 @@ void Service::HandleNewConnection(Rpc* rpc, bool ok) {
...
-                              rpc->server_completion_queue());
+void Service::HandleRead(Rpc* rpc, bool ok) {
+  if (ok) {
+    rpc->OnRequest();
+    rpc->RequestStreamingReadIfNeeded();
+    return;
+  }
+
+  // Reads completed.
+  rpc->OnReadsDone();
```
gaschler: Here, "ok" seems to have a different meaning, more like "continue".

cartographer_grpc/framework/service.cc
```diff
@@ -80,33 +84,42 @@ void Service::HandleNewConnection(Rpc* rpc, bool ok) {
...
+  RemoveIfNotPending(rpc);
+}
+
+void Service::HandleDone(Rpc* rpc, bool ok) { RemoveIfNotPending(rpc); }
+
+void Service::RemoveIfNotPending(Rpc* rpc) {
+  if (!rpc->GetRpcState(Rpc::State::DONE)->pending &&
+      !rpc->GetRpcState(Rpc::State::READ)->pending &&
+      !rpc->GetRpcState(Rpc::State::WRITE)->pending) {
+    active_rpcs_.Remove(rpc);
```
gaschler: Just mentioning: Is it necessary to post/queue a task to make sure Remove is called later?

cartographer_grpc/framework/rpc.h
cschuet: Done.

cartographer_grpc/framework/service.cc
cschuet: No it's not. The server start-up and shutdown test shows that. Here the RPCs that accept incoming connections never receive any reads or writes and are correctly cleaned up nonetheless.

cartographer_grpc/framework/service.cc
cschuet: Somehow. I mean you did ask for a READ in a situation where the client already indicated that reads are done. gRPC indicates that as a failed read and it is AFAIC the only way to learn about client-side read done.

cartographer_grpc/framework/server_test.cc
cschuet: Good point. Done.

cschuet:
@wally-the-cartographer merge

wally-the-cartographer:

PR #715, opened 2017-11-28
 Interface for MapBuilder

gaschler:
Defines an interface for MapBuilder that can be used to implement a gRPC stub.


gaschler:
Diffbase is 4f1e39d

cschuet:
@wohe The idea is that we have a gRPC-forwarding implementation of the MapBuilderInterface and the TrajectoryBuilder.

PR #716, opened 2017-11-29
Implement shared ExecutionContext for handlers.

cschuet:

### DIFF
--- a/cartographer_grpc/framework/execution_context.h
+++ b/cartographer_grpc/framework/execution_context.h
@@ -1,0 +1,62 @@
+/*
+ * Copyright 2017 The Cartographer Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef CARTOGRAPHER_GRPC_FRAMEWORK_EXECUTION_CONTEXT_H
+#define CARTOGRAPHER_GRPC_FRAMEWORK_EXECUTION_CONTEXT_H
+
+#include "cartographer/common/make_unique.h"
+#include "cartographer/common/mutex.h"
+#include "glog/logging.h"
+
+namespace cartographer_grpc {
+namespace framework {
+
+// Implementations of this class allow RPC handlers to share state among one
+// another. Using Server::SetExecutionContext(...) a server-wide
+// 'ExecutionContext' can be specified. This 'ExecutionContext' can be retrieved
+// by all implementations of 'RpcHandler' by calling
+// 'RpcHandler::GetContext<MyContext>()'.
+class ExecutionContext {
+ public:
+  // This non-movable, non-copyable class is used to broker access from various
+  // RPC handlers to the shared 'ExecutionContext'. Handles automatically lock
+  // the context they point to.
+  template <typename T>
+  class Handle {
+   public:
+    T* operator->() { return static_cast<T*>(execution_context_); }
+    Handle(cartographer::common::Mutex* lock,
+           ExecutionContext* execution_context)
+        : locker_(cartographer::common::make_unique<
+                  cartographer::common::MutexLocker>(lock)),
+          execution_context_(execution_context) {}
+    Handle(const Handle&) = delete;
+    Handle(Handle&&) = delete;
+
+   private:
+    std::unique_ptr<cartographer::common::MutexLocker> locker_;
+    ExecutionContext* execution_context_;
+  };
+  cartographer::common::Mutex* lock() { return &lock_; }
+
+ private:
+  cartographer::common::Mutex lock_;
+};
+
+}  // namespace framework
+}  // namespace cartographer_grpc
+
+#endif  // CARTOGRAPHER_GRPC_FRAMEWORK_EXECUTION_CONTEXT_H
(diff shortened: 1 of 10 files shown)


cartographer_grpc/framework/server.h
```diff
@@ -81,6 +83,9 @@ class Server {
   // Shuts down the server and all of its services.
   void Shutdown();
 
+  // Sets the server-wide context object shared between RPC handlers.
+  void SetExecutionContext(std::unique_ptr<ExecutionContext> execution_context);
```
gaschler: Perhaps the server-wide context is better named ServerContext instead of ExecutionContext?

cartographer_grpc/framework/server_test.cc
```diff
@@ -27,10 +28,16 @@ namespace cartographer_grpc {
 namespace framework {
 namespace {
 
+class MyContext : public ExecutionContext {
+ public:
+  int GetOffset() { return 10; }
```
gaschler: rename to additional_increment()?

cartographer_grpc/framework/server_test.cc
```diff
@@ -27,10 +28,16 @@ namespace cartographer_grpc {
 namespace framework {
 namespace {
 
+class MyContext : public ExecutionContext {
```
gaschler: rename to MathServerContext?

cartographer_grpc/framework/server.h
```diff
@@ -57,10 +58,11 @@ class Server {
           RpcHandlerInfo{
               RpcHandlerType::RequestType::default_instance().GetDescriptor(),
               RpcHandlerType::ResponseType::default_instance().GetDescriptor(),
-              [](Rpc* const rpc) {
+              [](Rpc* const rpc, ExecutionContext* const execution_context) {
```
gaschler: This factory function looks pretty standard. Is it going to have more logic in the future?
If not, can it not be created here but the simple logic just be executed where it is needed?

cartographer_grpc/framework/execution_context.h
```diff
@@ -0,0 +1,62 @@
...
+                  cartographer::common::MutexLocker>(lock)),
+          execution_context_(execution_context) {}
+    Handle(const Handle&) = delete;
+    Handle(Handle&&) = delete;
+
+   private:
+    std::unique_ptr<cartographer::common::MutexLocker> locker_;
+    ExecutionContext* execution_context_;
+  };
+  cartographer::common::Mutex* lock() { return &lock_; }
```
gaschler: Since the only usage is to get a Handle for an ExecutionContext, how about a method
`Handle<T> ExecutionContext::GetSynchronizedAccess()` (or a different name)? Is this possible?
This would have no more public API than needed.

cartographer_grpc/framework/execution_context.h
```diff
@@ -0,0 +1,62 @@
...
+// another. Using Server::SetExecutionContext(...) a server-wide
+// 'ExecutionContext' can be specified. This 'ExecutionContext' can be retrieved
+// by all implementations of 'RpcHandler' by calling
+// 'RpcHandler::GetContext<MyContext>()'.
+class ExecutionContext {
+ public:
+  // This non-movable, non-copyable class is used to broker access from various
+  // RPC handlers to the shared 'ExecutionContext'. Handles automatically lock
+  // the context they point to.
+  template <typename T>
```
gaschler: T -> ContextType

cartographer_grpc/framework/execution_context.h
```diff
@@ -0,0 +1,62 @@
...
+    Handle(cartographer::common::Mutex* lock,
+           ExecutionContext* execution_context)
+        : locker_(cartographer::common::make_unique<
+                  cartographer::common::MutexLocker>(lock)),
+          execution_context_(execution_context) {}
+    Handle(const Handle&) = delete;
+    Handle(Handle&&) = delete;
+
+   private:
+    std::unique_ptr<cartographer::common::MutexLocker> locker_;
```
gaschler: Seems like the unique_ptr can be omitted.

cartographer_grpc/framework/execution_context.h
```diff
@@ -0,0 +1,62 @@
...
+// 'ExecutionContext' can be specified. This 'ExecutionContext' can be retrieved
+// by all implementations of 'RpcHandler' by calling
+// 'RpcHandler::GetContext<MyContext>()'.
+class ExecutionContext {
+ public:
+  // This non-movable, non-copyable class is used to broker access from various
+  // RPC handlers to the shared 'ExecutionContext'. Handles automatically lock
+  // the context they point to.
+  template <typename T>
+  class Handle {
```
gaschler: The name "Handle" reminds me more of on ID. What it does is providing synchronized/synchronous access. How about the name "Synchronized" or "LockedAccess" or "SharedAccess"?

cartographer_grpc/framework/execution_context.h
cschuet: Done.

cartographer_grpc/framework/execution_context.h
cschuet: I like "Synchronized<Synchronized>"

cartographer_grpc/framework/execution_context.h
cschuet: Done.

cartographer_grpc/framework/execution_context.h
cschuet: I can add that, but it would not be used at the moment. In RpcHandler I have to construct Synchronized via implicit construction using a list initializer, because Synchronized is not copyable. I would like to keep Synchronized non-copyable so people can't store it an that way sit on the lock.

cartographer_grpc/framework/server.h
cschuet: We need the type information to instantiate the concrete handler type. So somehow that type information needs to get to the place where the handler is instantiated, i.e. Rpc. Here we "capture" the type information in the closure. An alternative would be to have templated `RpcImpl<HandlerType>` that implement Rpc. But that would only move the problem to Rpc which then also have to be created using some factory.

cartographer_grpc/framework/server.h
cschuet: The problem with that is that there is already a ::grpc::ServerContext that is store in every Rpc object. Would you prefer `HandlerContext`?

cartographer_grpc/framework/server_test.cc
cschuet: Done.

cartographer_grpc/framework/server_test.cc
cschuet: Done.

cschuet:
@wally-the-cartographer merge

wally-the-cartographer:

wally-the-cartographer:

PR #718, opened 2017-11-30
 Create AutoDiffCostFunction in cost functions.

gaschler:
Creating a ceres::AutoDiffCostFunction requires specifying numbers
of residuals and variables, so it is safer to implement this within
the cost functions, which know best.


cartographer/mapping_2d/scan_matching/rotation_delta_cost_functor.h
```diff
@@ -42,6 +41,14 @@ class RotationDeltaCostFunctor {
   }
 
  private:
+  // Constructs a new RotationDeltaCostFunctor for the given 'angle'.
```
cschuet: update comment and move to static public factory method?

gaschler:
@cschuet PTAL at the late commit f68cd854d9e8147f6108c718b7594b14c1202a5c

PR #719, opened 2017-11-30
Implement unary gRPC calls.

cschuet:


cartographer_grpc/framework/service.cc
```diff
@@ -65,6 +63,9 @@ void Service::HandleEvent(Rpc::Event event, Rpc* rpc, bool ok) {
     case Rpc::Event::WRITE:
       HandleWrite(rpc, ok);
       break;
+    case Rpc::Event::FINISH:
+      HandleFinish(rpc, ok);
+      break;
     case Rpc::Event::DONE:
       HandleDone(rpc, ok);
       break;
```
gaschler: (Not in this PR) Add default log fatal.

cschuet: I left out the default to guarantee that we get a compile error when we forget to handle a case.

PR #726, opened 2017-12-04
WIP: Refactor worker dispatch logic

jihoonl:
As discussed in #563

@cschuet  This is the rough logic how I get CHECK failure in constraint builder. This is how I reproduced. 

* start localization trajectory
* A few seconds later, finish it and start the new pure localization trajectory.
* Then, kill cartographer_node.

```
[/cartographer_node  INFO 1512404436.214935386, 1498225483.460233338]: I1204 17:20:36.000000 30475 submaps.cc:190] Added submap 1
[/cartographer_node  INFO 1512404436.215024605, 1498225483.460233338]: I1204 17:20:36.000000 30475 map_builder_bridge.cc:102] Added trajectory with ID '1'.
[/cartographer_node  INFO 1512404436.486713140, 1498225483.732305270]: I1204 17:20:36.000000 30475 ordered_multi_queue.cc:172] All sensor data for trajectory 1 is available starting at '636338222837143499'.
[/cartographer_node  INFO 1512404436.486774292, 1498225483.732305270]: I1204 17:20:36.000000 30475 local_trajectory_builder.cc:227] Extrapolator not yet initialized.
[/rviz  INFO 1512404442.862798002, 1498225490.109016016]: Setting pose: -0.975 3.187 0.065 [frame=map]
[/cartographer_node  INFO 1512404442.870230131, 1498225490.119075553]: I1204 17:20:42.000000 30475 node.cc:396] Shutdown the subscriber of [scan_1]
[/cartographer_node  INFO 1512404442.873909440, 1498225490.119075553]: I1204 17:20:42.000000 30475 node.cc:396] Shutdown the subscriber of [scan_2]
[/cartographer_node  INFO 1512404442.874931825, 1498225490.119075553]: I1204 17:20:42.000000 30475 node.cc:396] Shutdown the subscriber of [odom]
[/cartographer_node  INFO 1512404442.874972135, 1498225490.119075553]: I1204 17:20:42.000000 30475 map_builder_bridge.cc:124] Finishing trajectory with ID '1'...
[/cartographer_node  INFO 1512404442.883320257, 1498225490.129132641]: I1204 17:20:42.000000 30475 submaps.cc:190] Added submap 1
[/cartographer_node  INFO 1512404442.883375739, 1498225490.129132641]: I1204 17:20:42.000000 30475 map_builder_bridge.cc:102] Added trajectory with ID '2'.
[...]

PR #729, opened 2017-12-05
Refactor calling optimization into DispatchOptimization.

ojura:
I noticed that @jihoonl opened the PR #726 which performs a similar thing. As discussed in googlecartographer/cartographer_ros#613 (@cschuet  has already taken a look), I pulled this out of #481 (a really old PR whose merging has been postponed), which is an example where re-running optimization is triggered from elsewhere as well (besides from `ComputeConstraintsForNode`). This refactoring makes libcartographer friendlier for use cases such as that one.

An important detail is that I have changed the condition in `WaitForAllComputations` to also check if the work queue is empty. If there are other things on the worker queue besides `ComputeConstraintsForNode`, currently we will wrongfully conclude that all computations are done. (This detail was merged in #754, so it's no longer in the diff of this PR).

Also missing is the same thing for 3D. I can add that when we settle on this.

Also, I suggest that `run_loop_closure` gets renamed to `run_optimization`.


cschuet:
@wohe Wolfgang could you take a look at this?

ojura:
Also, I think that changing the optimization problem options from within the worker queue, and dispatching optimization just like `ComputeConstraintsForNode` does is more consistent with the promise given in
https://github.com/googlecartographer/cartographer/blob/7994b9ac6b451d1de90dc9405f52c32c0f44f8d8/cartographer/mapping_2d/pose_graph.cc#L534-L535

Sure, if we first wait for all computations (and even if we do it properly by checking if the queue is empty, like in this PR), we can change the options in `optimization_problem_` without having to worry about the promise above. But what if _someone_ pushes _something_ to the worker queue during the final optimization? Chaos ensues (e.g. check in `WhenDone` fails). By making the final optimization a proper citizen of the worker queue, we don't have to worry about that at all.

cartographer/mapping_2d/pose_graph.cc
```diff
@@ -506,14 +509,21 @@ void PoseGraph::AddTrimmer(std::unique_ptr<mapping::PoseGraphTrimmer> trimmer) {
...
+          options_.max_num_final_iterations());
+      DispatchOptimization();
+    });
+    AddWorkItem([this]() REQUIRES(mutex_) {
+      optimization_problem_.SetMaxNumIterations(
+          options_.optimization_problem_options()
+              .ceres_solver_options()
+              .max_num_iterations());
+    });
+  }
```
cschuet: Can you explain the motivation for this change? Naively I would say that the below is more readable.

cschuet: Oh, yes read your comment above which explains just that. So makes sense.

ojura:
We can mirror this for 3D in a separate PR, or in this one, just let me know.

ojura:
Updated.

cschuet:
@wally-the-cartographer merge

wally-the-cartographer:
Merge requested by authorized user cschuet. Merge queue now has a length of 2.

PR #740, opened 2017-12-06
 Remove empty leftover of pose_estimate.h

MichaelGrupp:
Follow-up of #670.

PR #741, opened 2017-12-06
Delete empty pose_estimate.h

ojura:
Ooops, I haven't deleted the file properly in #670. #740 should be merged first though.


ojura:
Ooops, I see that #740 also has this covered. Closing.

PR #747, opened 2017-12-08
Add binary for cartographer_grpc_server.

cschuet:


cartographer_grpc/map_builder_server_main.cc
```diff
@@ -0,0 +1,58 @@
...
+      LoadMapBuilderServerOptions(configuration_directory,
+                                  configuration_basename);
+  MapBuilderServer map_builder_server(map_builder_server_options);
+  map_builder_server.Start();
+  map_builder_server.WaitForShutdown();
+}
+
+}  // namespace cartographer_grpc
+
+int main(int argc, char** argv) {
```
gaschler: Obviously, there is a little common code in common, but I also don't see how to reuse it nicely.

configuration_files/map_builder_server.lua
```diff
@@ -0,0 +1,23 @@
...
+
+include "map_builder.lua"
+
+options = {
+  server_address = "0.0.0.0:50051",
+  num_grpc_threads = 4,
+  map_builder = MAP_BUILDER,
+}
+
+return options
```
gaschler: This return prohibits that this file can be included, which is done for the other config files.
It may be better to have (an additional) default config file without return in the future (future PR).

cschuet:
@wally-the-cartographer merge

wally-the-cartographer:

PR #750, opened 2017-12-09
Makes Rpc::Write() thread-safe.

cschuet:


cartographer_grpc/framework/rpc.h
```diff
@@ -121,11 +128,12 @@ class ActiveRpcs {
   ~ActiveRpcs() EXCLUDES(lock_);
 
   Rpc* Add(std::unique_ptr<Rpc> rpc) EXCLUDES(lock_);
+  std::shared_ptr<Rpc> Get(Rpc* rpc) EXCLUDES(lock_);
   bool Remove(Rpc* rpc) EXCLUDES(lock_);
 
  private:
   cartographer::common::Mutex lock_;
-  std::unordered_set<Rpc*> rpcs_;
+  std::unordered_map<Rpc*, std::shared_ptr<Rpc>> rpcs_;
```
gaschler: This complex data structure shouldn't be necessary for this behavior.
Rpc should create shared_ptr itself.

cartographer_grpc/framework/service.cc
```diff
@@ -52,7 +52,7 @@ void Service::StartServing(
 void Service::StopServing() { shutting_down_ = true; }
 
 void Service::HandleEvent(Rpc::Event event, Rpc* rpc, bool ok) {
-  rpc->GetRpcEvent(event)->pending = false;
+  rpc->UpdateRpcEventState(event, false);
```
gaschler: naming: SetRpcEventState or better SetEventState (this is obviously about the Rpc)

cartographer_grpc/framework/rpc.cc
```diff
@@ -305,16 +326,23 @@ ActiveRpcs::~ActiveRpcs() {
...
+  auto it = rpcs_.find(rpc);
+  CHECK(it != rpcs_.end()) << "RPC " << rpc << " not found.";
+  return it->second;
 }
 
 bool ActiveRpcs::Remove(Rpc* rpc) {
   cartographer::common::MutexLocker locker(&lock_);
   auto it = rpcs_.find(rpc);
   if (it != rpcs_.end()) {
-    delete rpc;
```
gaschler: Who deletes Rpc now?

cartographer_grpc/framework/rpc.cc
```diff
@@ -305,16 +326,23 @@ ActiveRpcs::~ActiveRpcs() {
 
 Rpc* ActiveRpcs::Add(std::unique_ptr<Rpc> rpc) {
   cartographer::common::MutexLocker locker(&lock_);
-  const auto result = rpcs_.emplace(rpc.release());
+  std::shared_ptr<Rpc> shared_rpc = std::move(rpc);
```
gaschler: I think the shared_ptr should be made by Rpc itself, see other comment.

cartographer_grpc/framework/rpc.cc
cschuet: The `shared_ptr` does, when the last one goes out of scope. The one stored in ActiveRpcs is deleted in the line below this one.

cartographer_grpc/framework/rpc.h
cschuet: I am not sure what you mean by "Rpc should create shared_ptr itself". `shared_ptr`s form cliques. So you can't create `shared_ptr`s for an object by a member function like this 

~~~
shared_ptr<Rpc> Rpc::GetSharedPtr() {
  return shared_ptr<Rpc>(this);
}
~~~
Creating a shared pointer by calling `shared_ptr<Rpc>(rpc)` starts a new clique and moves ownership of `rpc` to the clique. You extend the clique by making copies of the shared pointer. If you created another shared pointer with `shared_ptr<Rpc>(rpc)` you would get a double free when the second shared_ptr goes out of scope.

cartographer_grpc/framework/rpc.cc
cschuet: See my comment below.

cschuet:
> Is it possible to avoid the hard-to-verify locking by using a pattern where the users of Rpc can subscribe to its events and the Rpc then calls multiple users one after another?
> 
> Also, the ownership of Rpc looks too complex.
> Sorry I have to ask again, couldn't ActiveRpcs just own the Rpcs?
> (Then, maybe some other object that outlives ActiveRpcs should own the Rpcs.)

Consider the control flow for streaming sensor data. gRPC events are generated in the library and we learn about them by them showing up on one of our completion queues. `Server` runs those queues and pushes the events through the respective `Service` into the `Rpc`s responsible for the connection that generated them. The only *user* of those events is the attached handler and since we only support one handler per connection we don't need a subscription but we just call into the handler.

The situation is different for local SLAM results. The availability of a local SLAM result is an event that is generated in the local SLAM thread. This PR essentially proposes that in order to send out the event to all clients that subscribed to local SLAM results for the respective trajectory, we need to find the responsible `Rpc` objects (since only they have all the knowledge we need to push out a gRPC message, e.g. `async_reader_writer_interface`) and perform the `Write` by calling `Rpc::Write` directly from the local SLAM thread. That's why `Rpc::Write` needs to be thread-safe.

Here are two alternatives that avoid thread-safety.

1. We have a queue of local SLAM results to send out in the handler which needs to be thread-safe, e.g. `BlockingQueue`. We add an item from the local SLAM thread to the handler's queue. When the handler wakes up for some other reason (through an event on the completion queue) it also checks the queue and sends out anything in that *outbox*. Problem: what if the handler never wakes up because there is radio silence from the client. The client would essentially have to poll to avoid that.
2. We figure out a way to add our custom local-SLAM-result-available event into the completion queue as a custom event. This way it would enter the system by the usual means and be processed by the gRPC thread. Adding the event from the local SLAM thread to the completion queue is safe since it is already thread-safe. Problem: I find this ties us to strongly to implementation details of gRPC, *plus* I haven't found a way to add those events to the completion queue.

I think the "proper" solution would involve us building our own event queues. We'd have one thread which does nothing but dequeue events from a single gRPC completion queue and push them into our own event queues. Each event queue is serviced by a thread (like the gRPC completion queues now) and the thread would dequeue and push those events through `Service` to the respective `Rpc` object. This way we'd have more control over the queues and could easily have custom events like `write-needed`.

cartographer_grpc/framework/rpc.h
gaschler: Ok, I see that my suggestion was not a good idea. There could be an alternative around std::enable_shared_from_this, but that also sounds fishy. Let's stick to the current solution.

PR #751, opened 2017-12-12
Test global SLAM.

gaschler:


gaschler:
On the current master, this test is flaky because RunFinalOptimization does not actually finish all computations, and the following check (sometimes) fails during deletion: [pose_graph.cc:51] Check failed: work_queue_ == nullptr
I believe the public API should allow to finish all computations and delete everything.
At first glance, #729 seems to fix this (and change the meaning of RunFinalOptimization to finish the entire work queue).

cschuet:
As discussed let's fix the issue with the work queue first.

gaschler:
@cschuet This can now be reviewed, the dependency is merged.

cartographer/mapping/map_builder_test.cc
```diff
@@ -33,9 +33,9 @@ namespace {
 
 constexpr char kRangeSensorId[] = "range";
 constexpr char kIMUSensorId[] = "imu";
-constexpr double kDuration = 2.;
+constexpr double kDuration = 4.;
```
cschuet: Should we add units to these constants?

gaschler:
@wally-the-cartographer merge

gaschler:
@wally-the-cartographer merge

PR #763, opened 2017-12-15
PopWithTimeout for sensor data queue

gaschler:
PopWithTimeout is necessary to ensure the SLAM thread
in the MapBuilderServer finishes when the server is
shut down.

### DIFF
--- a/cartographer_grpc/map_builder_server.cc
+++ b/cartographer_grpc/map_builder_server.cc
@@ -26,10 +26,17 @@
 #include "glog/logging.h"
 
 namespace cartographer_grpc {
+namespace {
+
+const cartographer::common::Duration kPopTimeout =
+    cartographer::common::FromMilliseconds(100);
+
+}  // namespace
 
 MapBuilderServer::MapBuilderContext::MapBuilderContext(
     cartographer::mapping::MapBuilder* map_builder,
-    cartographer::common::BlockingQueue<SensorData>* sensor_data_queue)
+    cartographer::common::BlockingQueue<std::unique_ptr<SensorData>>*
+        sensor_data_queue)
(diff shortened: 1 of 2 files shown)


cartographer_grpc/map_builder_server.cc
```diff
@@ -26,18 +26,26 @@
...
+
+const cartographer::common::Duration kPopTimeout =
+    cartographer::common::FromMilliseconds(100);
+
+}  // namespace
 
 MapBuilderServer::MapBuilderContext::MapBuilderContext(
     cartographer::mapping::MapBuilder* map_builder,
-    cartographer::common::BlockingQueue<SensorData>* sensor_data_queue)
+    cartographer::common::BlockingQueue<std::unique_ptr<SensorData>>*
```
cschuet: I don't think it is necessary to wrap `SensorData` in a `std::unique_ptr` - it just adds noise. After all `SensorData` already wraps the large data in a `std::unique_ptr`

~~~
struct SensorData {
    int trajectory_id;
    std::unique_ptr<cartographer::sensor::Data> sensor_data;
};
~~~

gaschler: As discussed.
For documentation: unique_ptr here is used as an "optional" so PopWithTimeout can signal if the timeout was reached.

cschuet: Sorry as discussed I did not understand `PopWithTimeout` of course wants to return a sort of optional and therefore introduction of `std::unique_ptr` is necessary. We checked whether we can remove the `std::unique_ptr` from `SensorData` now, but we can't since `Data` is just a base class.

cschuet:
@wally-the-cartographer merge

wally-the-cartographer:

PR #764, opened 2017-12-15
TrajectoryBuilderStub implementation

gaschler:

### DIFF
--- a/cartographer_grpc/mapping/trajectory_builder_stub.cc
+++ b/cartographer_grpc/mapping/trajectory_builder_stub.cc
@@ -27,30 +27,31 @@ TrajectoryBuilderStub::TrajectoryBuilderStub(
     : client_channel_(client_channel), trajectory_id_(trajectory_id) {
   stub_ = proto::MapBuilderService::NewStub(client_channel_);
   CHECK(stub_) << "Failed to create stub.";
-  rangefinder_writer_.client_writer = stub_->AddRangefinderData(
-      &rangefinder_writer_.client_context, &rangefinder_writer_.response);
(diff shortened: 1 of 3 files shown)


cartographer_grpc/mapping/trajectory_builder_stub.cc
```diff
@@ -16,38 +16,83 @@
...
 
 TrajectoryBuilderStub::TrajectoryBuilderStub(
-    std::shared_ptr<grpc::Channel> client_channel,
-    proto::MapBuilderService::Stub* stub)
-    : client_channel_(client_channel), stub_(stub) {}
+    std::shared_ptr<grpc::Channel> client_channel, const int trajectory_id)
+    : client_channel_(client_channel), trajectory_id_(trajectory_id) {
+  stub_ = proto::MapBuilderService::NewStub(client_channel_);
+  CHECK(stub_) << "Failed to create stub.";
+  rangefinder_writer_.client_writer = stub_->AddRangefinderData(
```
cschuet: We should find out what actually happens when you call e.g. `stub_->AddRangefinderData`. Does that open a TCP connection? Is there some network cost associated with holding a client writer even when you are not using it? If so we might want to create the writers lazily as we need them. I guess only in very few cases a bridge will actually want to send all 4 types of sensor data.

gaschler: Right. As discussed, we create each writer lazily before using it the first time.

gaschler:
@cschuet PTAL at the changes

cschuet:
@wally-the-cartographer merge

wally-the-cartographer:

PR #783, opened 2017-12-21
Adding a minimal implementation of std::optional.

spielawa:
We're using C++11, which doesn't have std::optional. We need a few
features of std::optional for improved GPS support. These are
implemented here.

### DIFF
--- a/cartographer/common/optional.h
+++ b/cartographer/common/optional.h
@@ -19,7 +19,7 @@
 
 #include <memory>
 
-#include "common/make_unique.h"
+#include "cartographer/common/make_unique.h"
 #include "glog/logging.h"
 
 namespace cartographer {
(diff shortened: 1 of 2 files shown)


cartographer/common/optional.h
```diff
@@ -0,0 +1,39 @@
+#ifndef CARTOGRAPHER_COMMON_OPTIONAL_H_
```
wohe: This needs a license comment. Same with the other file.

spielawa: done

spielawa:
I signed it!

SirVer:
@googlebot ping?

cschuet:
@SirVer @spielawa We checked the Google CLA database and Github user spielawa has not signed the CLA yet. Can you double check that you actually specified the right email address when signing?

spielawa:
I signed it!

wohe:
@wally-the-cartographer merge

wally-the-cartographer:

PR #784, opened 2017-12-27
Pose graph: do not mark all submaps finished when finalizing a trajectory

ojura:
I think this was a mistake in #563 (@jihoonl). Not all submaps should be marked as finished, only those pertaining to the trajectory being finalized.

### DIFF
--- a/cartographer/mapping_2d/pose_graph.cc
+++ b/cartographer/mapping_2d/pose_graph.cc
@@ -381,9 +381,8 @@ void PoseGraph::FinishTrajectory(const int trajectory_id) {
     CHECK_EQ(finished_trajectories_.count(trajectory_id), 0);
     finished_trajectories_.insert(trajectory_id);
 
-    auto submap_data = optimization_problem_.submap_data();
-    for (auto submap_id_data : submap_data) {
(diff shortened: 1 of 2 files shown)


cartographer/mapping_2d/pose_graph.cc
```diff
@@ -381,9 +381,8 @@ void PoseGraph::FinishTrajectory(const int trajectory_id) {
     CHECK_EQ(finished_trajectories_.count(trajectory_id), 0);
     finished_trajectories_.insert(trajectory_id);
 
-    auto submap_data = optimization_problem_.submap_data();
-    for (auto submap_id_data : submap_data) {
-      submap_data_.at(submap_id_data.id).state = SubmapState::kFinished;
+    for (const auto& submap : submap_data_.trajectory(trajectory_id)) {
+      submap_data_.at(submap.id).state = SubmapState::kFinished;
```
ojura: This could look nicer if `MapById` provided a non-const iterator as well.

cschuet:
@wally-the-cartographer merge

wally-the-cartographer:

jihoonl:
Ah good catch. Thanks

PR #785, opened 2018-01-03
Define default num_event_threads config.

gaschler:
(This was not covered by tests because they manually set it to 1.)


gaschler:
@wally-the-cartographer merge

wally-the-cartographer:

PR #793, opened 2018-01-05
Fix RpcEvent lifetime

### LINKED ISSUE #788: RpcEvents are not always deleted.
Asan reports cases where an RpcEvent is created but not deleted.
For instance, cartographer_grpc::framework::Rpc::RequestNextMethodInvocation
creates a new RpcEvent as a tag to pass to grpc::ServerContext::AsyncNotifyWhenDone,
but this does not always return the tag to our event queue
(so we cannot delete it).

Here is the stack trace:
```
Direct leak of 160 byte(s) in 5 object(s) allocated from:
    #0 0x60be40 in operator new(unsigned long) ( cartographer/cartographer_grpc.framework.server_test+0x60be40)
    #1 0x62e2c2 in cartographer_grpc::framework::Rpc::RequestNextMethodInvocation()  cartographer/cartographer_grpc/framework/rpc.cc:83:7
    #2 0x6685ee in cartographer_grpc::framework::Service::StartServing(std::vector<cartographer_grpc::framework::CompletionQueueThread, std::allocator<cartographer_grpc::framework::CompletionQueueThread> >&, cartographer_grpc::framework::ExecutionContext*)  cartographer/cartographer_grpc/framework/service.cc:52:7
    #3 0x63ec12 in cartographer_grpc::framework::Server::Start()  cartographer/cartographer_grpc/framework/server.cc:130:5
    #4 0x60f56a in cartographer_grpc::framework::(anonymous namespace)::ServerTest_StartAndStopServerTest_Test::TestBody()  cartographer/cartographer_grpc/framework/server_test.cc:153:3
```

A possible fix is that
- Rpc owns a number of RpcEvents for the different callbacks by grpc
- The RpcEvent with Event::WriteNeeded, which needs to be created dynamically, is not passed to grpc but directly to the event queue and deleted after being processed.

gaschler:
Fixes #788.

Uses two different types of events whether the event goes through the CompletionQueue or not.
CompletionQueueRpcEvent is again a member of Rpc.

### DIFF
--- a/cartographer_grpc/framework/rpc.h
+++ b/cartographer_grpc/framework/rpc.h
@@ -57,13 +57,31 @@ class Rpc {
     const Event event;
   };
 
-  using EventQueue = cartographer::common::BlockingQueue<EventBase*>;
+
+  class EventDeleter {
+   public:
+    EventDeleter() : needs_deletion_(true) {}
+    EventDeleter(bool needs_deletion) : needs_deletion_(needs_deletion) {}
+    void operator()(EventBase* e) {
+      if (e != nullptr && needs_deletion_) {
+        delete e;
+      }
+    }
+
+   private:
+    bool needs_deletion_;
+  };
+
+  using UniqueEventPtr = std::unique_ptr<EventBase, EventDeleter>;
+  using EventQueue = cartographer::common::BlockingQueue<UniqueEventPtr>;
 
   // Flows through gRPC's CompletionQueue and then our EventQueue.
   struct CompletionQueueRpcEvent : public EventBase {
     CompletionQueueRpcEvent(Event event, Rpc* rpc)
         : EventBase(event), rpc_ptr(rpc), ok(false), pending(false) {}
-    void PushToEventQueue() { rpc_ptr->event_queue()->Push(this); }
+    void PushToEventQueue() {
+      rpc_ptr->event_queue()->Push(UniqueEventPtr(this, EventDeleter(false)));
+    }
     void Handle() override;
 
     Rpc* rpc_ptr;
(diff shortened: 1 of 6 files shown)


gaschler:
I also tried to make EventQueue hold unique_ptr with two different deleters. However, EventQueue sometimes returns a nullptr, and thereby assigns no deleter, so this approach did not work.

Other alternatives would be to
- make a shallow copy of CompletionQueueRpcEvent when it goes into the EventQueue that may be deleted
- do some renaming and move the "delete this" somewhere less surprising (like, have a method DeleteIfNeeded)

cschuet:
> I also tried to make EventQueue hold unique_ptr with two different deleters. However, EventQueue sometimes returns a nullptr, and thereby assigns no deleter, so this approach did not work.

The trick is to
1. Use a single deleter with a boolean member indicating whether to perform the deletion.
2. Make the deleter default constructable, so that the unique_ptr can be implicitly created from the `return nullptr;` statements in BlockingQueue.

Here is a PR against your repo that does that https://github.com/gaschler/cartographer/pull/1

cartographer_grpc/framework/rpc.h
```diff
@@ -50,11 +48,58 @@ class Rpc {
...
+
     const Event event;
-    std::weak_ptr<Rpc> rpc;
+  };
+
+  class EventDeleter {
+   public:
+    enum Action { DELETE = 0, DO_NOT_DELETE };
+
+    EventDeleter() : action_(DELETE) {}
```
cschuet: Add comment why we need this guy?

gaschler:
@wally-the-cartographer merge

wally-the-cartographer:

PR #799, opened 2018-01-09
Make LocalTrajectoryBuilder optional.

cschuet:
This is to prepare for the cloud-based mapping case where the robot is running cartographer in pure-localization mode and the cloud instance only solves the global problem. In that case a `LocalTrajectoryBuilder` need to be instantiated in for the cloud instance.

### DIFF
--- a/cartographer/internal/mapping/global_trajectory_builder.h
+++ b/cartographer/internal/mapping/global_trajectory_builder.h
@@ -29,12 +29,12 @@ template <typename LocalTrajectoryBuilder,
 class GlobalTrajectoryBuilder : public mapping::TrajectoryBuilderInterface {
  public:
   GlobalTrajectoryBuilder(
-      const LocalTrajectoryBuilderOptions& options, const int trajectory_id,
-      PoseGraph* const pose_graph,
+      std::unique_ptr<LocalTrajectoryBuilder> local_trajectory_builder,
+      const int trajectory_id, PoseGraph* const pose_graph,
       const LocalSlamResultCallback& local_slam_result_callback)
       : trajectory_id_(trajectory_id),
         pose_graph_(pose_graph),
-        local_trajectory_builder_(options),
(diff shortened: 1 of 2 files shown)


cartographer/internal/mapping/global_trajectory_builder.h
```diff
@@ -44,8 +44,10 @@ class GlobalTrajectoryBuilder : public mapping::TrajectoryBuilderInterface {
   void AddSensorData(
       const std::string& sensor_id,
       const sensor::TimedPointCloudData& timed_point_cloud_data) override {
+    CHECK(local_trajectory_builder_)
+        << "Cannot add TimedPointCloudData without a LocalTrajectoryBuilder";
```
wohe: nit: End sentence with a period.

cartographer/internal/mapping/global_trajectory_builder.h
```diff
@@ -44,8 +44,10 @@ class GlobalTrajectoryBuilder : public mapping::TrajectoryBuilderInterface {
   void AddSensorData(
       const std::string& sensor_id,
       const sensor::TimedPointCloudData& timed_point_cloud_data) override {
+    CHECK(local_trajectory_builder_)
```
wohe: Add a comment about this in the header file, i.e. that `nullptr` is acceptable but no `TimedPointCloudData` can be added in this case?

cartographer/internal/mapping/global_trajectory_builder.h
cschuet: Done.

cartographer/internal/mapping/global_trajectory_builder.h
cschuet: Done.

cschuet:
@wally-the-cartographer merge

wally-the-cartographer:

PR #801, opened 2018-01-09
Add InsertionResult to LocalSlamResult.

cschuet:
As discussed on https://github.com/googlecartographer/cartographer/pull/574 I need to include the `InsertionResult` in the `LocalSlamResult` in order to add a `Node` in the `PoseGraph` in the cloud Cartographer instance.

### DIFF
--- a/cartographer/internal/mapping/global_trajectory_builder.h
+++ b/cartographer/internal/mapping/global_trajectory_builder.h
@@ -62,10 +62,19 @@ class GlobalTrajectoryBuilder : public mapping::TrajectoryBuilderInterface {
               matching_result->insertion_result->insertion_submaps));
       CHECK_EQ(node_id->trajectory_id, trajectory_id_);
     }
+    auto insertion_result = common::make_unique<InsertionResult>();
+    insertion_result->constant_data =
+        matching_result->insertion_result->constant_data;
+    for (const auto& insertion_submap :
+         matching_result->insertion_result->insertion_submaps) {
+      insertion_result->insertion_submaps.push_back(
+          std::static_pointer_cast<const Submap>(insertion_submap));
+    }
     if (local_slam_result_callback_) {
       local_slam_result_callback_(
           trajectory_id_, matching_result->time, matching_result->local_pose,
-          std::move(matching_result->range_data_in_local), std::move(node_id));
+          std::move(matching_result->range_data_in_local), std::move(node_id),
+          std::move(insertion_result));
     }
   }
 
--- a/cartographer/internal/mapping/global_trajectory_builder.h
+++ b/cartographer/internal/mapping/global_trajectory_builder.h
@@ -58,16 +58,14 @@ class GlobalTrajectoryBuilder : public mapping::TrajectoryBuilderInterface {
       // The range data has not been fully accumulated yet.
       return;
     }
-    std::unique_ptr<mapping::NodeId> node_id;
     std::unique_ptr<InsertionResult> insertion_result;
     if (matching_result->insertion_result != nullptr) {
-      node_id = ::cartographer::common::make_unique<mapping::NodeId>(
-          pose_graph_->AddNode(
-              matching_result->insertion_result->constant_data, trajectory_id_,
-              matching_result->insertion_result->insertion_submaps));
-      CHECK_EQ(node_id->trajectory_id, trajectory_id_);
+      auto node_id = pose_graph_->AddNode(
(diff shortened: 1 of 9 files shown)


cartographer/internal/mapping/global_trajectory_builder.h
```diff
@@ -59,17 +59,26 @@ class GlobalTrajectoryBuilder : public mapping::TrajectoryBuilderInterface {
...
     if (matching_result->insertion_result != nullptr) {
       node_id = ::cartographer::common::make_unique<mapping::NodeId>(
           pose_graph_->AddNode(
               matching_result->insertion_result->constant_data, trajectory_id_,
               matching_result->insertion_result->insertion_submaps));
       CHECK_EQ(node_id->trajectory_id, trajectory_id_);
+      insertion_result = common::make_unique<InsertionResult>();
+      insertion_result->constant_data =
+          matching_result->insertion_result->constant_data;
+      for (const auto& insertion_submap :
```
wohe: Why not just
```
insertion_result->insertion_submaps = matching_result->insertion_result->insertion_submaps
```
or
```
common::make_unique<InsertionResult>(InsertionResult{
  matching_result->insertion_result->constant_data;
  matching_result->insertion_result->insertion_submaps;
});
```

cartographer/mapping/trajectory_builder_interface.h
```diff
@@ -44,14 +44,20 @@ proto::TrajectoryBuilderOptions CreateTrajectoryBuilderOptions(
 // optimized pose estimates.
 class TrajectoryBuilderInterface {
  public:
+  struct InsertionResult {
+    std::shared_ptr<const mapping::TrajectoryNode::Data> constant_data;
+    std::vector<std::shared_ptr<const Submap>> insertion_submaps;
+  };
```
wohe: Wouldn't it make sense to include the `NodeId` here as well and just pass one struct?

cartographer/internal/mapping/global_trajectory_builder.h
cschuet: That does not compile.

Note that the `matching_result->insertion_result->insertion_submaps` are of type `std::vector<mapping_2d/3d::Submap>` while the newly introduced one's are `std::vector<mapping::Submap>`, i.e. the base class.

cartographer/mapping/trajectory_builder_interface.h
cschuet: Is it true that you have a `NodeId` iff you have an `InsertionResult`.

wohe: I think so. The comment below also says that.

cartographer/internal/mapping/global_trajectory_builder.h
wohe: I see. What about the range constructor then, i.e.
```
common::make_unique<InsertionResult>(InsertionResult{
  matching_result->insertion_result->constant_data,
  std::vector<std::shared_ptr<const Submap>>(
    matching_result->insertion_result->insertion_submaps.begin(),
    matching_result->insertion_result->insertion_submaps.end())});
```
?

cartographer/internal/mapping/global_trajectory_builder.h
```diff
@@ -59,17 +59,24 @@ class GlobalTrajectoryBuilder : public mapping::TrajectoryBuilderInterface {
       return;
     }
     std::unique_ptr<mapping::NodeId> node_id;
```
wohe: Remove?

cartographer/internal/mapping/global_trajectory_builder.h
```diff
@@ -59,17 +59,24 @@ class GlobalTrajectoryBuilder : public mapping::TrajectoryBuilderInterface {
...
+    std::unique_ptr<InsertionResult> insertion_result;
     if (matching_result->insertion_result != nullptr) {
       node_id = ::cartographer::common::make_unique<mapping::NodeId>(
```
wohe: Remove the `unique_ptr`which is no longer used?

cartographer/internal/mapping/global_trajectory_builder.h
cschuet: Thanks!

cartographer/internal/mapping/global_trajectory_builder.h
cschuet: Done.

cschuet:
@wally-the-cartographer merge
@wohe Thanks for the review!

PR #805, opened 2018-01-10
In optimization_problem, use a map instead of vector for trajectory data

spielawa:
PAIR=wohe


wohe:
@wally-the-cartographer merge

wally-the-cartographer:

PR #816, opened 2018-01-12
Draft: TrajectoryCollator

gaschler:
[RFC=0008](https://github.com/googlecartographer/rfcs/blob/master/text/0008-collator-interface.md)

This is still a draft because the are a few open design questions:
- Should CollatorInferface::GetBlockingTrajectoryId return an optional? There is no useful return for per-trajectory collators, but MapBuilder should hold the interface to be flexible.
- Unfortunately, OrderedMultiQueue uses trajectory_id GetCommonStartTime so it has more logic for the per-trajectory case than it needs to.
- Should TrajectoryCollator have two maps or a map to struct?


gaschler:
I'm closing this one in favor of the simpler solution https://github.com/googlecartographer/cartographer/pull/827

PR #831, opened 2018-01-19
Send LandmarkData via gRPC.

pifon2a:
[RFC PR](https://github.com/googlecartographer/rfcs/pull/18)


cartographer_grpc/sensor/serialization.h
```diff
@@ -28,25 +29,34 @@ namespace sensor {
 
 void CreateSensorMetadata(const std::string& sensor_id, int trajectory_id,
                           proto::SensorMetadata* proto);
+
```
cschuet: We only use newlines to group methods that are somehow belong to one another. So please remove the newlines.

pifon2a: Done.

PR #834, opened 2018-01-22
Add a Bazel build for cartographer

drigz:
This does not include cartographer_grpc, nor any of the binaries
(*_main.cc). It has been tested with Bazel 0.9.0 on Ubuntu 14.04 with
the latest (at time of writing) dazel/dazel container, by running:

```
bazel build //...
bazel test //...
```

While this is a large CL, the files under third_party are not new - they have been through internal review, and this is their open-source release. The other files come to 212 lines of code.


tools/bazel.rc
```diff
@@ -0,0 +1,17 @@
+# Copyright 2018 The Cartographer Authors
```
cschuet: Is it important that this is in tools? We have a cmake subdirectory. Should we rather have this file in a bazel subdirectory?

cschuet: As discussed this one needs to stay but others have gone into the subfolder.

cschuet:
@wally-the-cartographer merge

wally-the-cartographer:

tools/bazel.rc
drigz: https://github.com/bazelbuild/bazel/issues/4502

PR #836, opened 2018-01-22
Rename 'tranform' to 'landmark_to_tracking_transform'.

pifon2a:
[RFC PR](https://github.com/googlecartographer/rfcs/pull/18)


cartographer/sensor/landmark_data.cc
```diff
@@ -40,7 +41,7 @@ LandmarkData FromProto(const proto::LandmarkData& proto) {
   for (const auto& item : proto.landmarks()) {
```
gaschler: drive-by: landmark_data.landmarks.reserve()?

PR #841, opened 2018-01-23
Build cartographer_grpc with Bazel

drigz:
This doesn't work when cartographer is an external workspace, due to a
limitation in cc_grpc_library. However, the //cartographer target will
still work it is external, and cartographer_grpc users can vendor the
repository.


cschuet:
@wally-the-cartographer merge

wally-the-cartographer:

PR #859, opened 2018-01-29
Serialize trajectory builder options

ojura:
Replaces #553. Towards #549.

The options proto is written as a second proto in the stream, behind the pose graph proto.

### DIFF
--- a/cartographer/mapping/map_builder.cc
+++ b/cartographer/mapping/map_builder.cc
@@ -129,12 +129,24 @@ int MapBuilder::AddTrajectoryBuilder(
         transform::ToRigid3(initial_trajectory_pose.relative_pose()),
         common::FromUniversal(initial_trajectory_pose.timestamp()));
   }
+  proto::TrajectoryBuilderOptionsWithSensorIds options_with_sensor_ids_proto;
+  for (const auto& sensor_id : expected_sensor_ids) {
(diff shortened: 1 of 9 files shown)


ojura:
@gaschler I updated this to also serialize the new sensor IDs, PTAL. This has been sitting since September, it would be nice if we could merge it.

cartographer/mapping/map_builder.cc
```diff
@@ -327,5 +361,10 @@ int MapBuilder::num_trajectory_builders() const {
 
 PoseGraphInterface* MapBuilder::pose_graph() { return pose_graph_; }
 
+const std::vector<proto::TrajectoryBuilderOptionsWithSensorIds>&
+MapBuilder::GetAllTrajectoryBuilderOptions() const {
```
cschuet: Is this currently unused?

cschuet: If there are no concrete plans to use it in a follow-up PR please remove it.

ojura: I don't have any concrete plans to use it inside libcartographer. I believe though it should be a part of the interface. E.g. the library user loads a `.pbstream` and wants to inspect/reuse the trajectory builder options used to build the loaded trajectories.

ojura: @cschuet ping

cschuet: Agreed.

PR #873, opened 2018-02-01
Move handlers to cc, use MapBuilderContextInterface

gaschler:
This is to avoid a circular dependency with LocalTrajectoryUploader and to clean up.


cschuet:
@wally-the-cartographer merge

wally-the-cartographer:

PR #883, opened 2018-02-02
Fix sphinx-build errors in evaluation.rst

MichaelGrupp:
Adds language arguments that are required for code blocks in newer Sphinx versions.

Resolves the missing code blocks in the HTML output.

### DIFF
--- a/docs/source/evaluation.rst
+++ b/docs/source/evaluation.rst
@@ -71,20 +71,20 @@ How-To
 
 Given a serialized state of a fully optimized trajectory (here: ``optimized.pbstream`` file), the ground truth relations can be generated with the following command:
 
-.. code-block::
+.. code-block:: bash
   
   cd <build>  # (directory where Cartographer's binaries are located)
   ./cartographer_autogenerate_ground_truth -pose_graph_filename optimized.pbstream -output_filename relations.pbstream -min_covered_distance 100 -outlier_threshold_meters 0.15 -outlier_threshold_radians 0.02
 
 Then, a non-optimized trajectory ``test.pbstream`` can be evaluated against the generated relations with:
 
-.. code-block::
+.. code-block:: bash
 
   ./cartographer_compute_relations_metrics -relations_filename relations.pbstream -pose_graph_filename test.pbstream
 
 This will produce output in this form:
 
-.. code-block::
+.. code-block:: bash
 
   Abs translational error 0.01944 +/- 0.01819 m
   Sqr translational error 0.00071 +/- 0.00189 m^2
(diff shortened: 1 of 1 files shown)


MichaelGrupp:
Fixes missing argument warnings, see build: https://readthedocs.org/projects/google-cartographer/builds/6674610/

docs/source/evaluation.rst
```diff
@@ -71,20 +71,20 @@ How-To
...
-.. code-block::
+.. code-block:: console
   
   cd <build>  # (directory where Cartographer's binaries are located)
   ./cartographer_autogenerate_ground_truth -pose_graph_filename optimized.pbstream -output_filename relations.pbstream -min_covered_distance 100 -outlier_threshold_meters 0.15 -outlier_threshold_radians 0.02
 
 Then, a non-optimized trajectory ``test.pbstream`` can be evaluated against the generated relations with:
 
-.. code-block::
+.. code-block:: console
```
wohe: Consistency with other Cartographer documentation: `bash`? Similar above.

docs/source/evaluation.rst
```diff
@@ -71,20 +71,20 @@ How-To
...
 
   ./cartographer_compute_relations_metrics -relations_filename relations.pbstream -pose_graph_filename test.pbstream
 
 This will produce output in this form:
 
-.. code-block::
+.. code-block:: console
```
wohe: This should be the language for syntax highlighting. Since the output isn't bash commands: `none`?

MichaelGrupp: Didn't know there's also `none`. Nevertheless, it behaves the same as `console` in my local build, so should I change it?

docs/source/evaluation.rst
MichaelGrupp: I had `bash` initially, but since we also have plain text and not only bash it looks odd in some details, see the zeros in the attached image.
screenshot from 2018-02-02 14-49-15

I can also change the commands to `bash` and leave the output as `console` - probably the best solution?

MichaelGrupp:
@wohe Changed commands to `bash` and output to `none`. Looks like the clearest solution to me.

wohe:
@MichaelGrupp thanks, this is what I suggested. What you describe sounds exactly like it is documented, see http://www.sphinx-doc.org/en/stable/markup/code.html. `none` disables syntax highlighting, `bash` is for shell scripts, `console` for shell sessions.

wohe:
@wally-the-cartographer merge

wally-the-cartographer:

PR #884, opened 2018-02-02
Use landmarks in optimization (both 2D & 3D).

pifon2a:


cartographer/mapping_3d/pose_graph/optimization_problem.cc
```diff
@@ -98,6 +100,45 @@ std::unique_ptr<transform::Rigid3d> Interpolate(
...
+    const mapping::MapById<mapping::NodeId, NodeData>& node_data,
+    mapping::MapById<mapping::NodeId, CeresPose>* C_nodes,
+    std::map<std::string, CeresPose>* C_landmarks, ceres::Problem* problem) {
+  for (const auto& landmark_node : landmark_nodes) {
+    for (const auto& observation : landmark_node.second.landmark_observations) {
+      const std::string& landmark_id = landmark_node.first;
+      // Find the trajectory nodes before and after the landmark observation.
+      auto prev =
+          node_data.lower_bound(observation.trajectory_id, observation.time);
+      if (prev == node_data.BeginOfTrajectory(observation.trajectory_id)) {
```
cschuet: BeginOfTrajectory() is a valid node. Are you sure you should skip in all cases? What if observation.time is between first and second trajectory node?

ojura:
@pifon2a you should not be accessing `landmark_nodes_` outside the worker queue in `AddLandmarkData`. The optimization problem is acessing it without taking the mutex. You are breaking the promise given in 
https://github.com/googlecartographer/cartographer/blob/b2581b0d5843441256d5364b98ac96cdf5c8d67f/cartographer/mapping_3d/pose_graph.cc#L578-L581

1) `landmark_nodes_` should also be mentioned in the comment above
2) `PoseGraph::AddLandmarkData` should, like all `PoseGraph::Add<sensordata>` functions, interact with the optimization problem's data only through the worker queue

cartographer/mapping_2d/pose_graph/optimization_problem.cc
```diff
@@ -52,6 +57,63 @@ transform::Rigid2d ToPose(const std::array<double, 3>& values) {
...
+    const mapping::MapById<mapping::NodeId, NodeData>& node_data,
+    mapping::MapById<mapping::NodeId, std::array<double, 3>>* C_nodes,
+    std::map<std::string, CeresPose>* C_landmarks, ceres::Problem* problem) {
+  for (const auto& landmark_node : landmark_nodes) {
+    for (const auto& observation : landmark_node.second.landmark_observations) {
+      const std::string& landmark_id = landmark_node.first;
+      // Find the trajectory nodes before and after the landmark observation.
+      auto next =
+          node_data.lower_bound(observation.trajectory_id, observation.time);
+      // The landmark observation was made before the trajectory was created.
```
ojura: If the landmark was observed somewhere between the first and the second node, it will be dropped here.

How about
```c++
if (observation.time < 
    node_data.BeginOfTrajectory(observation.trajectory_id)->data.time) {
  continue;
}
````
?

File: cartographer/mapping_2d/pose_graph/optimization_problem.cc [hunk 3]
```diff
@@ -52,6 +57,63 @@ transform::Rigid2d ToPose(const std::array<double, 3>& values) {
   return transform::Rigid2d({values[0], values[1]}, values[2]);
 }
 
+// Selects a trajectory node closest in time to the landmark observation and
+// applies a relative transform from it.
```
ojura: Why not interpolate here as well? Sure, it's just the starting solution so it doesn't matter as much, but since you already went to the trouble of determining the closest node, I wonder why not interpolate.

pifon2a: Oh, it will be interpolated for sure. It was just a drafty solution to make a demo.

cartographer/mapping_2d/pose_graph/optimization_problem.cc
pifon2a: Good catch, it is very unlikely though, but should be fixed. Thanks for fixing it in #901, btw.

PR #891, opened 2018-02-07
Enable loading unfrozen state

ojura:
Replaces #554.

This is the big finish to the serialization series of PRs (#549). If we managed to restore everything properly, we should be able to load the prior state unfrozen and enable full optimization, as if we built all the deserialized data in the current run and finalized the corresponding trajectories.

`LoadMap` and everything related has been relabeled to `LoadState` to indicate that it can now restore the full SLAM state, and not just the map.


ojura:
@cschuet will you be able to finish this on the gRPC side?

ojura:
@pifon2a I mentioned you in a TODO for serializing/deserializing landmarks in `MapBuilder::Write/LoadState`. The PR for that should probably similar to #666,  /s/Odometry/Landmark/.

pifon2a:
@ojura Thank you! The PR for serializing global poses of landmarks was approved yesterday. I have to add landmark observations as well. Also deserialization is not yet done.

ojura:
@cschuet can you please review to avoid this getting stale?

ojura:
Yes, that would be googlecartographer/cartographer_ros#710.

cartographer/mapping/map_builder.cc
```diff
@@ -350,24 +362,45 @@ void MapBuilder::LoadMap(io::ProtoStreamReaderInterface* const reader) {
...
+                   constraint_proto.submap_id().submap_index()});
     }
-    const NodeId node_id{
-        trajectory_remapping.at(constraint_proto.node_id().trajectory_id()),
-        constraint_proto.node_id().node_index()};
-    const SubmapId submap_id{
-        trajectory_remapping.at(constraint_proto.submap_id().trajectory_id()),
-        constraint_proto.submap_id().submap_index()};
-    pose_graph_->AddNodeToSubmap(node_id, submap_id);
+  } else {
```
ojura: @cschuet Maybe we can add a comment here that `AddSerializedConstraints` performs all the work in the branch above (when `load_frozen_state` is true), and more (actually adds all the constraints). If I remember correctly, node-to-submap information is required for 3D pure lozalization. Restoring all constraints would restore this information, but it would be wasteful, so @wohe added `AddNodeToSubmap` which adds only the node-to-submap information.

When we are doing full unfrozen state restoration, we are actually restoring all constraints, which results in adding the node-to-submap information as well, so we don't have to bother with `AddNodeToSubmap`.

ojura:
@cschuet added a comment which should clarify that part I mentioned in the comment and changed also one occurence od `LoadMap` to `LoadState` which appeared in upstream in the meantime in #919. Should be ready now.

PR #897, opened 2018-02-12
Remove dotfiles from build files in CMakeLists.txt

MichaelGrupp:
Avoids including unwanted files that could appear in a developer
workspace (e.g. auto-generated files in "dot" folders).


MichaelGrupp:
Concrete example: I noticed that when configuring the [cquery](https://github.com/cquery-project/cquery) language server for Cartographer. The corresponding vscode extension generates in a `.vscode` folder with header files that mess up the CMake build if the current brute-force glob is used.

Since most IDEs generate some kind of metadata folder in the root dir, this could affect also other developers at some point. Therefore: glob a little bit more specific.

MichaelGrupp:
Any idea why the build only fails on "Jessie"? https://travis-ci.org/googlecartographer/cartographer/builds/340585647

gaschler:
The build fails iff `BUILD_GRPC` is enabled (which is currently tested only on one build server).
`../cartographer_grpc/framework/rpc_handler_interface.h:23:27: fatal error: grpc++/grpc++.h: No such file or directory #include "grpc++/grpc++.h"`

MichaelGrupp:
Ah ok, didn't know it's only on one server. I'll try to reproduce on my machine.

MichaelGrupp:
I just ran this and it worked fine:
```bash
catkin build cartographer --force-cmake -DBUILD_GRPC=true -v --pre-clean > build.txt
grep grpc build.txt > build_grpc_grep.txt
```
The logs show that grpc was built successfully, so at least with this build command it doesn't look like CMake is the issue? You probably use another build config, or there's sth. wrong with the Jessie environment? Hard to say.

[build_grpc_grep.txt](https://github.com/googlecartographer/cartographer/files/1719652/build_grpc_grep.txt)
[build.txt](https://github.com/googlecartographer/cartographer/files/1719653/build.txt)

cschuet:
Drive by comment: I don't think the Jessie build bot has gRPC installed.

gaschler:
I correct myself, our [Jessie build server](https://github.com/googlecartographer/cartographer/blob/master/Dockerfile.jessie) should not compile or require the grpc parts. It builds the default config.

Just an idea, does `cartographer*/*.cc` recurse multiple directories?

MichaelGrupp:
> Just an idea, does cartographer*/*.cc recurse multiple directories?

It has to, otherwise subdirs like `sensor`, `mapping` etc. wouldn't be built. Also my own build would have failed.

MichaelGrupp:
> I correct myself, our Jessie build server should not compile or require the grpc parts. It builds the default config.

Note (to myself): If `-DBUILD_GRPC` is false, the gRPC dirs should be removed from the list by [this logic](https://github.com/magazino/cartographer/blob/3c5358beb03911fa550e97b50fa37c09c80a30ea/CMakeLists.txt#L62). Check.

If **all** build environments don't build gRPC, then this is still strange that it (unintentionally) builds gRPC support **only** on Jessie.

MichaelGrupp:
Just tried to hide my gRPC files in order to reproduce that:
```bash
sudo mv /usr/local/include/grpc /usr/local/include/no-grpc

# this fails (as expected)
catkin build cartographer --force-cmake -DBUILD_GRPC=true -v --pre-clean

# this works (as expected)
catkin build cartographer --force-cmake -DBUILD_GRPC=false -v --pre-clean
```

Another possibility might be that the glob command behaves differently across the CMake versions:
* Jessie: 3.0 (https://packages.debian.org/jessie/cmake) 
* Stretch: 3.7
* Xenial (like my PC): 3.5.1
* Trusty: 2.8.12

Other than that, I have no idea what could be wrong.

ojura:
Some CMake debugging tips:
- you can always do traditional printf-style debugging with message(STATUS ...)
- make a private github fork, go to www.travis-ci.com (not .org!) and start a debug build (this feature is not available for public repos). It provides you with an ssh shell to the debug build VM instance, where you can manually start a travis build and snoop around when it breaks.

MichaelGrupp:
I finally took the time today to continue working on this and came with a better approach (in my opinion). Now, all dotfiles are removed from the glob'd source files in a [separate step](https://github.com/magazino/cartographer/blob/c0329a4093371dd1316948b3395dffe0807bb4e3/CMakeLists.txt#L60) and it's clearer what's going on. More importantly, it builds now on every CI pipeline :tada: 

@gaschler PTAL

CMakeLists.txt
```diff
@@ -56,6 +56,15 @@ install(DIRECTORY cmake DESTINATION share/cartographer/)
 file(GLOB_RECURSE ALL_LIBRARY_HDRS "*.h")
 file(GLOB_RECURSE ALL_LIBRARY_SRCS "*.cc")
 file(GLOB_RECURSE ALL_TESTS "*_test.cc")
+
+# Remove dotfiles/-folders that could potentially pollute the build.
+file(GLOB_RECURSE ALL_DOTFILES ".*/*")
+if (ALL_DOTFILES)
+  list(REMOVE_ITEM ALL_LIBRARY_HDRS ${ALL_DOTFILES})
+  list(REMOVE_ITEM ALL_LIBRARY_SRCS ${ALL_DOTFILES})
+  list(REMOVE_ITEM ALL_TESTS ${ALL_DOTFILES})
```
gaschler: apply remove_item also to ALL_EXECUTABLES?

MichaelGrupp: done

gaschler:
@wally-the-cartographer merge

wally-the-cartographer:

PR #899, opened 2018-02-13
Avoid auto for Eigen expressions.

gaschler:
While harmless in most cases, auto can delay evaluation
of expressions in unexpected ways.
So it is better to avoid auto for Eigen expressions.
https://eigen.tuxfamily.org/dox/TopicPitfalls.html


gaschler:
@wally-the-cartographer merge

wally-the-cartographer:

PR #901, opened 2018-02-14
Landmark bugfixes

ojura:
- do not drop a valid observation taken at the same time as the first trajectory node
- update observations via work queue to avoid interfering with running optimization, and to preserve determinism


ojura:
Okay, I've taken out the landmark-sided observations in a separate PR.

ojura:
I think the current design could still use some improvements, though. Storing the sensor data (observations) is done within the optimization problem for other sensor types (IMU, odometry, fixed frame poses).

 Also, we could extrapolate landmark positions as soon as the observations arrive, like `PoseGraph::AddNode` does for `trajectory_nodes_`. However, notice that it's doing that outside the worker queue - `optimization_problem_` is not using `trajectory_nodes_`, so it's safe to modify it outside the worker queue (unlike `landmark_nodes_`, currently!).

PR #903, opened 2018-02-14
Make clang compile without warnings.

gaschler:
Rule out accidental usage of integer-only ::abs.
Correct static, override and protected annotations.


gaschler:
@wally-the-cartographer merge

wally-the-cartographer:

PR #905, opened 2018-02-14
Landmark-sided observations

ojura:
[RFC=0011](https://github.com/googlecartographer/rfcs/blob/master/text/0011-landmarks.md)
Enable adding landmark-sided observations of the tracking frame.

### DIFF
--- a/cartographer/mapping/2d/pose_graph/landmark_cost_function_2d.h
+++ b/cartographer/mapping/2d/pose_graph/landmark_cost_function_2d.h
@@ -89,10 +89,15 @@ class LandmarkCostFunction2D {
                          T(interpolation_parameter_));
 
     const std::array<T, 6> error = ScaleError(
-        ComputeUnscaledError(landmark_to_tracking_transform_,
-                             interpolated_pose_rotation.data(),
-                             interpolated_pose_translation.data(),
-                             landmark_rotation, landmark_translation),
+        observed_from_tracking_
+            ? ComputeUnscaledError(landmark_to_tracking_transform_,
+                                   interpolated_pose_rotation.data(),
+                                   interpolated_pose_translation.data(),
+                                   landmark_rotation, landmark_translation)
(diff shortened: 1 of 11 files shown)


ojura:
@pifon2a Okay, I moved this out of #901. This is my proposal how to do it. This makes landmarks flexible enough to replace fixed frame pose measurements, which I find lacking.

ojura:
I disallowed edits from maintainers here because Wally keeps force pushing ugly reformats (unlike how it looked before) into this branch.

cschuet:
@pifon2a Can you and @ojura figure out whether this should be merged or not?

pifon2a:
@cschuet I do not think that at this point we want to merge it. However, when (if) we migrate to the generic pose graph that will support adding interpolated nodes easier, it should be more convenient for @ojura to work with it.

ojura:
Okay. Do you plan on supporting a use case where the robot is observed by landmarks?

pifon2a:
Not really, the plan is the following: 
currently, `PoseGraph` does data conversion, calls `ConstraintBuilder` and `OptimizationProblem`. The plan is to split the current `PoseGraph` into two classes:

*  One will do data conversion only
*  Another will be a templated class that supports adding/removing edges, vertices. It will also support adding edges between
1. node & node (like for submaps & trajectory nodes)
2. pair of nodes & node (like for landmarks now)
3. pair of nodes & another pair of nodes (e.g. one robot observed another one and it happened between their trajectory nodes)
```
template<typename VertexType, typename EdgeType>
class PoseGraph {
  // some methods to add stuff, including interpolated nodes.
};
```

ojura:
That sounds like a good idea, something like that (a fully generic pose graph) was also on my mind a couple of months ago.

"pair of nodes & node (like for landmarks now)" - this sounds exactly what I need. This PR is just about being able to specify which frame the measurement was taken in, which matters very much when you have only positional measurements (i.e. you don't know the orientation of the measured frame).

PR #907, opened 2018-02-15
Introduce interfaces for metrics

gaschler:
Adds interfaces for metrics and metrics families.
Adds creation of null (noop) metrics.
Declares global register function.

### DIFF
--- a/cartographer/metrics/counter.h
+++ b/cartographer/metrics/counter.h
@@ -33,13 +33,6 @@ class Counter {
   virtual void Increment(double by_value) = 0;
 };
 
-class CounterFamily {
- public:
-  virtual ~CounterFamily() = default;
-
-  virtual Counter* Add(const std::map<std::string, std::string>& labels) = 0;
-};
-
 }  // namespace metrics
 }  // namespace cartographer
 
--- a/cartographer/metrics/counter.h
+++ b/cartographer/metrics/counter.h
@@ -1,0 +1,46 @@
+/*
+ * Copyright 2018 The Cartographer Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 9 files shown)


cartographer/metrics/counter.h
```diff
@@ -0,0 +1,46 @@
...
+ public:
+  // Counter instance that does nothing. Safe for use in static initializers.
+  static Counter* Null();
+
+  virtual ~Counter() = default;
+  virtual void Increment() = 0;
+  virtual void Increment(double by_value) = 0;
+};
+
+class CounterFamily {
```
stevewolter: At your option, it might make sense to template Family like Family<Counter>.

cartographer/metrics/gauge.h
```diff
@@ -0,0 +1,49 @@
...
+#ifndef CARTOGRAPHER_METRICS_GAUGE_H_
+#define CARTOGRAPHER_METRICS_GAUGE_H_
+
+#include <map>
+#include <vector>
+
+namespace cartographer {
+namespace metrics {
+
+class Gauge {
```
stevewolter: There was some review discussion about typing Gauge. Probably best with templates: Gauge<bool>, Gauge<int>, Gauge<double>. Should we implement this right away? On the other hand, the Prometheus implementation can't differentiate between them anyway and double should have enough precision to store bools and ints as well, so it's not terribly important.

gaschler: I would rather wait until at least one implementation handles it.

cartographer/metrics/counter.h
gaschler: Done, I made it a template.

gaschler: Please take another look.

gaschler:
@stevewolter Thanks for your review, PTAL at the new change.

gaschler:
@wally-the-cartographer merge

wally-the-cartographer:

Issue #910, opened 2018-02-16
Multiple laser scanners with per-point time are not unwarped


gaschler:
For multiple laser scans that have per-point timestamps, the following check fails:
```
[ INFO] [1518779619.525013825, 1405076296.631876676]: I0216 12:13:39.000000 31015 local_trajectory_builder.cc:103] Extrapolator is still initializing.
F0216 12:13:39.525270 31015 pose_extrapolator.cc:229] Check failed: time >= imu_tracker->time() (635406730966328436 vs. 635406730966334607) 
[FATAL] [1518779619.526051032, 1405076296.631876676]: F0216 12:13:39.000000 31015 pose_extrapolator.cc:229] Check failed: time >= imu_tracker->time() (635406730966328436 vs. 635406730966334607) 
    @           0x579513  cartographer::mapping::PoseExtrapolator::ExtrapolateRotation()
    @           0x579eab  cartographer::mapping::PoseExtrapolator::ExtrapolatePose()
    @           0x5b4427  cartographer::mapping_2d::LocalTrajectoryBuilder::AddRangeData()
    @           0x580cf0  cartographer::mapping::GlobalTrajectoryBuilder<>::AddSensorData()
    @           0x58305b  cartographer::mapping::CollatedTrajectoryBuilder::HandleCollatedSensorData()
    @           0x583181  _ZNSt17_Function_handlerIFvRKSsSt10unique_ptrIN12cartographer6sensor4DataESt14default_deleteIS5_EEEZNS3_7mapping25CollatedTrajectoryBuilderC1EPNS4_17CollatorInterfaceEiRKSt3setINSA_26TrajectoryBuilderInterface8SensorIdESt4lessISG_ESaISG_EES2_ISF_S6_ISF_EEEUlS1_S8_E_E9_M_invokeERKSt9_Any_dataS1_S8_
    @           0x620d9b  _ZNSt17_Function_handlerIFvSt10unique_ptrIN12cartographer6sensor4DataESt14default_deleteIS3_EEEZNS2_8Collator13AddTrajectoryEiRKSt13unordered_setISsSt4hashISsESt8equal_toISsESaISsEERKSt8functionIFvRKSsS6_EEEUlS6_E_E9_M_invokeERKSt9_Any_dataS6_
```

How to reproduce:
Download https://storage.googleapis.com/cartographer-public-data/bags/backpack_2d/b0-2014-07-11-10-58-16.bag.
Modify backpack_2d.lua as follows:
```
[...]

PR #911, opened 2018-02-16
WIP: Fix issue 910

gaschler:
Not ready for review.
This is a work around to allow multiple range sensors with per-point timing.
3D part is not implemented (but equivalent when we are happy with a solution).

Fixes=[910](https://github.com/googlecartographer/cartographer/issues/910)


gaschler:
Abandoned in favor of #961

PR #921, opened 2018-02-20
Instrument metrics in constraint builders.

gaschler:

### DIFF
--- a/cartographer/mapping_2d/pose_graph/constraint_builder.cc
+++ b/cartographer/mapping_2d/pose_graph/constraint_builder.cc
@@ -31,12 +31,23 @@
 #include "cartographer/common/thread_pool.h"
 #include "cartographer/mapping_2d/scan_matching/proto/ceres_scan_matcher_options.pb.h"
 #include "cartographer/mapping_2d/scan_matching/proto/fast_correlative_scan_matcher_options.pb.h"
+#include "cartographer/metrics/counter.h"
+#include "cartographer/metrics/gauge.h"
+#include "cartographer/metrics/histogram.h"
 #include "cartographer/transform/transform.h"
 #include "glog/logging.h"
 
 namespace cartographer {
 namespace mapping_2d {
 namespace pose_graph {
+
+static auto* kConstraintsSearchedMetric = metrics::Counter::Null();
+static auto* kConstraintsFoundMetric = metrics::Counter::Null();
+static auto* kGlobalConstraintsSearchedMetric = metrics::Counter::Null();
+static auto* kGlobalConstraintsFoundMetric = metrics::Counter::Null();
+static auto* kQueueLengthMetric = metrics::Gauge::Null();
+static auto* kConstraintScoresMetric = metrics::Histogram::Null();
+static auto* kGlobalConstraintScoresMetric = metrics::Histogram::Null();
 
 transform::Rigid2d ComputeSubmapPose(const Submap& submap) {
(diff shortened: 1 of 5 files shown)


cartographer/mapping_2d/pose_graph/constraint_builder.cc
```diff
@@ -287,6 +307,29 @@ void ConstraintBuilder::DeleteScanMatcher(const mapping::SubmapId& submap_id) {
...
+  kConstraintsSearchedMetric =
+      counts->Add({{"search_region", "local"}, {"matcher", "searched"}});
+  kConstraintsFoundMetric =
+      counts->Add({{"search_region", "local"}, {"matcher", "found"}});
+  kGlobalConstraintsSearchedMetric =
+      counts->Add({{"search_region", "global"}, {"matcher", "searched"}});
+  kGlobalConstraintsFoundMetric =
+      counts->Add({{"search_region", "global"}, {"matcher", "found"}});
+  auto* queues = factory->NewGaugeFamily(
+      "/mapping_2d/pose_graph/constraint_builder/queues", "Queue lengths");
```
stevewolter: .../queue_length would be more idiomatic. "queues" sounds like you're monitoring multiple queues.

cartographer/mapping_2d/pose_graph/constraint_builder.cc
```diff
@@ -246,6 +265,7 @@ void ConstraintBuilder::FinishComputation(const int computation_index) {
   std::unique_ptr<std::function<void(const Result&)>> callback;
   {
     common::MutexLocker locker(&mutex_);
+    kQueueLengthMetric->Set(constraints_.size());
```
stevewolter: constraints_ is cleared further below in this function?

cartographer/mapping_2d/pose_graph/constraint_builder.cc
```diff
@@ -70,6 +81,7 @@ void ConstraintBuilder::MaybeAddConstraint(
   if (sampler_.Pulse()) {
     common::MutexLocker locker(&mutex_);
     constraints_.emplace_back();
+    kQueueLengthMetric->Set(constraints_.size());
```
stevewolter: Queue lengths are often difficult because it assumes that ConstraintBuilder is a singleton.

Alternatively, it's often easier to have two counters (computations started and finished), and compute their difference in Prometheus to get the queue length.

cartographer/mapping_2d/pose_graph/constraint_builder.cc
gaschler: Done

cartographer/mapping_2d/pose_graph/constraint_builder.cc
gaschler: Good point. I moved the measurement so we also measure this case.

cartographer/mapping_2d/pose_graph/constraint_builder.cc
gaschler: A MapBuilder owns only one ConstraintBuilder, so the assumption is ok here (also for cloud-based mapping).
It would measure the length here directly and remember the two-counter solution for other cases (like scan matchers).

gaschler:
Thanks for your review, PTAL at the change.

gaschler:
@wally-the-cartographer merge

wally-the-cartographer:

PR #922, opened 2018-02-20
Remove 'mapping_2d' namespace.

pifon2a:
It is removed from everywhere but 'scan_matching', 'pose_graph'
subfolders of 'mapping_2d'.

[Code structure RFC](https://github.com/pifon2a/rfcs/blob/e11bca586f8cc733063ea9c94bdade71086ed9ea/text/0000-code-structure.md)

### DIFF
--- a/cartographer/io/probability_grid_points_processor.cc
+++ b/cartographer/io/probability_grid_points_processor.cc
@@ -30,7 +30,7 @@ namespace io {
 namespace {
 
 void DrawTrajectoriesIntoImage(
-    const mapping_2d::ProbabilityGrid& probability_grid,
+    const mapping::ProbabilityGrid& probability_grid,
     const Eigen::Array2i& offset,
     const std::vector<mapping::proto::Trajectory>& trajectories,
     cairo_surface_t* cairo_surface) {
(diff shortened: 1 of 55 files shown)


cartographer/io/probability_grid_points_processor.cc
```diff
@@ -148,14 +147,14 @@ std::unique_ptr<Image> DrawProbabilityGrid(
...
+mapping::ProbabilityGrid CreateProbabilityGrid(const double resolution) {
   constexpr int kInitialProbabilityGridSize = 100;
   Eigen::Vector2d max =
       0.5 * kInitialProbabilityGridSize * resolution * Eigen::Vector2d::Ones();
-  return mapping_2d::ProbabilityGrid(cartographer::mapping_2d::MapLimits(
-      resolution, max,
-      mapping_2d::CellLimits(kInitialProbabilityGridSize,
-                             kInitialProbabilityGridSize)));
+  return mapping::ProbabilityGrid(
+      mapping::MapLimits(resolution, max,
```
gaschler: No need for MapLimits2D, CellLimits2D?

cartographer/mapping_3d/submaps.h
```diff
@@ -27,7 +27,6 @@
 #include "cartographer/mapping/proto/serialization.pb.h"
 #include "cartographer/mapping/proto/submap_visualization.pb.h"
 #include "cartographer/mapping/submaps.h"
-#include "cartographer/mapping_2d/range_data_inserter.h"
```
gaschler: good catch!

cartographer/io/probability_grid_points_processor.cc
pifon2a: Nope, they are defined only for 2d.

PR #929, opened 2018-02-22
Check ifstream::good() in proto_stream.cc

MichaelGrupp:
Check for basic stream errors before doing more involved read actions.
The log message in this case is also more descriptive than
`Check failed: reader->ReadProto(&pose_graph)`.

### DIFF
--- a/cartographer/io/proto_stream.cc
+++ b/cartographer/io/proto_stream.cc
@@ -15,6 +15,7 @@
  */
 
 #include "cartographer/io/proto_stream.h"
+#include "glog/logging.h"
 
 namespace cartographer {
 namespace io {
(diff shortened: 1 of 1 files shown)


MichaelGrupp:
Discussion point: whether to check also in Writer.

I think that from a usability POV it's better to not check in the constructor, because the `close()` function reveals whether a write was successful without aborting the whole program immediately (can be frustrating if manually called with a typo, e.g. from `/write_state` in ROS).

cschuet:
@wally-the-cartographer merge

cschuet:
Thanks!

wally-the-cartographer:

PR #930, opened 2018-02-22
Fix the crash when filtering RangeData in local_trajectory_builder for very long runs.

zjwoody:
The pr fix the crash for the long runs(like car). The points in the local frame have very large position value, and it causes the crash when filtering. So I transformed the accumulated_range_data_ in the first tracking frame to avoid large position value, then filtering it.
Tested long runs.

### DIFF
--- a/cartographer/internal/mapping_3d/local_trajectory_builder.cc
+++ b/cartographer/internal/mapping_3d/local_trajectory_builder.cc
@@ -78,7 +78,8 @@ LocalTrajectoryBuilder::AddRangeData(const common::Time time,
   }
 
   if (num_accumulated_ == 0) {
-    first_pose_estimate_ = extrapolator_->ExtrapolatePose(time).cast<float>();
+    first_pose_estimate_ =
+        extrapolator_->ExtrapolatePose(time_first_point).cast<float>();
     accumulated_range_data_ =
         sensor::RangeData{Eigen::Vector3f::Zero(), {}, {}};
   }
--- a/cartographer/internal/mapping_3d/local_trajectory_builder.cc
+++ b/cartographer/internal/mapping_3d/local_trajectory_builder.cc
@@ -77,10 +77,7 @@ LocalTrajectoryBuilder::AddRangeData(const common::Time time,
     return nullptr;
   }
 
-  sensor::TimedPointCloud hits =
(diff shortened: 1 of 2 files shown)


cartographer/internal/mapping_3d/local_trajectory_builder.cc
```diff
@@ -77,37 +77,39 @@ LocalTrajectoryBuilder::AddRangeData(const common::Time time,
     return nullptr;
   }
 
-  sensor::TimedPointCloud hits =
-      sensor::VoxelFilter(0.5f * options_.voxel_filter_size())
-          .Filter(range_data.returns);
+  if (num_accumulated_ == 0) {
+    first_pose_estimate_ = extrapolator_->ExtrapolatePose(time).cast<float>();
```
cschuet: What is the first_pose_estimate_ here? Shouldn't the time for the pose of the first point in the cloud be `time + common::FromSeconds(range_data.returns.front()[3])` ? 

I am confused: I thought that `common::FromSeconds(range_data.returns.front()[3])` is smaller than zero. So it seems to me that this line is causing the `ImuTracker` to be advanced to the time of the last point in the point cloud. When you later call `ExtrapolatePose()` with `t = time + common::FromSeconds(hit[3])` that should be before the time of the last point and hence cause a crash in the `ImuTracker` in my opinion. But since you are saying that you tested this by running it I must be misunderstanding something.

zjwoody: Thanks for finding this bug, I didn't test the timed point cloud, I only use the same time for all points in same packet now. I will test the timed point cloud later.

gaschler:
I created the bug https://github.com/googlecartographer/cartographer/issues/937 and will have a look at the voxel filter itself.

gaschler:
I fixed it with https://github.com/googlecartographer/cartographer/pull/938 in the VoxelFilter itself, so there is no need to work around by transforming to the tracking frame.
@zjwoody Can you verify the fix with a long run?

zjwoody:
@gaschler , I tested #938 , it fixed the crash in long runs now, thanks!

gaschler:
@zjwoody Thanks for verifying!
I'll close this PR for now.

Issue #912, opened 2018-02-16
Check fails in PoseExtrapolator


gaschler:
```
F0216 16:48:40.710235 85212 pose_extrapolator.cc:229] Check failed: time >= imu_tracker->time() (635436203912412354 vs. 635436203912429982) 
[FATAL] [1518796120.710657648, 1408023591.249092350]: F0216 16:48:40.000000 85212 pose_extrapolator.cc:229] Check failed: time >= imu_tracker->time() (635436203912412354 vs. 635436203912429982) 
*** Check failure stack trace: ***
    @           0x579513  cartographer::mapping::PoseExtrapolator::ExtrapolateRotation()
    @           0x579eab  cartographer::mapping::PoseExtrapolator::ExtrapolatePose()
    @           0x5b4427  cartographer::mapping_2d::LocalTrajectoryBuilder::AddRangeData()
    @           0x580cf0  cartographer::mapping::GlobalTrajectoryBuilder<>::AddSensorData()
    @           0x58305b  cartographer::mapping::CollatedTrajectoryBuilder::HandleCollatedSensorData()
```

How to reproduce:

Download https://storage.googleapis.com/cartographer-public-data/bags/backpack_2d/b0-2014-08-14-13-36-48.bag

Run the following launch file:
```
<launch>
  <param name="/use_sim_time" value="true" />
  <param name="robot_description"
    textfile="$(find cartographer_ros)/urdf/backpack_2d.urdf" />
  <node name="robot_state_publisher" pkg="robot_state_publisher"
    type="robot_state_publisher" />
  <node name="cartographer_node" pkg="cartographer_ros"
      type="cartographer_node" args="
          -configuration_directory $(find cartographer_ros)/configuration_files
          -configuration_basename backpack_2d.lua"
      output="screen">
    <remap from="echoes" to="horizontal_laser_2d" />
  </node>
  <node name="cartographer_occupancy_grid_node" pkg="cartographer_ros"
      type="cartographer_occupancy_grid_node" args="-resolution 0.05" />
  <node name="rviz" pkg="rviz" type="rviz" required="true"
      args="-d $(find cartographer_ros)/configuration_files/demo_2d.rviz" />
  <node name="playbag" pkg="rosbag" type="play"
      args="--start=170 -r 3 --clock $(env HOME)/Downloads/b0-2014-08-14-13-36-48.bag" />
</launch>
```
Check fails after ~5 seconds.

gaschler:
[...]

PR #936, opened 2018-02-26
Gracefully handle time-overlapping point clouds.

gaschler:
Per-point relative times are usually computed by multiplying
a per-point time increment by the number of points.
So it is not uncommon for consecutive point clouds of a single
sensor to overlap in time.
When this happens, we act as if no backward jump in time happened
and warn once per input point cloud.

FIXES=#912


gaschler:
I verified the fix on https://storage.googleapis.com/cartographer-public-data/bags/backpack_2d/b0-2014-08-14-13-36-48.bag, cartographer_freight_demo.bag and the first 5 minutes of cartographer_paper_deutsches_museum.bag and https://storage.googleapis.com/cartographer-public-data/bags/backpack_2d/b0-2014-08-14-13-23-01.bag.

gaschler:
@wally-the-cartographer merge

wally-the-cartographer:

PR #945, opened 2018-02-28
Instrument metrics in GlobalTrajectoryBuilder.

gaschler:


cartographer/mapping/internal/global_trajectory_builder.cc
```diff
@@ -76,6 +85,14 @@ class GlobalTrajectoryBuilder : public mapping::TrajectoryBuilderInterface {
           trajectory_id_, matching_result->time, matching_result->local_pose,
           std::move(matching_result->range_data_in_local),
           std::move(insertion_result));
+
+      const std::chrono::milliseconds now =
+          std::chrono::duration_cast<std::chrono::milliseconds>(
+              std::chrono::system_clock::now().time_since_epoch());
+      common::Duration local_slam_latency_duration =
+          common::FromUniversal(now.count()) - matching_result->time;
```
cschuet: Did you test this with a real setup to determine whether ROS time to system time difference is actually what you want to measure?

gaschler: Since I cannot test this now, I will postpone this to a later PR. PTAL.

gaschler:
@wally-the-cartographer merge

PR #949, opened 2018-03-01
WIP: Allow serialization of selected trajectory IDs.

MichaelGrupp:
The optional parameter `trajectory_ids` allows to serialize only
a subset of the data. This is for example useful for re-mapping
on top of an existing map and serializing only the new map.

### DIFF
--- a/cartographer/mapping/map_builder.cc
+++ b/cartographer/mapping/map_builder.cc
@@ -168,7 +168,10 @@ std::string MapBuilder::SubmapToProto(
   return "";
 }
 
-void MapBuilder::SerializeState(io::ProtoStreamWriterInterface* const writer) {
+void MapBuilder::SerializeState(io::ProtoStreamWriterInterface* const writer,
+                                const std::set<int> trajectory_ids) {
+  const bool serialize_all_trajectories = trajectory_ids.empty();
+
   // We serialize the pose graph followed by all the data referenced in it.
   writer->WriteProto(pose_graph_->ToProto());
   // Serialize trajectory builder options.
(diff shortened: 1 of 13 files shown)


MichaelGrupp:
Follow-up for `cartographer_ros`: https://github.com/googlecartographer/cartographer_ros/pull/749

MichaelGrupp:
@cschuet This also affects the grpc interface, I just added the optional param there without functionality yet. What's your opinion?

MichaelGrupp:
CLA bump

cartographer/mapping/map_builder.cc
```diff
@@ -168,7 +168,10 @@ std::string MapBuilder::SubmapToProto(
   return "";
 }
 
-void MapBuilder::SerializeState(io::ProtoStreamWriterInterface* const writer) {
+void MapBuilder::SerializeState(io::ProtoStreamWriterInterface* const writer,
+                                const std::set<int> trajectory_ids) {
```
hauke76: Should this by any chance be a `const std::set<int>&` ?

MichaelGrupp: Yes. The params will change anyway as I used a default parameter value in the `map_builder_interface.h` in this prototype, which is not conform to the Google style guide.

MichaelGrupp:
As discussed in the open house session, this is more a simple prototype to illustrate the behavior. Some interface details like the function params need to be discussed.

If a more thorough discussion regarding this feature is needed (e.g. more selection features than just trajectory IDs), I'm also fine with opening an RFC. @gaschler

MichaelGrupp:
Closing in favor of WIP on new trimmers.

PR #951, opened 2018-03-01
Fix warning about incorrect workspace name

drigz:
The workspace name for prometheus-cpp has been updated to the correct
canonical name.


drigz:
Rebased and pushed, hope that's what you needed.

PR #960, opened 2018-03-05
Move mapping:: related mocks to cartographer/internal/testing.

pifon2a:


cartographer/mapping/internal/testing/mock_map_builder.h
```diff
@@ -26,7 +26,7 @@
 using testing::_;
 
 namespace cartographer {
-namespace cloud {
+namespace mapping {
```
gaschler: update include guards, also in other header files

PR #968, opened 2018-03-08
Move point processors to io/internal.

pifon2a:
[Internal Headers RFC](https://github.com/googlecartographer/rfcs/blob/master/text/0003-internal-headers.md)


SirVer:
I disagree with this change. Points Processors are public API - people are meant to create pipelines without lua files.

pifon2a:
@SirVer : Could you please elaborate a bit more how the PointsProcessors will be used? Probably, i'll be able to provide smth while moving these files to internal. They have the same interface except for `xray_points_processor` which has `Eigen::AlignedBox3i bounding_box() const { return bounding_box_; } `.

pifon2a:
Friendly ping.

SirVer:
@pifon2a Sorry for the slow reply, for some reasons GitHub stopped sending me mail.

This is a typical use:
```
 std::vector<std::unique_ptr<cartographer::io::PointsProcessor>> pipeline;
  // The last consumer in the pipeline must exist, so that the one created after
  // it (and being before it in the pipeline) has a valid 'next' to point to.
  // The last consumer will just drop all points.
  pipeline.emplace_back(std::make_unique<cartographer::io::NullPointsProcessor>());
  pipeline.emplace_back(std::make_unique<cartographer::io::PlyWritingPointsProcessor>(
      std::make_unique<cartographer::io::StreamFileWriter>(output_filename),
      pipeline.back().get()));
  pipeline.emplace_back(std::make_unique<cartographer::io::IntensityToColorPointsProcessor>(
      0, 255, "", pipeline.back().get()));

// then feed stuff through the pipeline.
```

Some Point Processor (like the Xray one) could have side effects that can then be queried - of course the type must then not be erased.

pifon2a:
@SirVer , thank you. Then I am closing this PR.

PR #977, opened 2018-03-10
Update descriptions in pose_graph.h

ojura:
Remove "frozen", since deserialized trajectories can be unfrozen since #891.


cschuet:
@wally-the-cartographer merge

PR #987, opened 2018-03-13
Remove dead code: fast_global_localizer.

pifon2a:
@wohe does it make sense to keep it?


wohe:
@pifon2a We do not use it. It was intended as part of public API to allow localizing without running SLAM. If it stays in `internal` it is not useful.

pifon2a:
Ok, let's remove it then.

Issue #900, opened 2018-02-14
Ceres optimization: casting scalars to Jets should be avoided when possible


ojura:
I stumbled upon [this](https://ceres-solver-review.googlesource.com/c/ceres-solver/+/9241/6) and [this](https://groups.google.com/d/msg/ceres-solver/7ZH21XX6HWU/Z3E-k2fbAwAJ) (doesn't seem to apply for quaternions, though - I posted a question over there about this).

Perhaps if we got rid of these in Cartographer, we would get a small speedup. E.g. replacing things like `T(2.0)` with `2.0`.

pifon2a:
Solved by https://github.com/googlecartographer/cartographer/pull/998.

PR #998, opened 2018-03-15
Remove float-to-jet casts where possible.

pifon2a:
Averaged time over 500K evaluations:
       before   | after this change
2d | 3766 ns | 3767 ns -> no change
3d | 3533 ns | 3467 ns -> 2% faster
@ojura: thanks for https://github.com/googlecartographer/cartographer/issues/900


ojura:
Why is CI using Ceres from `/usr/include`?

ojura:
Unfortunately, seems like the change in `acceleration_cost_function_3d.h` doesn't work with Eigen < 3.3.

PR #1002, opened 2018-03-16
Fix PureLocalizationTrimmer

gaschler:
Do not assume that submap index is sequential from 0.


pifon2a:
@wally-the-cartographer merge

PR #1007, opened 2018-03-19
Acquire mutex before copying landmark poses.

pifon2a:


pifon2a:
@wally-the-cartographer merge

PR #1011, opened 2018-03-20
Make FindGMock.cmake more robust

ojura:
Handle case with multiple invocations (avoid pasting ${CMAKE_THREAD_LIBS_INIT} multiple times).
Also check if target 'gmock' already exists before building GMock to avoid failures if GMock was already
find_packaged() from another subproject.

Want to contribute? Great! Make sure you've read, understood and considered all
the points below before creating your PR:

- Keep your PR under 200 lines of code and address a single concern.
- Add unit test(s) and documentation (these do not count toward your 200 lines).
- Adhere to the [Google C++ Style Guide](https://google.github.io/styleguide/cppguide.html).
- Run `ninja test` or `catkin_make_isolated --install --use-ninja --pkg cartographer --make-args test` as appropriate.
- Keep rebasing (or merging) of master branch to a minimum. It triggers Travis
  runs for every update which blocks merging of other changes. Our merge bot
  will rebase your branch, reformat your source code and merge as the last step
  in the review process.
- Please replace this template text with the commit message you want for your
  PR. You and/or the reviewer should keep it updated during the course of the
  review using the GitHub edit feature.

### DIFF
--- a/cmake/modules/FindGMock.cmake
+++ b/cmake/modules/FindGMock.cmake
@@ -12,56 +12,60 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-find_path(GMOCK_INCLUDE_DIRS gmock/gmock.h
-  HINTS
-    ENV GMOCK_DIR
-  PATH_SUFFIXES include
-  PATHS
-    /usr
-)
-
-# Find system-wide installed gmock.
-find_library(GMOCK_LIBRARIES
-  NAMES gmock_main
-  HINTS
-    ENV GMOCK_DIR
-  PATH_SUFFIXES lib
-  PATHS
-    /usr
-)
-
-# Find system-wide gtest header.
-find_path(GTEST_INCLUDE_DIRS gtest/gtest.h
-  HINTS
-    ENV GTEST_DIR
-  PATH_SUFFIXES include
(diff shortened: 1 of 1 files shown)


ojura:
The diff is a bit hard to read, but there are only two changes:

- everything is wrapped into an `if(NOT GMock_FOUND)`
- `add_subdirectory(gmock)` is inside an `if` which checks if the `gmock` target is defined

ojura:
@gaschler you're right, the estimate of when Catkin would get a release is a couple of weeks, we can't wait that long.

In the meantime, I believe I have a quick dirty fix available, PTAL. I just pushed it to this branch, it's the last line, the rest of the PR is just making this module more robust.
Catkin looks at the variable "GMOCK_FOUND" (all uppercase), and we can set it to false (without consequences) so Catkin doesn't enter [this branch](https://github.com/ros/catkin/blob/5f2c92b187bc3d23146418040e19e348f71af552/cmake/test/gtest.cmake#L380).

gaschler:
@ojura, thank you very much for this work-around. I know that most "obvious fixes" don't work and you found a very practical way around.

MichaelGrupp:
Thanks a lot for the patch!

Issue #1015, opened 2018-03-23
build failed with the master branch on Ubuntu 16.04 gcc-5


manifolds:
[  27%] Building CXX object CMakeFiles/cartographer.dir/cartographer/mapping/internal/2d/pose_graph/optimization_problem_2d.cc.o
In file included from /home/tong/Desktop/cartographer/cartographer/mapping/internal/pose_graph/cost_helpers.h:87:0,
                 from /home/tong/Desktop/cartographer/cartographer/mapping/internal/2d/pose_graph/landmark_cost_function_2d.h:23,
                 from /home/tong/Desktop/cartographer/cartographer/mapping/internal/2d/pose_graph/optimization_problem_2d.cc:30:
/home/tong/Desktop/cartographer/cartographer/mapping/internal/pose_graph/cost_helpers_impl.h: In instantiation of ‘std::array<FloatType, 3ul> cartographer::mapping::pose_graph::ScaleError(std::array<FloatType, 3ul>, double, double) [with T = ceres::Jet<double, 6>]’:
/home/tong/Desktop/cartographer/cartographer/mapping/internal/2d/pose_graph/spa_cost_function_2d.h:51:46:   required from ‘bool cartographer::mapping::pose_graph::SpaCostFunction2D::operator()(const T*, const T*, T*) const [with T = ceres::Jet<double, 6>]’
/usr/include/ceres/internal/variadic_evaluate.h:166:19:   required from ‘static bool ceres::internal::VariadicEvaluate<Functor, T, N0, N1, 0, 0, 0, 0, 0, 0, 0, 0>::Call(const Functor&, const T* const*, T*) [with Functor = cartographer::mapping::pose_graph::SpaCostFunction2D; T = ceres::Jet<double, 6>; int N0 = 3; int N1 = 3]’
/usr/include/ceres/internal/autodiff.h:282:72:   required from ‘static bool ceres::internal::AutoDiff<Functor, T, N0, N1, N2, N3, N4, N5, N6, N7, N8, N9>::Differentiate(const Functor&, const T* const*, int, T*, T**) [with Functor = cartographer::mapping::pose_graph::SpaCostFunction2D; T = double; int N0 = 3; int N1 = 3; int N2 = 0; int N3 = 0; int N4 = 0; int N5 = 0; int N6 = 0; int N7 = 0; int N8 = 0; int N9 = 0]’
[...]

PR #1018, opened 2018-03-26
Do not use operator*= with Jets.

pifon2a:
Should fix https://github.com/googlecartographer/cartographer/issues/1015.

Jets in Ceres do not have an overloaded `operator*=` that can work with scalars.


ojura:
Why not bump up the Ceres version?

pifon2a:
Ceres 1.13 is already used in all our Docker files and it does not solve the problem. I will file a bug to Ceres folks to add the operator. In the meantime, this fix will do.

ojura:
Ah, okay. [As you noticed](https://github.com/googlecartographer/cartographer/issues/1015#issuecomment-375585331), how did it manage to pass CI earlier when we merged #1000 and #998?

pifon2a:
Our CI use different dockerfiles (do not ask me why :) )

PR #1023, opened 2018-03-27
New fixes for Jets.

pifon2a:
Still issue #1015.


ojura:
Sad face. Still wondering how it's possible that the same version of Ceres has operators with doubles defined on one box, and not on another.

PR #1033, opened 2018-04-03
Store timestamp of the latest range data in Submap*D.

pifon2a:
This will be needed to select the submaps to be trimmed w.r.t. their freshness.


ojura:
Why not also store the earliest range data? For completeness' sake.

cschuet:
Let's consider extending the TimmingHandle to allow the Trimmer read access to constraints and trajectory nodes then you can calculate these timestamps as needed.

PR #1034, opened 2018-04-04
Allow trimmer to get nodes/constraints.

pifon2a:
We can compute timestamp for submaps using this data.


cartographer/mapping/internal/2d/pose_graph_2d.cc
```diff
@@ -798,6 +798,16 @@ std::vector<SubmapId> PoseGraph2D::TrimmingHandle::GetSubmapIds(
   return submap_ids;
 }
 
+MapById<NodeId, TrajectoryNode>
+PoseGraph2D::TrimmingHandle::GetTrajectoryNodes() {
```
gaschler: const, also below and in the interface

pifon2a:
Done.

PR #1041, opened 2018-04-06
Do not install test libraries (libgmock* / libgtest*)

mikaelarguedas:
Issue
====
Cartographer installs custom gtest and gmock libraries and headers.

I face this issue recently when packaging cartographer in ROS. cartographer currently installs gtest and gmock library in the global library directory. This makes other packages fail as they will find these custom libraries rather than the system ones.
This problem happens only on "recent" operating systems (Debian Stretch and Ubuntu Artful and Bionic).
More details available on https://github.com/ros-gbp/cartographer-release/issues/8

I'm opening this PR for feedback mostly about the approach and will remove the commented code if the maintainers decide this is the right approach.

Nature of the change:
=================
- modify the macro `_common_compile_stuff` to respect the passed `VISIBILITY` argument
- modify the macro `google_test` to make the linkage of the gmock libraries private
- modify the main CMakeLists.txt to not add the GMock includes and libraries as part of the main target.

How to test the patch:
 ===========
Requirement:
- a machine running one of the following OSs: Debian Stretch (9), Ubuntu Artful (17.10) or Ubuntu Bionic (18.04)
- the packages `python-rosdep` and `ros-lunar-catkin` installed from the ROS repositories

To save time I use ROS docker images:
`docker run -it --rm osrf/ros:lunar-desktop-full-stretch`

1. Get the code with the patch
```
mkdir -p /tmp/catkin_ws/src && cd /tmp/catkin_ws/src
git clone https://github.com/mikaelarguedas/cartographer -b dont_install_test_libs
```
2. Edit package.xml to use the ceres rosdep key
```
cd /tmp/catkin_ws/src/cartographer
sed -i 's/depend>ceres-solver</depend>libceres-dev</g' package.xml
cd /tmp/catkin_ws
```
3. Get the dependencies
```
rosdep update
rosdep install --from-paths src --ignore-src --rosdistro lunar -y
```
4. Build cartographer
```
source /opt/ros/lunar/setup.bash
catkin_make_isolated --install
```
here libgtest* and libgmock* should NOT be listed in the installation step:
```
ls install_isolated/lib/
libcartographer.a
```

5. Run the tests:
```
source install_isolated/setup.bash
catkin_make_isolated --install --make-args test
```
All tests should run and pass.


gaschler:
Thanks for providing a fix for this issue and thanks for the detailed instructions for testing.

I just build the three Docker files from https://github.com/googlecartographer/cartographer_ros locally on this change and all tests passed. (These images are based on trusty and xenial so they do not reproduce your issue, but they validate that the gmock script works for cartographer_ros, which is somewhat brittle.)

CMakeLists.txt
```diff
@@ -245,9 +245,9 @@ target_include_directories(${PROJECT_NAME} PUBLIC
 )
 
 # TODO(damonkohler): Create a testing library.
```
gaschler: Fine, you can remove the TODO and these lines.

mikaelarguedas: :+1: second commit amended accordingly

mikaelarguedas:
Testing that one more time before releasing it, it doesnt fix the issue anymore for me o_O.
@gaschler it seems that it worked for you a few hours ago so my guess is some dependency version messing with me. Do you have the steps you used to test this ? I'd like to double-check and investigate what's wrong with my current testing setup. Thanks!

As another data point, it looks like the travis run on this PR still installs the gtest libraries for Debian Stretch...

gaschler:
I investigated the issue further and I think I found the root cause.
Gmock is only a source package, so we usually build it with `add_subdirectory` called by
https://github.com/googlecartographer/cartographer/blob/master/cmake/modules/FindGMock.cmake#L57
Until version ~1.6 (see https://github.com/google/googlemock/blob/release-1.6.0/CMakeLists.txt), gmock just did `add_library` and this simple approach worked.
Unfortunately, gmock 1.7 added an unconditional `install`, which causes the unwanted installed files.
This was later changed in PR https://github.com/google/googletest/pull/1160/files, but apparently is not fixed in the stretch `googletest` package yet.

Now, I think the simplest fix is to do `add_subdirectory(... EXCLUDE_FROM_ALL)`, which I uploaded in 
https://github.com/googlecartographer/cartographer/pull/1047 together with a docker file to reproduce this easily. (I also verified that this works when called from cartographer_ros, which used to be brittle.)

A longer-term option would of course be `ExternalProject_Add`, which is actually recommended by gmock in https://github.com/google/googletest/blob/master/googletest/README.md#incorporating-into-an-existing-cmake-project.
However, I tried this a month ago and Indigo'd CMake 2.8 did not understand the dependency to build it before cartographer. Perhaps this can be worked around with the `CMakeLists.txt.in` trick (gmock readme), but it all was not convincing to me at that point.

What do you think? Do you agree with https://github.com/googlecartographer/cartographer/pull/1047 as a fix?

mikaelarguedas:
Thanks @gaschler for looking into this. I can confirm that #1047 fixes the issue in my Ububtu and Debian environments :+1: 

If you think part of this change is still valuable I can adjust the PR if necessary. Otherwise feel free to close this in favor of #1047

mikaelarguedas:
no worries, thanks for the investigation and the fix :+1:

PR #1053, opened 2018-04-11
Improve pose terminology of 2D optimization problem.

MichaelGrupp:
Removes the confusing term `initial_pose` from global 2D optimization.
It makes sense in other areas like scan matching, but has a completely
different meaning there. Also matches the 3D equivalents better.

- `local_pose_2d` replaces `initial_pose` as the name for the
   non-gravity-aligned, local 2D pose in the submap
- `global_pose_2d` replaces `pose` as the name for the
   non-gravity-aligned, global 2D pose that is optimized

See the comment in the PR for how this improves readability.


MichaelGrupp:
For example, it makes the intent of expressions like this:
```cpp
const transform::Rigid2d optimized_pose =
    optimization_problem_->submap_data().at(matching_id).global_pose *
    pose_graph::ComputeSubmapPose(*insertion_submaps.front()).inverse() *
    pose;
```
...clearer and easier to understand:
```cpp
const transform::Rigid2d global_2d_pose =
    optimization_problem_->submap_data().at(matching_id).global_pose *
    pose_graph::ComputeSubmapPose(*insertion_submaps.front()).inverse() *
    local_2d_pose;
```

pifon2a:
@MichaelGrupp : great, thanks for doing that. The naming indeed has to cleaned up. 

nit: please, make it ` local_pose_2d` and `global_pose_2d`, since we already have in `pose_graph_2d.cc`: `global_submap_pose_2d`.

pifon2a:
@wally-the-cartographer merge

wally-the-cartographer:
Merge requested by authorized user pifon2a. Merge queue now has a length of 2.

PR #1054, opened 2018-04-11
Suppress warnings for Bazel third_party deps

drigz:
These warnings occur when building with gcc 7.3.0.


drigz:
There are two remaining warnings. One in glog that can't (easily?) be suppressed, and one in prometheus-cpp (https://github.com/jupp0r/prometheus-cpp/issues/102).

cschuet:
@wally-the-cartographer merge

PR #1061, opened 2018-04-12
Update unfinished submap list on SLAM thread

### LINKED ISSUE #1052: Cloud Based Mapping Concurrency Problem for LocalSlamResult
On the Cartographer cloud side we don't have a LocalTrajectoryBuilder and therefore no ActiveSubmaps. We therefore need to emulate the behavior of ActiveSubmaps on the cloud side and that's what `ProcessLocalSlamResultData()` does. 

https://github.com/googlecartographer/cartographer/blob/master/cartographer/cloud/internal/handlers/add_local_slam_result_data_handler.cc#L34

However `ProcessLocalSlamResultData()` runs on the event handler thread while it should be running on the local SLAM thread. One way to do this would be to enqueue the whole `proto::LocalSlamResultData` and execute the logic in `ProcessLocalSlamResultData()` that updates the `unfinished_submaps_` member of `MapBuilderContext` when the item is dequeued by the `MapBuilderServer`.

cschuet:
Fixes #1052


cartographer/mapping/internal/2d/local_slam_result_2d.h
```diff
@@ -27,20 +27,23 @@ namespace mapping {
...
   void AddToTrajectoryBuilder(
       TrajectoryBuilderInterface* const trajectory_builder) override;
   void AddToPoseGraph(int trajectory_id, PoseGraph* pose_graph) const override;
 
  private:
-  std::shared_ptr<const TrajectoryNode::Data> node_data_;
-  std::vector<std::shared_ptr<const Submap2D>> insertion_submaps_;
+  const std::string sensor_id_;
+  const mapping::proto::LocalSlamResultData local_slam_result_data_;
+  cloud::SubmapController<mapping::Submap2D>* submap_controller_;
```
pifon2a: I wonder if SubmapController should belong to mapping/ and not to cloud/. WDYT?

cschuet:
@wally-the-cartographer merge

wally-the-cartographer:

PR #1074, opened 2018-04-17
Add sparse Eigen support to Bazel build.

cschuet:


cschuet:
@wally-the-cartographer merge

wally-the-cartographer:

ojura:
Is EigenSparse as fast for you as SuiteSparse? IIRC I got slow performance as soon as I forgot to install SuiteSparse, even with EigenSparse.

cschuet:
@ojura The problem with SuiteSparse is that is extremely difficult to make a hermetic Bazel build for it. That's why I put in Eigen sparse for now to get away from DENSE_QR which is super super slow.

PR #1080, opened 2018-04-17
Use area instead of cells count for overlapping trimmer options.

pifon2a:


wally-the-cartographer:
Merge requested by authorized user pifon2a. Merge queue now has a length of 3.

PR #1093, opened 2018-04-19
Ignore non-ordered data added to an ordered queue

drigz:
This stops Cartographer from crashing when it gets occasionally bad data
from one of its sources.


cschuet:
@gaschler @pifon2a Let's discuss this tomorrow. We have talked about turning some of our checks into warnings.

cschuet:
Agreed. How about we introduce a FLAG for that purpose? --strict_mode with a default to true?

cschuet:
@drigz Thanks again for starting this discussion. I am closing this PR in accordance.

PR #1094, opened 2018-04-19
Limit `bazel test` parallelism more precisely

drigz:
Since it is only the tests in cartographer/cloud that cannot run
concurrently with each other, use `tags = ["exclusive"]` to say that
they should not be run at the same time as other tests.

This works locally, not just in CI, and should speed up the CI job by
2-3 minutes.


pifon2a:
@wally-the-cartographer merge

PR #1097, opened 2018-04-20
Replace ProbabilityGrid in Submap2D by Grid2D

kdaun:
- Replace the grid type `ProbabilityGrid` in `Submap2D` by `Grid2D`
- Grid getter in `Submap2D` returns `Grid2D*` instead of `ProbabilityGrid&`
- `ComputeCroppedGrid()` becomes a virtual member function `Grid2D` 
- Move grid rendering logic from `Submap2D` to `ProbabilityGrid`
- For using the `RealTimeCorrelativeScanMatcher` `Grid2D`  is statically down-casted to `ProbabilityGrid` as for now `RealTimeCorrelativeScanMatcher` is specific to the `ProbabilityGrid` 
 - Step towards [RFC 0019](https://github.com/googlecartographer/rfcs/blob/master/text/0019-probability-grid-and-submap2d-restructuring.md)


cartographer/mapping/2d/probability_grid.cc
```diff
@@ -76,5 +78,67 @@ proto::Grid2D ProbabilityGrid::ToProto() const {
...
+  CellLimits cell_limits;
+  ComputeCroppedLimits(&offset, &cell_limits);
+  const double resolution = limits().resolution();
+  const Eigen::Vector2d max =
+      limits().max() - resolution * Eigen::Vector2d(offset.y(), offset.x());
+  std::unique_ptr<ProbabilityGrid> cropped_grid =
+      common::make_unique<ProbabilityGrid>(
+          MapLimits(resolution, max, cell_limits));
+  for (const Eigen::Array2i& xy_index : XYIndexRangeIterator(cell_limits)) {
+    if (IsKnown(xy_index + offset)) {
```
pifon2a:
```
 if (!IsKnown(xy_index + offset)) continue;
 cropped_grid->SetProbability(xy_index, GetProbability(xy_index + offset));
```

cartographer/mapping/2d/probability_grid.cc
```diff
@@ -76,5 +78,67 @@ proto::Grid2D ProbabilityGrid::ToProto() const {
...
+bool ProbabilityGrid::DrawToSubmapTexture(
+    proto::SubmapQuery::Response::SubmapTexture* const texture,
+    transform::Rigid3d local_pose) const {
+  Eigen::Array2i offset;
+  CellLimits cell_limits;
+  ComputeCroppedLimits(&offset, &cell_limits);
+
+  std::string cells;
+  for (const Eigen::Array2i& xy_index : XYIndexRangeIterator(cell_limits)) {
+    if (IsKnown(xy_index + offset)) {
```
pifon2a:
```
if (!IsKnown(xy_index + offset)) {
   cells.push_back(0 /* unknown log odds value */)); 
   cells.push_back(0 /* alpha */);                                   
    continue;
}
      // We would like to add 'delta' but this is not possible using a value and
  // alpha. We use premultiplied alpha, so when 'delta' is positive we can
  // add it by setting 'alpha' to zero. If it is negative, we set 'value' to
  // zero, and use 'alpha' to subtract. This is only correct when the pixel
  // is currently white, so walls will look too gray. This should be hard to
  // detect visually for the user, though.
  const int delta =
         128 - ProbabilityToLogOddsInteger(GetProbability(xy_index + offset));
   const uint8 alpha = delta > 0 ? 0 : -delta;
   const uint8 value = delta > 0 ? delta : 0;
   cells.push_back(value);
   cells.push_back((value || alpha) ? alpha : 1);
```

kdaun:
@wally-the-cartographer merge

PR #1098, opened 2018-04-20
ThreadPoolForTesting

gaschler:
Necessary to test ConstraintBuilder and other classes.

### DIFF
--- a/cartographer/common/internal/testing/thread_pool_for_testing.h
+++ b/cartographer/common/internal/testing/thread_pool_for_testing.h
@@ -1,0 +1,54 @@
+/*
+ * Copyright 2018 The Cartographer Authors
+ *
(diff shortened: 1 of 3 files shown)


cartographer/common/internal/testing/thread_pool_for_testing.h
```diff
@@ -0,0 +1,54 @@
...
+namespace cartographer {
+namespace common {
+namespace testing {
+
+class ThreadPoolForTesting : public ThreadPoolInterface {
+ public:
+  ThreadPoolForTesting();
+  ~ThreadPoolForTesting();
+
+  void Schedule(const std::function<void()> &work_item) override;
```
pifon2a: super-super-nit: `std::function<void()>& work_item` instead of `std::function<void()> &work_item`

PR #1069, opened 2018-04-16
Writing probability grid as proto with probability grid points processor

schwoere:
The probability grid points processor can now be configured to write the probability grid as pbstream. Before it was only possible to write the probability grid as png.

PR #1101, opened 2018-04-20
Added a fake file writer (writes to string instead of file)

schwoere:
Added fake filer writer which writes the output to a string instead of a file.

This will help writing tests for classes that use the file writer interface to write data.


wohe:
This is related to #1069.

cartographer/io/fake_file_writer.cc
```diff
@@ -0,0 +1,42 @@
+/*
+ * Copyright 2016 The Cartographer Authors
```
wohe: nit: 2018. Here and in other files?

cartographer/io/fake_file_writer.cc
```diff
@@ -0,0 +1,42 @@
...
+  return true;
+}
+
+bool FakeFileWriter::Close() { return true; }
+
+bool FakeFileWriter::WriteHeader(const char* const data, const size_t len) {
+  return Write(data, len);
+}
+
+std::string FakeFileWriter::GetFilename() { return ""; }
```
wohe: Construct using a filename and return that?

cartographer/io/fake_file_writer.cc
```diff
@@ -0,0 +1,42 @@
...
+FakeFileWriter::FakeFileWriter() {}
+
+FakeFileWriter::~FakeFileWriter() {}
+
+bool FakeFileWriter::Write(const char* const data, const size_t len) {
+  out_.append(data, len);
+  return true;
+}
+
+bool FakeFileWriter::Close() { return true; }
```
wohe: Keep track of whether `Close()` was called and `EXPECT` in the appropriate places?

cartographer/io/fake_file_writer_test.cc
```diff
@@ -0,0 +1,45 @@
...
+namespace io {
+namespace {
+
+TEST(FakeFileWriter, Write) {
+  const std::string header = "dummy header";
+  const std::vector<std::string> data_stream = {"data 1", "data 2"};
+  std::stringstream expected_result;
+
+  FakeFileWriter writer;
+  writer.WriteHeader(header.c_str(), header.size());
```
wohe: It's better to check different things in separate tests, i.e. have separate tests for `Write` and `WriteHeader`.

cartographer/io/fake_file_writer_test.cc
```diff
@@ -0,0 +1,45 @@
...
+  expected_result << header;
```
wohe: Better to directly express the expected result if possible.

cartographer/io/fake_file_writer_test.cc
```diff
@@ -0,0 +1,45 @@
...
+  ASSERT_EQ(expected_result.str(), writer.GetOutput());
+
+  for (const auto& data : data_stream) {
+    writer.Write(data.c_str(), data.size());
+    expected_result << data;
+  }
+  ASSERT_EQ(expected_result.str(), writer.GetOutput());
```
wohe: As long as the test can still continue: `EXPECT_EQ`.

cartographer/io/fake_file_writer.cc
```diff
@@ -0,0 +1,42 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "cartographer/io/fake_file_writer.h"
+
+namespace cartographer {
+namespace io {
+
+FakeFileWriter::FakeFileWriter() {}
```
pifon2a: nit: move this constructor & destructor to the header. You can use `= default`.

cartographer/io/fake_file_writer.cc
```diff
@@ -0,0 +1,33 @@
...
+
+#include "cartographer/io/fake_file_writer.h"
+
+#include <sstream>
+#include "cartographer/common/make_unique.h"
+
+namespace cartographer {
+namespace io {
+
+FakeStreamFileWriter::FakeStreamFileWriter(const std::string filename)
```
pifon2a: you can pass by reference `const std::string& filename`

cartographer/io/fake_file_writer_test.cc
```diff
@@ -0,0 +1,84 @@
...
+    EXPECT_TRUE(writer.Write(data.c_str(), data.size()));
+  }
+
+  EXPECT_EQ(writer.GetOutput(), "data 1data 2");
+  EXPECT_TRUE(writer.Close());
+}
+
+TEST(FakeFileWriter, HeaderAndWrite) {
+  const std::string header = "dummy header";
+  const std::vector<std::string> data_stream = {"data 1", "data 2"};
```
pifon2a: nit: you do not need ` = `.

cartographer/io/file_writer.cc
```diff
@@ -16,40 +16,61 @@
...
 namespace io {
 
-StreamFileWriter::StreamFileWriter(const std::string& filename)
-    : filename_(filename), out_(filename, std::ios::out | std::ios::binary) {}
+StreamWriter::StreamWriter(std::unique_ptr<std::ostream> ostream,
+                           const std::string& filename)
+    : filename_(filename), is_closed_(false), out_(std::move(ostream)) {}
 
-StreamFileWriter::~StreamFileWriter() {}
+StreamWriter::~StreamWriter() {}
```
pifon2a: put it to the header (destructor)

cartographer/io/file_writer.cc
```diff
@@ -16,40 +16,61 @@
...
 
-bool StreamFileWriter::Write(const char* const data, const size_t len) {
-  if (out_.bad()) {
+bool StreamWriter::Write(const char* const data, const size_t len) {
+  if (out_->bad() || is_closed_) {
```
pifon2a: it is allowed to write short statements in one line if they fit, i.e.:
`if (out_->bad() || is_closed_) return false;`

cartographer/io/file_writer.cc
```diff
@@ -16,40 +16,61 @@
...
-  out_.write(data, len);
-  return !out_.bad();
+  out_->write(data, len);
+  return !out_->bad();
 }
 
-bool StreamFileWriter::Close() {
-  if (out_.bad()) {
+bool StreamWriter::Close() {
+  if (out_->bad() || is_closed_) {
```
pifon2a: here as well

cartographer/io/file_writer.h
```diff
@@ -45,21 +45,31 @@ class FileWriter {
...
-  bool Write(const char* data, size_t len) override;
-  bool WriteHeader(const char* data, size_t len) override;
-  bool Close() override;
+  virtual bool Write(const char* data, size_t len) override;
+  virtual bool WriteHeader(const char* data, size_t len) override;
+  virtual bool Close() override;
   std::string GetFilename() override;
 
- private:
+ protected:
```
pifon2a: data members have to be public.
https://google.github.io/styleguide/cppguide.html#Access_Control

schwoere:
Thanks @wohe  and @pifon2a for the feedback. @pifon2a I reverted back to the initial solution which doesn't use inheritance for better readability, sorry about that. 

The std::shared_ptr<std::string> on_close_output in FakeFileWriter will be used in a mock FileWriterFactory for #1069.

Can one of you have a look again?

cartographer/io/fake_file_writer.h
```diff
@@ -0,0 +1,52 @@
...
+#include "cartographer/io/file_writer.h"
+
+namespace cartographer {
+namespace io {
+
+// Fakes a FileWriter by just writing the data to a std::string.
+class FakeFileWriter : public FileWriter {
+ public:
+  FakeFileWriter(const std::string& filename,
+                 std::shared_ptr<std::string> on_close_output =
```
pifon2a: do you really need shared_ptr? Can you use unique_ptr?

cartographer/io/fake_file_writer.h
```diff
@@ -0,0 +1,52 @@
...
+
+  bool WriteHeader(const char* data, size_t len) override;
+  bool Write(const char* data, size_t len) override;
+  bool Close() override;
+  std::string GetFilename() override;
+  std::string GetOutput() const;
+
+ private:
+  bool is_closed_;
+  std::string out_;
```
pifon2a: if you rename it to `output`, you can rename your getter `GetOutput()` to `output()`.

cartographer/io/fake_file_writer.h
```diff
@@ -0,0 +1,52 @@
...
+ public:
+  FakeFileWriter(const std::string& filename,
+                 std::shared_ptr<std::string> on_close_output =
+                     std::make_shared<std::string>());
+  ~FakeFileWriter() override = default;
+
+  bool WriteHeader(const char* data, size_t len) override;
+  bool Write(const char* data, size_t len) override;
+  bool Close() override;
+  std::string GetFilename() override;
```
pifon2a: rename `GetFilename()` to `filename()`. It is allowed for getters/setters to have names like `filename` and `set_filename`.

cartographer/io/fake_file_writer.h
pifon2a: ok, cool

cartographer/io/fake_file_writer.h
schwoere: std::string GetFilename() is defined in the FileWriter interface. This would require changes in cartographer_ros. As both naming conventions are allowed, I would like to stick with the current implementation. 
I named 
std::string GetOutput() const;
to be consistent with the GetFilename().

cartographer/io/fake_file_writer.h
```diff
@@ -0,0 +1,52 @@
...
+  std::string GetOutput() const;
```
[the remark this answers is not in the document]
schwoere: Removed it out_ member and GetOutput() and as it's currently not used/required.

cartographer/io/fake_file_writer.cc
```diff
@@ -0,0 +1,55 @@
...
+
+#include "glog/logging.h"
+
+namespace cartographer {
+namespace io {
+
+FakeFileWriter::FakeFileWriter(const std::string& filename,
+                               std::shared_ptr<std::string> content)
+    : is_closed_(false), content_(content), filename_(filename) {
+  CHECK(content);
```
wohe: Consistency in the Cartographer code base: Explicitly check for null pointers for readability, i.e. `content != nullptr`.

cartographer/io/fake_file_writer.cc
```diff
@@ -0,0 +1,55 @@
...
+}
+
+bool FakeFileWriter::Write(const char* const data, const size_t len) {
+  if (is_closed_) return false;
```
wohe: Consistency in the code base: Always use curly braces and multiple lines for if statements. Here it might be better though to `ASSERT_FALSE(is_closed_)` to catch wrong uses of the API. Similar below.

cartographer/io/fake_file_writer.cc
```diff
@@ -0,0 +1,55 @@
...
+
+bool FakeFileWriter::Close() {
+  if (is_closed_) return false;
+  is_closed_ = true;
+  return true;
+}
+
+bool FakeFileWriter::WriteHeader(const char* const data, const size_t len) {
+  if (is_closed_) return false;
+  if (content_->size() == 0 || content_->size() < len) {
```
wohe: Why the first check? Seems unnecessary.

cartographer/io/fake_file_writer.cc
```diff
@@ -0,0 +1,55 @@
...
+    *content_ = "";
+    return Write(data, len);
+  }
+  content_->replace(0, len, data);
```
wohe: This is not what we want. The first `len` characters should be replaced by `len` bytes from `data`. This code assumes data is a null terminated string. Add a test that fails with the current code, then replace? How about `std::copy(data, data + len, content_.begin());` or so?

cartographer/io/fake_file_writer_test.cc
```diff
@@ -0,0 +1,89 @@
...
+  FakeFileWriter writer("file", on_close_output);
+  EXPECT_EQ(writer.GetFilename(), "file");
+  EXPECT_TRUE(writer.Close());
+  EXPECT_EQ(*on_close_output, "");
+  EXPECT_FALSE(writer.Close());
+}
+
+TEST(FakeFileWriter, WriteHeader) {
+  auto on_close_output = std::make_shared<std::string>();
+  const std::string header("dummy header");
```
wohe: You could add a `\0` in the middle of the string here for example.

cartographer/io/fake_file_writer.cc
schwoere: Done

cartographer/io/fake_file_writer.cc
schwoere: ASSERT_FALSE doesn't work in this context, as the marcro will try to generate code that returns void which conflicts with the method signature. I left it like it was before. The other option would be to use CHECK().

cartographer/io/fake_file_writer.cc
schwoere: Right, I fixed that.

cartographer/io/fake_file_writer.cc
schwoere: You're right. I changed the content member from std::string to std::vector<char> to avoid any bugs resulting from using string methods.

cartographer/io/fake_file_writer_test.cc
schwoere: I actually can not use std::string("dummy\0 header") to construct the test cases, as the std::string internally will only store "dummy" in this case. That's also why I added test cases using std::vector<char> as an input.

schwoere:
@wohe, can you have a look again?

cartographer/io/fake_file_writer_test.cc
```diff
@@ -0,0 +1,129 @@
...
+namespace cartographer {
+namespace io {
+namespace {
+
+std::string toString(const std::vector<char>& data) {
+  return std::string(data.data(), data.size());
+}
+
+TEST(FakeFileWriter, CloseStream) {
+  auto on_close_output = std::make_shared<std::vector<char>>();
```
wohe: nit: this should be called `content` now. Similar below.

cartographer/io/fake_file_writer_test.cc
```diff
@@ -0,0 +1,129 @@
...
+  FakeFileWriter writer("file", on_close_output);
+  EXPECT_EQ("file", writer.GetFilename());
```
wohe: Extract this in a test by itself and remove everywhere else.

cartographer/io/fake_file_writer.cc
```diff
@@ -0,0 +1,55 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "cartographer/io/fake_file_writer.h"
+
+#include "glog/logging.h"
+#include "gtest/gtest.h"
```
ojura: @schwoere @wohe why are we including `gtest.h` here? This is the only place in cartographer's codebase where gtest is included in a file which is not related to testing (i.e. not a `_test.cc` or not a test helper). So this gets included in compilation when building the libcartographer target itself. And it's also incorrect - libcartographer is not linked against gmock/gtest, and does not have the include path for gtest set up. This passing CI is probably just pure coincidence, since the header was available from `/usr/include` or wherever. In an environment where gtest is not available in a standard location, this causes a build failure.

schwoere: This fake class is directly used in the fake_file_writer_test and therefore is just used for testing purposes. I updated the bazel BUILD file in this PR as well: All test helpers, mock and fake classes are now built into the cartographer_test_library. I am sending another PR for updating the CMake file now.
As a side note here: Before I updated the BUILD file, CI didn't compile because of gtest being included in fake_file_writer.cc.

PR #1133, opened 2018-05-01
Register MapBuilderServer metrics

cschuet:
I am not adding these registration calls to `cartographer/metrics/register.cc` because they are specific to cloud based mapping.

### DIFF
--- a/cartographer/cloud/map_builder_server_main.cc
+++ b/cartographer/cloud/map_builder_server_main.cc
@@ -41,7 +41,7 @@ void Run(const std::string& configuration_directory,
 #if USE_PROMETHEUS
   metrics::prometheus::FamilyFactory registry;
   ::cartographer::metrics::RegisterAllMetrics(&registry);
-  MapBuilderServerInterface::RegisterMetrics(&registry);
+  RegisterMapBuilderServerMetrics(&registry);
   ::prometheus::Exposer exposer("0.0.0.0:9100");
   exposer.RegisterCollectable(registry.GetCollectable());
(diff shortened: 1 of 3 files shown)


cartographer/cloud/map_builder_server_main.cc
```diff
@@ -41,6 +41,7 @@ void Run(const std::string& configuration_directory,
 #if USE_PROMETHEUS
   metrics::prometheus::FamilyFactory registry;
   ::cartographer::metrics::RegisterAllMetrics(&registry);
+  MapBuilderServerInterface::RegisterMetrics(&registry);
```
pifon2a: Calling a static method of an interface looks a bit confusing to me. Maybe, move `#if USE_PROMETHES` after
```
 std::unique_ptr<MapBuilderServerInterface> map_builder_server =
      CreateMapBuilderServer(map_builder_server_options,
                             std::move(map_builder));
```
and use `map_builder_server.RegisterMetrics` instead. 

Or maybe move  `MapBuilderServerInterface::RegisterMetrics` outside of the interface, but keep it in the same header. It can be renamed to 
```
namespace metrics { 
  RegisterMapBuilderServer(metrics::FamilyFactory* factory){}
}
```

cschuet: Haha, `RegisterMapBuilderServerMetrics` was what I originally had but I did not like the long function name so I moved it into the interface. I guess that's a matter of taste. Ok let's go with that.

cschuet:
@wally-the-cartographer merge

wally-the-cartographer:

PR #1142, opened 2018-05-07
Option for only local constraint search

AlexReimann:
Quite often the starting point of the vehicle is known and the _"kidnapped robot"_ problem has not to be solved. 

When running with an existing trajectory (e.g. in `pure_localization` mode) a global constraint search is performed at start. The global search and continues until a constraint is found and the local search kicks in only afterwards. 
This behavior might be undesired if the starting position of the vehicle is known and the trajectory start pose is set close to the true pose.

This PR allows to specify `global_constraint_search_after_n_seconds` to `<= 0.0` to disable all global constraint search.

### DIFF
--- a/cartographer/mapping/internal/2d/pose_graph_2d.cc
+++ b/cartographer/mapping/internal/2d/pose_graph_2d.cc
@@ -214,6 +214,7 @@ void PoseGraph2D::ComputeConstraint(const NodeId& node_id,
       data_.trajectory_connectivity_state.LastConnectionTime(
           node_id.trajectory_id, submap_id.trajectory_id);
   if (node_id.trajectory_id == submap_id.trajectory_id ||
+      options_.global_constraint_search_after_n_seconds() <= 0.0 ||
       node_time <
           last_connection_time +
               common::FromSeconds(
--- a/cartographer/mapping/internal/3d/pose_graph_3d.cc
+++ b/cartographer/mapping/internal/3d/pose_graph_3d.cc
@@ -235,6 +235,7 @@ void PoseGraph3D::ComputeConstraint(const NodeId& node_id,
       data_.trajectory_connectivity_state.LastConnectionTime(
           node_id.trajectory_id, submap_id.trajectory_id);
   if (node_id.trajectory_id == submap_id.trajectory_id ||
+      options_.global_constraint_search_after_n_seconds() <= 0.0 ||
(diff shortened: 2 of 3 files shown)


AlexReimann:
I signed it!

cschuet:
@AlexReimann Thanks. What about 2D?

wohe:
FYI. This is already possible. Just set the option to a large value, e.g. 1e9.

AlexReimann:
Even with a large value it gets triggered once at start, because of the way the timestamps get set.

wohe:
If you do not set an initial pose in the trajectory builder options, it will be triggered not just once at the start, but continuously until a match is found. This is correct, because without an initial pose there is not guaranteed connection between the two coordinate systems of the robot and the map.

If the initial pose is set, there should not be global search at the start.

We do not observe this happening. If you do, this would be a bug and needs a bug fix, not a new option IMO.

AlexReimann:
>it will be triggered not just once at the start, but continuously until a match is found.

Yeah that was what I experienced and which I wanted to avoid.

>If the initial pose is set, there should not be global search at the start.
>
> We do not observe this happening. If you do, this would be a bug and needs a bug fix, not a new option IMO.

I have the search triggering when using the `cartographer_start_trajectory` ROS node with a `.launch` file like this:
```xml
<launch>
  <arg name="x" default="0.0" />
  <arg name="y" default="0.0" />
  <arg name="yaw" default="0.0" />

  <node name="start_trajectory" pkg="cartographer_ros"
      type="cartographer_start_trajectory" args="
          -initial_pose '{to_trajectory_id = 0, relative_pose = { translation = { $(arg x), $(arg y), 0. }, rotation = { 0., 0., $(arg yaw), } } }'
          -configuration_directory $(find enway_localization)/config/generic/cartographer
          -configuration_basename localization.lua"
      output="screen">
  </node>
</launch>
```
This sets `initial_pose` so it should not be triggering?

My second way is setting the initial pose via `.lua` config file, for which I made [this patch](https://github.com/enwaytech/cartographer/commit/2620f1ac13f3086bccc94bf1bbc44dbe09d03903). There also the matching is triggered.

How do you set the initial pose of the trajectory builder options?

pifon2a:
@AlexReimann is this still needed?

AlexReimann:
No let's drop it for now.

PR #1143, opened 2018-05-07
Introduce a GlobalSlamResultCallback

cschuet:
RFC: https://github.com/googlecartographer/rfcs/blob/master/text/0020-global-slam-callback.md


ojura:
What a coincidence. I've just started working on this [like 5 minutes ago](https://user-images.githubusercontent.com/9062353/39715291-1f16ee68-522d-11e8-86bb-d1ee795d26af.png)
 :-)

cartographer/mapping/internal/2d/pose_graph_2d.cc
```diff
@@ -595,6 +605,10 @@ void PoseGraph2D::RunOptimization() {
           old_global_to_new_global * mutable_trajectory_node.global_pose;
     }
   }
+  if (global_slam_result_callback_) {
```
ojura: I think we should be invoking the callback without holding the mutex. Suppose the user wants to call `GetTrajectoryNodePoses()`, since they're interested in updating their outdated ones. If they do it inside their callback, they will get a deadlock.

cschuet: @ojura That's a great point.I hadn't thought of that. I will update PR accordingly once the RFC is merged.

cartographer/mapping/pose_graph.h
```diff
@@ -52,6 +52,9 @@ class PoseGraph : public PoseGraphInterface {
     common::Time time;
   };
 
+  using GlobalSlamResultCallback = std::function<void(
+      const std::map<int, SubmapId>&, const std::map<int, NodeId>&)>;
```
gaschler: `/* trajectory_id */`

cartographer/mapping/internal/2d/pose_graph_2d.cc
gaschler: I'd call it in `HandleWorkQueue` (and as ojura pointed out, without holding mutexes).

gaschler:
(Please link the RFC in the first comment/commit message.)

cschuet:
> @cschuet what if this was the last item in the work queue, and constraint_builder_.WhenDone will not fire again (e.g. we are shutting down)? WaitForAllComputations in the destructor will wait indefinitely until work_queue_ gets reset. I am getting this kind of deadlock, where I get stuck with an empty (not nullptr!) work_queue_.

@ojura Interesting, what do you mean by 'this was the last work item'? I don't think this commit changed the way 'WhenDone()' is called or the work items on the queue. So is this a general observation unrelated to the introduction of the global SLAM optimization callback?

ojura:
Sorry, that was purely me, I messed up multithreading. In the main thread, before destroying the map builder, I locked a mutex which I also lock in the global SLAM callback. 

`WaitForAllComputations` (which got invoked in the main thread, in the pose graph destructor) got stuck because the other Cartographer thread was trying to lock the mutex in the global SLAM callback. This appeared as an infinite `Optimizing....` message stream from Cartographer, so my first reaction was to bug you. Sorry about that :-)

cschuet:
@ojura No problem at all. Glad you were able to pinpoint the root so quickly and of course that I did not cause it :)

PR #1155, opened 2018-05-15
fix: cartographer depends on libceres-dev

Minipada:
The entry in the [rosdistro base file](https://github.com/ros/rosdistro/blob/f7a4e335f2db9827d701784cfa47ede954142f38/rosdep/base.yaml#L1501) is not `ceres-solver` but `libceres-dev`

This is needed when using rosdep and will allow detection of ceres provided by ubuntu

### DIFF
--- a/package.xml
+++ b/package.xml
@@ -37,9 +37,9 @@
   <build_depend>python-sphinx</build_depend>
 
   <depend>boost</depend>
-  <depend>ceres-solver</depend>
   <depend>eigen</depend>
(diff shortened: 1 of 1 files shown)


Minipada:
I signed it!

ojura:
Thanks for the PR. I believe this is intentional: https://github.com/googlecartographer/cartographer/pull/551

Minipada:
Thanks for the reference. I'm closing it

PR #1157, opened 2018-05-15
Add PoseUploader to stream poses to a PoseService

cschuet:


ojura:
Could throw in a description what this is for? Sounds interesting.

cartographer/cloud/internal/map_builder_server.cc
```diff
@@ -51,8 +51,11 @@ const common::Duration kPopTimeout = common::FromMilliseconds(100);
 
 MapBuilderServer::MapBuilderServer(
     const proto::MapBuilderServerOptions& map_builder_server_options,
-    std::unique_ptr<mapping::MapBuilderInterface> map_builder)
-    : map_builder_(std::move(map_builder)) {
+    MapBuilderServer::MapBuilderFactory map_builder_factory)
```
gaschler: This pattern is also necessary for cartographer_ros Node and MapBuilderBridge. Is there an alternative to specify the global SLAM callback later?

cartographer/cloud/internal/map_builder_server.h
```diff
@@ -69,12 +71,16 @@ class MapBuilderContext : public MapBuilderContextInterface {
...
+      std::function<std::unique_ptr<mapping::MapBuilderInterface>(
+          const mapping::proto::MapBuilderOptions&,
+          mapping::PoseGraph::GlobalSlamOptimizationCallback)>;
   friend MapBuilderContext<mapping::Submap2D>;
   friend MapBuilderContext<mapping::Submap3D>;
 
   MapBuilderServer(
       const proto::MapBuilderServerOptions& map_builder_server_options,
-      std::unique_ptr<mapping::MapBuilderInterface> map_builder);
+      MapBuilderFactory map_builder_factory);
```
gaschler: This change should be made in a separate commit with a follow up in cartographer_ros

cartographer/cloud/internal/pose_uploader.h
```diff
@@ -0,0 +1,49 @@
...
+#include "cartographer/common/optional.h"
+#include "cartographer/mapping/id.h"
+#include "cartographer/transform/rigid_transform.h"
+
+namespace cartographer {
+namespace cloud {
+
+class PoseUploaderInterface {
+ public:
+  virtual void EnqueueLocalPose(common::Time time, int trajectory_id,
```
gaschler: User shouldn't need to call these, only *active_trajectory should be public

cartographer/cloud/internal/pose_uploader.cc
```diff
@@ -0,0 +1,94 @@
...
+                                     const transform::Rigid3d &local_pose,
+                                     const transform::Rigid3d &global_pose) {
+  CHECK_EQ(node_id.trajectory_id, active_trajectory_id_);
+  last_global_node_id_.node_index = node_id.node_index;
+  // TODO(cschuet): Enqueue pose for uploading.
+}
+
+void PoseUploader::SetActiveTrajectory(int trajectory_id) {
+  active_trajectory_id_ = trajectory_id;
+  last_global_node_id_ = mapping::NodeId{trajectory_id, -1};
```
gaschler: use optional instead of -1

cschuet:
@gaschler Thanks for taking a look. My thinking around this feature actually changed so I will go ahead and close this pull request. I don't think this code should be living as part of Cartographer, rather it belongs into a separate binary. I will send a PR to add a gRPC method to `MapBuilderService` to subscribe to global SLAM updates (similar to what we have for local SLAM updates).

@ojura I was working on the possibility to hook up Cartographer to a system that keeps track of robotic poses.

PR #1161, opened 2018-05-22
Add analytical 2d cost function for pose graph (not used yet).

pifon2a:
Benchmark for costs after 2*10^7 evaluations with randomly pregenerated poses.

analytical 1713.24 ms
autodiff 3390.28 ms

### DIFF
--- a/cartographer/mapping/internal/optimization/cost_functions/spa_cost_function_2d.cc
+++ b/cartographer/mapping/internal/optimization/cost_functions/spa_cost_function_2d.cc
@@ -95,6 +95,8 @@ class AnalyticalSpaCostFunction2D
     const double weighted_sin_start_rotation =
         translation_weight_ * sin_start_rotation;
 
+    // Jacobians in Ceres are ordered by the parameter blocks:
+    // jacobian[i] = [(dr_0 / dx_i)^T, ..., (dr_n / dx_i)^T].
     if (jacobians[0] != NULL) {
       jacobians[0][0] = weighted_cos_start_rotation;
       jacobians[0][1] = weighted_sin_start_rotation;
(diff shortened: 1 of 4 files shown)


pifon2a:
Average over 10 pose graph optimization launches (Deutsches Museum 1. floor was loaded as an unfrozen state):             
                                            analytical             autodiff
jacobian computation:         0.324915             0.4290803333  // analytical jacobian worked ~32% faster.
total time                             2.954363667       3.247346333 
fraction of time in jac comp 0.1100644721     0.1322957539

cartographer/mapping/internal/optimization/cost_functions/spa_cost_function_2d.cc
```diff
@@ -0,0 +1,148 @@
...
+        common::NormalizeAngleDifference(
+            observed_relative_pose_.rotation().angle() - (end[2] - start[2]));
+    if (jacobians == NULL) return true;
+
+    const double weighted_cos_start_rotation =
+        translation_weight_ * cos_start_rotation;
+    const double weighted_sin_start_rotation =
+        translation_weight_ * sin_start_rotation;
+
+    if (jacobians[0] != NULL) {
```
gaschler: add comment why these are ordered like this

cartographer/mapping/internal/optimization/cost_functions/spa_cost_function_2d.cc
```diff
@@ -0,0 +1,148 @@
...
+      jacobians[0][0] = weighted_cos_start_rotation;
```
gaschler: I think the sign is wrong, in line 81 it's -sin so here it should be -cos.

pifon2a:
PTAL

ojura:
@pifon2a link is not public, is that intentional?

pifon2a:
@ojura yes, it was just a rendered version of Sphinx documentation, to make the review easier.

PR #1165, opened 2018-05-28
Defining new serialization proto

sebastianklose:
Definition of new serialization protos as defined in [RFC 0021](https://github.com/sebastianklose/rfcs/blob/serialization_format/text/0021-serialization-format.md)

The "old" proto definition is moved to an internal legacy_serialized_data.proto and still the ones being used for now.


sebastianklose:
Thanks Alex.

PR #1166, opened 2018-05-28
Mapping state serialization

sebastianklose:
Adds a method that serialized the mapping state (PoseGraph with options) using the new serialization format defined in [RFC 0021](https://github.com/sebastianklose/rfcs/blob/serialization_format/text/0021-serialization-format.md)

Function not called from `MapBuilder` yet, will hook up in a separate PR once migration tool and deserialization helper is merged as well.

### DIFF
--- a/cartographer/io/internal/mapping_state_serialization.cc
+++ b/cartographer/io/internal/mapping_state_serialization.cc
@@ -1,0 +1,171 @@
+/*
+ * Copyright 2018 The Cartographer Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "cartographer/io/internal/mapping_state_serialization.h"
+#include "cartographer/mapping/proto/serialization.pb.h"
+#include "cartographer/transform/transform.h"
+
+namespace cartographer {
+namespace io {
+
+namespace {
+
+mapping::proto::AllTrajectoryBuilderOptions
+CreateAllTrajectoryBuilderOptionsProto(
+    const std::vector<mapping::proto::TrajectoryBuilderOptionsWithSensorIds>&
+        all_options_with_sensor_ids) {
+  mapping::proto::AllTrajectoryBuilderOptions all_options_proto;
+  for (const auto& options_with_sensor_ids : all_options_with_sensor_ids) {
+    *all_options_proto.add_options_with_sensor_ids() = options_with_sensor_ids;
+  }
+  CHECK_EQ(all_options_with_sensor_ids.size(),
+           all_options_proto.options_with_sensor_ids_size());
+  return all_options_proto;
+}
+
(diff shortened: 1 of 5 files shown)


cartographer/io/internal/mapping_state_serialization.cc
```diff
@@ -0,0 +1,171 @@
...
+ * limitations under the License.
+ */
+
+#include "cartographer/io/internal/mapping_state_serialization.h"
+#include "cartographer/mapping/proto/serialization.pb.h"
+#include "cartographer/transform/transform.h"
+
+namespace cartographer {
+namespace io {
+
```
pifon2a: super-nit: remove the empty line between `namespace io {` and `namespace {`.

cartographer/io/internal/mapping_state_serialization.h
```diff
@@ -0,0 +1,40 @@
...
+ * limitations under the License.
+ */
+#ifndef CARTOGRAPHER_IO_INTERNAL_MAPPING_STATE_SERIALIZATION_H_
+#define CARTOGRAPHER_IO_INTERNAL_MAPPING_STATE_SERIALIZATION_H_
+
+#include "cartographer/io/proto_stream_interface.h"
+#include "cartographer/mapping/pose_graph.h"
+#include "cartographer/mapping/proto/trajectory_builder_options.pb.h"
+
+// Helper function to serialize a PoseGraph to a protobuf stream.
```
pifon2a: nit: it looks like this comment belongs to `ToPbStream` func.  I do not think there is a need for a file-level comment. Up to you.

cartographer/io/internal/mapping_state_serialization.h
```diff
@@ -0,0 +1,40 @@
...
+namespace cartographer {
+namespace io {
+
+// The current serialization format version.
+static constexpr int kMappingStateSerializationFormatVersion = 1;
```
pifon2a: will the version number be used outside of `mapping_state_serialization.cc`?

cartographer/io/internal/mapping_state_serialization.cc
```diff
@@ -0,0 +1,171 @@
...
+  // Next serialize all submaps.
+  for (const auto& submap_id_data : pose_graph.GetAllSubmapData()) {
+    auto* const submap_proto = proto.mutable_submap();
+    submap_proto->mutable_submap_id()->set_trajectory_id(
+        submap_id_data.id.trajectory_id);
+    submap_proto->mutable_submap_id()->set_submap_index(
+        submap_id_data.id.submap_index);
+    submap_id_data.data.submap->ToProto(submap_proto,
+                                        /*include_probability_grid_data=*/true);
+    writer->WriteProto(proto);
```
pifon2a: I would wrap these blocks into functions, that way there won't be a need for comments and also for {} blocks.

cartographer/io/internal/mapping_state_serialization.cc
sebastianklose: done.

cartographer/io/internal/mapping_state_serialization.cc
sebastianklose: done

cartographer/io/internal/mapping_state_serialization.h
sebastianklose: rm'ed

cartographer/io/internal/mapping_state_serialization.h
sebastianklose: In a different branch I am using it to verify the version in the deserialization. We can also do it differently.

pifon2a: LGTM/WLDDTM (wubba lubba dub dub to me).

PR #1167, opened 2018-05-29
Migration tool for serialization format

sebastianklose:
Adds a tool to convert from the old serialization format without header to the new serialization format with header [RFC 0021](https://github.com/sebastianklose/rfcs/blob/serialization_format/text/0021-serialization-format.md).

### DIFF
--- a/cartographer/io/serialization_format_migration.cc
+++ b/cartographer/io/serialization_format_migration.cc
@@ -121,7 +121,7 @@ mapping::proto::SerializationHeader CreateSerializationHeader() {
 void SerializeToVersion1Format(
     const ProtoMap& deserialized_data,
     cartographer::io::ProtoStreamWriterInterface* const output) {
-  static const std::vector<int> kFieldSerializationOrder = {
+  const std::vector<int> kFieldSerializationOrder = {
       SerializedData::kPoseGraphFieldNumber,
       SerializedData::kAllTrajectoryBuilderOptionsFieldNumber,
       SerializedData::kSubmapFieldNumber,
--- a/cartographer/io/serialization_format_migration.cc
+++ b/cartographer/io/serialization_format_migration.cc
@@ -33,7 +33,7 @@ using ProtoMap = std::unordered_map<int, std::vector<SerializedData>>;
 bool ReadPoseGraph(cartographer::io::ProtoStreamReaderInterface* const input,
                    ProtoMap* proto_map) {
   auto& pose_graph_vec = (*proto_map)[SerializedData::kPoseGraph];
-  pose_graph_vec.push_back({});
+  pose_graph_vec.emplace_back();
   return input->ReadProto(pose_graph_vec.back().mutable_pose_graph());
 }
 
@@ -42,7 +42,7 @@ bool ReadBuilderOptions(
     ProtoMap* proto_map) {
   auto& options_vec =
       (*proto_map)[SerializedData::kAllTrajectoryBuilderOptions];
-  options_vec.push_back({});
+  options_vec.emplace_back();
   return input->ReadProto(
       options_vec.back().mutable_all_trajectory_builder_options());
 }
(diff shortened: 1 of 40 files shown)


cartographer/io/serialization_format_migration_test.cc
```diff
@@ -0,0 +1,121 @@
...
+          std::string msg_string;
+          TextFormat::PrintToString(*proto, &msg_string);
+          this->output_messages_.push_back(msg_string);
+          return true;
+        }));
+
+    mapping::proto::PoseGraph pose_graph;
+    mapping::proto::AllTrajectoryBuilderOptions all_options;
+    mapping::proto::LegacySerializedData submap;
+    submap.mutable_submap()->mutable_submap_id()->set_trajectory_id(0);
```
gaschler: Not sure if this makes a difference here, but you are setting default values here. (I believe this line has no effect.)

cartographer/io/serialization_format_migration_test.cc
```diff
@@ -0,0 +1,121 @@
...
+
+TEST_F(MigrationTest, SerializedDataOrderIsCorrect) {
+  MigrateStreamFormatToVersion1(&reader_, writer_.get());
+  EXPECT_THAT(output_messages_, SizeIs(kNumOriginalMessages + 1));
+
+  std::vector<mapping::proto::SerializedData> serialized(
+      output_messages_.size() - 1);
+  for (size_t i = 1; i < output_messages_.size(); ++i) {
+    EXPECT_THAT(
+        TextFormat::ParseFromString(output_messages_[i], &serialized[i - 1]),
```
gaschler: Just as a note, if the messages are identical, you can compare them with google::protobuf::util::MessageDifferencer.

cartographer/io/serialization_format_migration.cc
```diff
@@ -0,0 +1,153 @@
...
+  constexpr uint32_t kVersion1 = 1;
+  mapping::proto::SerializationHeader header;
+  header.set_format_version(kVersion1);
+  return header;
+}
+
+void SerializeToVersion1Format(
+    const ProtoMap& deserialized_data,
+    cartographer::io::ProtoStreamWriterInterface* const output) {
+  static const std::vector<int> kFieldSerializationOrder = {
```
pifon2a: nit: it does not have to be static since it is used only once.

cartographer/io/serialization_format_migration.cc
```diff
@@ -0,0 +1,153 @@
...
+      SerializedData::kTrajectoryDataFieldNumber,
+      SerializedData::kImuDataFieldNumber,
+      SerializedData::kOdometryDataFieldNumber,
+      SerializedData::kFixedFramePoseDataFieldNumber,
+      SerializedData::kLandmarkDataFieldNumber};
+
+  output->WriteProto(CreateSerializationHeader());
+  for (auto field_index : kFieldSerializationOrder) {
+    const auto proto_vector_it = deserialized_data.find(field_index);
+    if (proto_vector_it == deserialized_data.end()) continue;
```
pifon2a: Is it needed to check whether `deserialized_data` has `field_index`? I would replace these three lines with:
` for (const auto& proto : deserialized_data[field_index]) {`

for range loops, `deserialized_data[]` will be evaluated only once.

cartographer/io/serialization_format_migration_test.cc
```diff
@@ -0,0 +1,122 @@
...
+    reader_.AddProto(fixed_frame_pose);
+    reader_.AddProto(trajectory_data);
+    reader_.AddProto(landmark_data);
+  }
+
+  InMemoryProtoStreamReader reader_;
+  std::unique_ptr<ForwardingProtoStreamWriter> writer_;
+  std::vector<std::string> output_messages_;
+
+  static constexpr int kNumOriginalMessages = 9;
```
pifon2a: no need for static.

cartographer/io/serialization_format_migration_test.cc
```diff
@@ -0,0 +1,122 @@
...
+};
+
+TEST_F(MigrationTest, MigrationAddsHeaderAsFirstMessage) {
+  MigrateStreamFormatToVersion1(&reader_, writer_.get());
+  // We expect one message more than the original number of messages, because of
+  // the added header.
+  EXPECT_THAT(output_messages_, SizeIs(kNumOriginalMessages + 1));
+
+  mapping::proto::SerializationHeader header;
+  EXPECT_THAT(TextFormat::ParseFromString(output_messages_[0], &header),
```
pifon2a: EXPECT_TRUE? here and everywhere

sebastianklose: I got into the habit of using the `EXPECT_THAT(...)` versions for everything for consistency. But I'm fine fine changing to `EXPECT_TRUE/FALSE`

cartographer/io/serialization_format_migration_test.cc
sebastianklose: good luck with passing this by the compiler ;-)
It's needed here.

cartographer/io/serialization_format_migration.cc
sebastianklose: done.

cartographer/io/serialization_format_migration.cc
sebastianklose: I could remove the check.
But you can only use the bracket operator on non-const maps since it would insert-if-not present. So I would vote for keeping all three lines here.

pifon2a: yeah, right, did not think about "non-constness". Then leave it like that.

cartographer/io/serialization_format_migration_test.cc
pifon2a: you can make it `const int` though ;) . Up to you.

PR #1169, opened 2018-05-29
Implement receive_global_slam_optimizations_handler

cschuet:


cartographer/cloud/internal/map_builder_context_interface.h
```diff
@@ -43,6 +44,12 @@ class MapBuilderContextInterface : public async_grpc::ExecutionContext {
   // Calling with 'nullptr' signals subscribers that the subscription has ended.
   using LocalSlamSubscriptionCallback =
       std::function<void(std::unique_ptr<LocalSlamResult>)>;
+
+  // A return value of 'false' indicates that the subscription should end.
```
gaschler: Who signals whom? In the local slam case, server signals the subscribers by calling the callback with a nullptr argument.
Why is this different here? I think the bool should be an argument.

cartographer/cloud/internal/map_builder_context_interface.h
```diff
@@ -68,6 +75,9 @@ class MapBuilderContextInterface : public async_grpc::ExecutionContext {
       int trajectory_id, LocalSlamSubscriptionCallback callback) = 0;
   virtual void UnsubscribeLocalSlamResults(
       const SubscriptionId& subscription_id) = 0;
+  virtual int SubscribeGlobalSlamOptimizations(
+      GlobalSlamOptimizationCallback callback) = 0;
+  virtual void UnsubscribeGlobalSlamOptimizations(int subscription_index) = 0;
```
gaschler: replace int by SubscriptionId, if you cannot reuse that, rename the latter to LocalSlamSubscriptionId

cartographer/cloud/internal/map_builder_context_interface.h
```diff
@@ -21,6 +21,7 @@
 #include "cartographer/cloud/internal/local_trajectory_uploader.h"
 #include "cartographer/common/blocking_queue.h"
 #include "cartographer/mapping/map_builder_interface.h"
+#include "cartographer/mapping/pose_graph.h"
```
gaschler: pose_graph_interface?

cartographer/cloud/internal/handlers/receive_global_slam_optimizations_handler.cc
```diff
@@ -0,0 +1,81 @@
...
+    const google::protobuf::Empty &request) {
+  auto writer = GetWriter();
+  const int subscription_index =
+      GetUnsynchronizedContext<MapBuilderContextInterface>()
+          ->SubscribeGlobalSlamOptimizations(
+              [writer](const std::map<int, mapping::SubmapId>
+                           &last_optimized_submap_ids,
+                       const std::map<int, mapping::NodeId>
+                           &last_optimized_node_ids) {
+                if (!writer.Write(GenerateResponse(last_optimized_submap_ids,
```
gaschler: What does the special case Write(nullptr) mean? (Add comment?)
When the given maps are empty, should that Write(nullptr)?

cartographer/cloud/internal/map_builder_context_interface.h
cschuet: As discussed made this more symmetric and added explaining comments.

cartographer/cloud/internal/map_builder_context_interface.h
cschuet: SubscriptionId is not suitable here as it contains a trajectory_id. I will rename to LocalSlamSubscriptionId.

cartographer/cloud/internal/map_builder_context_interface.h
cschuet: Done.

PR #1170, opened 2018-05-29
Add maintainers and authors to package.xml

cschuet:


cschuet:
@wally-the-cartographer merge

Issue #846, opened 2018-01-24
Optimizing: -nan%...


ojura:
I've seen this happen today. Seems like the divisor in `WaitForAllComputations` can be zero.

gaschler:
Closing as fixed by #1183

PR #1178, opened 2018-06-01
Cartographer bug 846

sotnik-github:
Fixing "Optimizing: -nan%... #846" [https://github.com/googlecartographer/cartographer/issues/846](url).
The issue was triggered in a multithreaded execution, when all nodes were processed, but the working queue was still non empty (having other kind of jobs).
Unfortunately I failed to reproduce the bug in tests, so no new tests were added.


cschuet:
@sotnik-github Hmm the CLA bot is not picking up your agreement. I think this might be caused by the email address that you used to sign the CLA being different from the one in the commit messages. Can you make sure those email addresses are the same, fix them if necessary and force push this branch again?

PR #1184, opened 2018-06-06
Skip the Collator for the landmark data.

pifon2a:
The landmark data does not need to be collated. It contains a timestamp that will be later used to find the neighboring trajectory nodes. We can directly pass the landmarks to the GlobalTrajectoryBuilder (for some reason called `wrapped_trajectory_builder` in code), which in turn will throw them to the pose graph.


pifon2a:
PTAL

pifon2a:
@wally-the-cartographer merge

ojura:
@pifon2a  Does this preserve Cartographer's determinism? Now the order of arrival of landmark observations is important, right?

cschuet:
https://github.com/googlecartographer/cartographer/issues/1220

PR #1195, opened 2018-06-14
WIP: delete trajectory, LoadFromFile

gaschler:


cartographer/mapping/internal/2d/pose_graph_2d.cc
```diff
@@ -640,6 +702,36 @@ void PoseGraph2D::RunOptimization() {
   data_.global_submap_poses_2d = submap_data;
 }
 
+bool PoseGraph2D::ShouldSkipTrajectoryWorkItem(int trajectory_id) {
```
MichaelGrupp: Renaming to `bool PoseGraph2D::CanModify(int trajectory_id)` and negating the logic would be more readable in my opinion.

cartographer/mapping/internal/2d/pose_graph_2d.cc
```diff
@@ -640,6 +702,36 @@ void PoseGraph2D::RunOptimization() {
...
+  }
+  if (it->second.state == TrajectoryState::FINISHED) {
+    // TODO(gaschler): Replace all FATAL to WARNING after some testing.
+    LOG(FATAL) << "trajectory_id " << trajectory_id
+               << " has finished "
+                  "but modification is requested, skipping.";
+    return true;
+  }
+  if (it->second.deletion_state != InternalTrajectoryState::NORMAL) {
+    LOG(FATAL) << "trajectory_id " << trajectory_id
```
MichaelGrupp: Why fatal?

MichaelGrupp: nevermind, overlooked your comment `// TODO(gaschler): Replace all FATAL to WARNING after some testing.`

cartographer/mapping/map_builder.cc
```diff
@@ -366,5 +367,18 @@ void MapBuilder::LoadState(io::ProtoStreamReaderInterface* const reader,
   CHECK(reader->eof());
 }
 
+void MapBuilder::LoadStateFromFile(const std::string& state_filename) {
+  // Check if suffix of the state file is ".pbstream".
+  const std::string suffix = ".pbstream";
+  CHECK_EQ(state_filename.substr(
```
MichaelGrupp: Looks a bit hacky, is there no more sophisticated way to check if the file format is correct?

cartographer/mapping/internal/2d/pose_graph_2d.cc
gaschler: Thanks! I named it CanAddWorkItemModifying, hope this is also fine.

cartographer/mapping/map_builder.cc
gaschler: I just copied this function from cartographer_ros. On a second thought, it should only warn if the suffix differs because checks should only fail if the algorithm won't work. They should fail if the contents cannot be loaded, and this LoadState will do.

gaschler:
@MichaelGrupp Thanks for your review, I'm uploading changes.

cartographer/mapping/internal/2d/pose_graph_2d.cc
MichaelGrupp: Looks better. Don't know if this method could be useful for anything else than adding work items though.

PR #1202, opened 2018-06-19
LoadStateFromFile

gaschler:
[RFC=0023](https://github.com/googlecartographer/rfcs/blob/master/text/0023-delete-load.md)

### DIFF
--- a/cartographer/mapping/map_builder_interface.h
+++ b/cartographer/mapping/map_builder_interface.h
@@ -80,9 +80,15 @@ class MapBuilderInterface {
   // Serializes the current state to a proto stream.
   virtual void SerializeState(io::ProtoStreamWriterInterface* writer) = 0;
 
-  // Loads the SLAM state from a proto stream.
-  virtual void LoadState(io::ProtoStreamReaderInterface* reader,
-                         bool load_frozen_state) = 0;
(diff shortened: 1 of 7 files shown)


cartographer/mapping/map_builder_interface.h
```diff
@@ -80,9 +80,15 @@ class MapBuilderInterface {
...
-                         bool load_frozen_state) = 0;
+  // Loads the SLAM state from a proto stream. Returns the remapping of new
+  // trajectory_ids.
+  virtual std::map<int /* trajectory id in proto */, int /* trajectory id */>
+  LoadState(io::ProtoStreamReaderInterface* reader, bool load_frozen_state) = 0;
+
+  // Loads the SLAM state froma a pbstream file. Returns the remapping of new
+  // trajectory_ids.
+  virtual std::map<int /* trajectory id in proto */, int /* trajectory id */>
+  LoadStateFromFile(const std::string& filename) = 0;
```
pifon2a: I would just call both of them `LoadState`. Up to you.

gaschler: As discussed, I'd leave it as in the RFC.
Thanks for your review!

PR #1216, opened 2018-06-27
Separate raycast and grid update

kdaun:
Separate raycast and grid update logic.
Rename raycast to ray_to_pixel_mask.


cartographer/mapping/internal/2d/ray_casting.h
```diff
@@ -19,21 +19,16 @@
...
 
-// For each ray in 'range_data', inserts hits and misses into
-// 'probability_grid'. Hits are handled before misses.
-void CastRays(const sensor::RangeData& range_data,
-              const std::vector<uint16>& hit_table,
-              const std::vector<uint16>& miss_table, bool insert_free_space,
-              ProbabilityGrid* probability_grid);
+// Compute all pixels in which some part of the line segment connecting 'begin'
+// and 'end' lies with 'subpixel_scale' precision.
+std::vector<Eigen::Array2i> RayToPixelMask(const Eigen::Array2i& begin,
```
gaschler: rename to LineToPixelMask/Set or just RenderLine?
(after discussion has concluded, rename this file and the include guards accordingly)

cartographer/mapping/internal/2d/ray_casting.h
```diff
@@ -19,21 +19,16 @@
...
+                                           const Eigen::Array2i& end,
+                                           int subpixel_scale);
```
gaschler: subpixel_scale is misleading here as the returned result is only pixels, I wouldn't expose it as a parameter unless your new feature requires this

cartographer/mapping/internal/2d/ray_casting.h
```diff
@@ -19,21 +19,16 @@
...
 namespace mapping {
 
-// For each ray in 'range_data', inserts hits and misses into
-// 'probability_grid'. Hits are handled before misses.
-void CastRays(const sensor::RangeData& range_data,
-              const std::vector<uint16>& hit_table,
-              const std::vector<uint16>& miss_table, bool insert_free_space,
-              ProbabilityGrid* probability_grid);
+// Compute all pixels in which some part of the line segment connecting 'begin'
+// and 'end' lies with 'subpixel_scale' precision.
```
gaschler: Comment that begin and end coordinates must be non-zero or drop those checks

cartographer/mapping/2d/probability_grid_range_data_inserter_2d.cc
```diff
@@ -58,10 +79,55 @@ void ProbabilityGridRangeDataInserter2D::Insert(
...
+
+  if (!options_.insert_free_space()) {
+    return;
+  }
+
+  // Now add the misses.
+  for (const Eigen::Array2i& end : ends) {
+    std::vector<Eigen::Array2i> ray =
+        RayToPixelMask(begin, end, kSubpixelScale);
+    for (const auto& cell_index : ray) {
```
gaschler: auto may behave unexpectedly with expression templates like Eigen types, write Eigen::Array2i

cartographer/mapping/2d/probability_grid_range_data_inserter_2d.h
```diff
@@ -51,6 +51,10 @@ class ProbabilityGridRangeDataInserter2D : public RangeDataInserterInterface {
                       GridInterface* grid) const override;
 
  private:
+  // For each ray in 'range_data', inserts hits and misses into
+  // 'probability_grid'. Hits are handled before misses.
+  void CastRays(const sensor::RangeData& range_data,
```
gaschler: can this be static?

cartographer/mapping/internal/2d/ray_casting.cc
```diff
@@ -18,79 +18,75 @@
...
   // The full pixel x coordinate of 'end'.
-  const int end_x = std::max(begin.x(), end.x()) / kSubpixelScale;
+  const int end_x = std::max(begin.x(), end.x()) / subpixel_scale;
 
   // Move from 'begin' to the next pixel border to the right.
   sub_y += dy * first_pixel;
   if (dy > 0) {
     while (true) {
-      probability_grid->ApplyLookupTable(current, miss_table);
+      if (!ray.back().isApprox(current)) ray.push_back(current);
```
gaschler: is it guaranteed that the same pixel can only be at ray.back()?
Don't use isApprox() to compare integer vectors

cartographer/mapping/internal/2d/ray_casting.cc
```diff
@@ -18,79 +18,75 @@
...
-    CastRay(end, begin, miss_table, probability_grid);
-    return;
+    return RayToPixelMask(end, begin, subpixel_scale);
   }
 
   CHECK_GE(begin.x(), 0);
   CHECK_GE(begin.y(), 0);
   CHECK_GE(end.y(), 0);
-
+  std::vector<Eigen::Array2i> ray;
```
gaschler: rename

cartographer/mapping/2d/probability_grid_range_data_inserter_2d.h
kdaun: It depends on nonstatic members, therefore no.

pifon2a: I would hide it in .cc file in anonymous namespace. You will have to pass more parameters `hit_table_`, `miss_table_`, `insert_free_space`.

cartographer/mapping/internal/2d/ray_casting.cc
kdaun: > is it guaranteed that the same pixel can only be at ray.back()?
Yes.

cartographer/mapping/internal/2d/ray_casting.h
kdaun: begin and end are expected to be scaled by subpixel scale, that should be indicated more clearly. As the parameter allows a tradeoff between speed and accuracy, I don't believe forcing a fixed value for all potential use-cases is a good idea.

cartographer/mapping/internal/2d/ray_to_pixel_mask.h
```diff
@@ -24,13 +24,15 @@
...
 namespace mapping {
 
-// Compute all pixels in which some part of the line segment connecting 'begin'
-// and 'end' lies with 'subpixel_scale' precision.
-std::vector<Eigen::Array2i> RayToPixelMask(const Eigen::Array2i& begin,
-                                           const Eigen::Array2i& end,
+// Compute all pixels in which some part of the line segment connecting
+// 'scaled_begin' and 'scaled_end'. 'scaled_begin' and 'scaled_end' are scaled
+// by 'subpixel_scale'. 'scaled_begin' and 'scaled_end' are expected to be
+// greater than zero.
```
gaschler: Comment that return values are in pixels and not scaled.

cartographer/mapping/internal/2d/ray_casting.h
gaschler: I'd rename the parameters to "begin_in_subpixels",..., "subpixel_per_pixel" to make it clear. (This was not clear to me during the previous review.) "scaled by" in my eyes has two potential meanings.

cartographer/mapping/internal/2d/ray_to_pixel_mask.cc
```diff
@@ -0,0 +1,159 @@
...
+
+#include "cartographer/mapping/internal/2d/ray_to_pixel_mask.h"
+
+#include "Eigen/Dense"
+
+namespace cartographer {
+namespace mapping {
+namespace {
+
+bool isEqual(Eigen::Array2i a, Eigen::Array2i b) {
```
gaschler: const ref

cartographer/mapping/internal/2d/ray_to_pixel_mask.h
```diff
@@ -14,28 +14,25 @@
...
 namespace cartographer {
 namespace mapping {
 
-// For each ray in 'range_data', inserts hits and misses into
-// 'probability_grid'. Hits are handled before misses.
-void CastRays(const sensor::RangeData& range_data,
-              const std::vector<uint16>& hit_table,
-              const std::vector<uint16>& miss_table, bool insert_free_space,
-              ProbabilityGrid* probability_grid);
+// Compute all pixels in which some part of the line segment connecting
```
gaschler: replace "in which" by "that contain"

cartographer/mapping/internal/2d/ray_to_pixel_mask_test.cc
```diff
@@ -65,7 +65,7 @@ TEST(RayCastingTest, AxisAlignedY) {
   EXPECT_TRUE(ray[2].isApprox(Eigen::Array2i({1, 3})));
```
gaschler: You are checking for the exact sequence, which is strict. It would be slightly better to use
https://github.com/google/googletest/blob/master/googlemock/docs/CheatSheet.md#container-matchers
that ignore the ordering. For this, you need to define a new matcher for Array2i similar to what's done in cartographer/sensor/compressed_point_cloud_test.cc:43

cartographer/mapping/internal/2d/ray_to_pixel_mask_test.cc
```diff
@@ -116,7 +116,7 @@ TEST(RayCastingTest, SteepLine) {
   EXPECT_TRUE(ray[3].isApprox(Eigen::Array2i({2, 4})));
 }
 
-TEST(RayCastingTest, FlatLine) {
+TEST(RayToPixelMaskTest, FlatLine) {
   Eigen::Array2i begin = {1, 1};
```
gaschler: maybe swap begin and end in one test to cover the other code path?

cartographer/mapping/internal/2d/ray_to_pixel_mask_test.cc
```diff
@@ -0,0 +1,221 @@
...
+      CellLimits(num_cells_x * subpixel_scale, num_cells_y * subpixel_scale));
+  Eigen::Array2i begin =
+      superscaled_limits.GetCellIndex(Eigen::Vector2f({0.01, 0.09}));
+  Eigen::Array2i end =
+      superscaled_limits.GetCellIndex(Eigen::Vector2f({0.21, 0.19}));
+  std::vector<Eigen::Array2i> ray = RayToPixelMask(begin, end, subpixel_scale);
+  EXPECT_EQ(ray.size(), 4);
+  EXPECT_TRUE(ray[0].isApprox(Eigen::Array2i({8, 7})));
+  EXPECT_TRUE(ray[1].isApprox(Eigen::Array2i({8, 8})));
+  // expected wrong pixel as subpixel_scale is too low
```
gaschler: style: First letter capital, end with full stop.

cartographer/mapping/internal/2d/ray_to_pixel_mask_test.cc
```diff
@@ -0,0 +1,221 @@
...
+      superscaled_resolution, max,
+      CellLimits(num_cells_x * subpixel_scale, num_cells_y * subpixel_scale));
+  Eigen::Array2i begin =
+      superscaled_limits.GetCellIndex(Eigen::Vector2f({0.05, 0.05}));
+  Eigen::Array2i end =
+      superscaled_limits.GetCellIndex(Eigen::Vector2f({0.35, 0.05}));
+  const std::vector<Eigen::Array2i> base_scale_ray =
+      RayToPixelMask(begin, end, subpixel_scale);
+  EXPECT_EQ(base_scale_ray.size(), 4);
+  EXPECT_TRUE(base_scale_ray[0].isApprox(Eigen::Array2i({9, 6})));
```
gaschler: Couldn't this be in the loop when you start the loop with scale 1?

cartographer/mapping/internal/2d/ray_to_pixel_mask_test.cc
kdaun: The other code path is already tested in multiple other tests. So I don't think it is necessary here again.

cartographer/mapping/internal/2d/ray_to_pixel_mask.cc
kdaun: Yes!

cartographer/mapping/internal/2d/ray_to_pixel_mask_test.cc
kdaun: Yes.

wally-the-cartographer:
Merge requested by authorized user kdaun. Merge queue now has a length of 1.

PR #1223, opened 2018-07-02
Avoid uninitialized NodeId, SubmapId

gaschler:
Previously, NodeId and SubmapId could be (partially) uninitialized,
for instance like this:

```
NodeId node_id;
SubmapId submap{0}; // uninitialized submap_index
```

This introduces constructors to prevent this.

### DIFF
--- a/cartographer/mapping/internal/constraints/constraint_builder_2d_test.cc
+++ b/cartographer/mapping/internal/constraints/constraint_builder_2d_test.cc
@@ -79,12 +79,12 @@ TEST_F(ConstraintBuilder2DTest, FindsConstraints) {
   for (int i = 0; i < 2; ++i) {
     EXPECT_EQ(constraint_builder_->GetNumFinishedNodes(), expected_nodes);
     for (int j = 0; j < 2; ++j) {
-      constraint_builder_->MaybeAddConstraint(submap_id, &submap, NodeId{},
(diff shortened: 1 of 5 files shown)


gaschler:
@wally-the-cartographer merge

wally-the-cartographer:

cartographer/mapping/internal/constraints/constraint_builder_2d_test.cc
```diff
@@ -79,12 +79,12 @@ TEST_F(ConstraintBuilder2DTest, FindsConstraints) {
   for (int i = 0; i < 2; ++i) {
     EXPECT_EQ(constraint_builder_->GetNumFinishedNodes(), expected_nodes);
     for (int j = 0; j < 2; ++j) {
-      constraint_builder_->MaybeAddConstraint(submap_id, &submap, NodeId{},
+      constraint_builder_->MaybeAddConstraint(submap_id, &submap, NodeId{0, 0},
```
pifon2a: Maybe just specify defaults in `NodeId` struct directly?

gaschler: In my eyes, a Node/SubmapId should never be default-initialized (outside of tests).
If you don't set it, you shouldn't create an id.

pifon2a: Then maybe delete the default constructor?

PR #1225, opened 2018-07-02
Fully delete trajectory

gaschler:
Previously, constraints across trajectories (and some nodes)
were not deleted because the trimmer interface was used,
which resulted in an inconsistent pose graph and a MapById
out-of-range exception.
Deleting cross-trajectory constraints is now covered by an integration
test.

### DIFF
--- a/cartographer/mapping/internal/2d/pose_graph_2d.cc
+++ b/cartographer/mapping/internal/2d/pose_graph_2d.cc
@@ -359,20 +359,45 @@ void PoseGraph2D::UpdateTrajectoryConnectivity(const Constraint& constraint) {
 }
 
 void PoseGraph2D::DeleteTrajectoriesIfNeeded() {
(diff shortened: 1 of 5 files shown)


cschuet:
@gaschler Thanks for taking this on. Does this mean trimmers are not deleting all the things they ought to be deleting? So is there not a fix needed for the TrimSubmap() function?

E.g. what I don't  understand from looking over the code is this: How are INTER_SUBMAP constraints that go from a submap in the background trajectory to a node in the pure localization trajectory ever deleted and why is that not creating a problem already?

cschuet:
@gaschler ok because all constraints related to orphaned nodes are deleted here

https://github.com/googlecartographer/cartographer/blob/7839f3f21608909ae1415df5770af15fae2f7961/cartographer/mapping/internal/2d/pose_graph_2d.cc#L1039

PR #1231, opened 2018-07-03
Fix the local SLAM latency metric.

wohe:
Before, the metric was in integer seconds and since measurements
are usually quite small, it was mostly outputting 0.


wohe:
FYI @CodeArno @spielawa

wohe:
@wally-the-cartographer merge

wally-the-cartographer:

PR #1239, opened 2018-07-05
fixed casting of high_resolution_max_range (SubmapsOptions3D)

schwoere:
Fixed unintentional casting of high_resolution_max_range from double to int, to float

In SubmapsOptions3D the parameter "high_resolution_max_range" is defined as double. In the code it gets casted to int when calling Submap3D::InsertRangeData and to float when calling FilterRangeDataByMaxRange.

### DIFF
--- a/cartographer/mapping/3d/submap_3d.cc
+++ b/cartographer/mapping/3d/submap_3d.cc
@@ -39,7 +39,7 @@ struct PixelData {
 // 'max_range' distance from the origin. Removes misses and reflectivity
 // information.
 sensor::RangeData FilterRangeDataByMaxRange(const sensor::RangeData& range_data,
-                                            const double max_range) {
+                                            const float max_range) {
   sensor::RangeData result{range_data.origin, {}, {}};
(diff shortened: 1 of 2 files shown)


cartographer/mapping/3d/submap_3d.cc
```diff
@@ -261,7 +261,7 @@ void Submap3D::ToResponseProto(
 
 void Submap3D::InsertRangeData(const sensor::RangeData& range_data,
                                const RangeDataInserter3D& range_data_inserter,
-                               const int high_resolution_max_range) {
+                               const double high_resolution_max_range) {
```
wohe: Let's change these to `float`. It is quite common that options are stored as `double`, since there is no need for float support in Lua options. The data this gets compared against is `float`, so keeping it as `double` does not provide any benefit.

wohe:
FYI @SirVer

wohe:
@wally-the-cartographer merge

wally-the-cartographer:
Merge requested by authorized user wohe. Merge queue now has a length of 4.

wally-the-cartographer:

PR #1243, opened 2018-07-06
Removed redundant 2d voxel filtering

schwoere:
Removed redundant adaptive voxel fitering in trajectory builder 2d.

Adaptive voxel filtering of the lidar point cloud was performed in InsertIntoSubmap and ScanMatch. Both methods are called from AddAccumulatedRangeData. Now, adaptive voxel filtering is done only once in AddAccumulatedRangeData and the filtered point cloud is then forwarded to InsertIntoSubmap and ScanMatch.


cartographer/mapping/internal/2d/local_trajectory_builder_2d.cc
```diff
@@ -206,6 +203,13 @@ LocalTrajectoryBuilder2D::AddRangeData(
   return nullptr;
 }
 
+sensor::PointCloud LocalTrajectoryBuilder2D::RunAdaptiveVoxelFilter(
```
wohe: Extracting this one liner into a function does not improve readability. Inline below? If you wanted to, you could refactor the `AdaptiveVoxelFilter` which only keeps the option around and seems to be only ever used for a single call. This would be a follow up PR though: Make the function visible in the voxel filter header and remove the `AdaptiveVoxelFilter` class.

schwoere:
PTAL @wohe

schwoere:
@wally-the-cartographer merge

wally-the-cartographer:
Merge requested by authorized user schwoere. Merge queue now has a length of 2.

PR #1247, opened 2018-07-09
Fix the name of real_time_correlative metric in 2D.

wohe:
This was called fast_correlative before, but the metric was about
the score of the real-time correlative scan matcher in local SLAM.


cartographer/mapping/internal/2d/local_trajectory_builder_2d.cc
```diff
@@ -82,7 +82,7 @@ std::unique_ptr<transform::Rigid2d> LocalTrajectoryBuilder2D::ScanMatch(
         pose_prediction, filtered_gravity_aligned_point_cloud,
```
schwoere: please in line 81. Add const:
`const double score =`

wohe: Done.

wohe:
@wally-the-cartographer merge

wally-the-cartographer:

PR #1249, opened 2018-07-10
Add .clang-format file.

wohe:
This adds a .clang-format file, so that git clang-format uses
Google style without the need to remember the commandline flag.

Moreover, this ensures correct formatting if the Cartographer
code is in a subtree of a repo.


wohe:
@cschuet If you think this is a good idea, we could also add this to the other repos as appropriate. Maybe it is even possible to remove the command-line flag from wally. @SirVer What do you think?

cartographer/.clang-format
```diff
@@ -0,0 +1 @@
+BasedOnStyle: Google
```
cschuet: Should we also add
```
DerivePointerAlignment: false
PointerAlignment: Left
```
Otherwise the first pointer alignment (int* a vs int *a) determines the formatting for the rest of the file?

wohe: I can add this. Right now I just mirrored what wally does, i.e. set style to Google.

gaschler: Cool, please specify `PointerAlignment: Left` as we do in 95% of the code base

wohe: I'll add `DerivePointerAlignment: false` which is the important part here. `Left` is already set in Google style by default. This is now similar to what tensorflow does. If you want me to explicitly set `PointerAlignment` just let me know.

wohe:
@CodeArno @cschuet @gaschler PTAL

cartographer/.clang-format
CodeArno: Makes sense, I think this should do what we expect it to.

wally-the-cartographer:

PR #946, opened 2018-02-28
Instrument metrics in local trajectory builders.

gaschler:
RFC=[0014](https://github.com/googlecartographer/rfcs/blob/master/text/0014-monitoring.md)

PR #1251, opened 2018-07-10
Measure accumulation_duration from last accumulation stop

spielawa:
[PR 946](https://github.com/googlecartographer/cartographer/pull/946) introduced metrics in local trajectory builder. The accumulation duration was measured from accumulation_start to accumulation_stop. 

This PR changes this to measure the entire time elapsed between two accumulations (i.e. from accumulation_stop to the next accumulation_stop). 

Mostly the two measurements result in roughly the same, with the new way measuring slightly larger durations, as expected. But occasionally the measurements differ significantly. This is probably due to a lock contention somewhere outside of what was measured previously. Since we are interested in real time metrics, we need to track the entire time passed.


wohe:
@wally-the-cartographer merge

wally-the-cartographer:

PR #1253, opened 2018-07-10
Create first submap with first scan

schwoere:
The first submap is now created with first call to `ActiveSubmaps:InsertRangeData`.

Originally the first submap was created at the origin. Creating the first submap with the first range data insertion allows to align the first submap with the gravity estimate of the first scan. This change makes the interface of ActiveSubmaps most consistent, as `ActiveSubmaps::submaps()` will return the correct active submaps after insertion.

The change affects the result of cartographer. The results were verified and provide the same quality as current master.


cartographer/mapping/2d/submap_2d.cc
```diff
@@ -164,17 +164,12 @@ std::unique_ptr<GridInterface> ActiveSubmaps2D::CreateGrid(
...
 
-void ActiveSubmaps2D::FinishSubmap() {
-  Submap2D* submap = submaps_.front().get();
-  submap->Finish();
-  submaps_.erase(submaps_.begin());
-}
-
 void ActiveSubmaps2D::AddSubmap(const Eigen::Vector2f& origin) {
-  if (submaps_.size() > 1) {
+  if (submaps_.size() == 2) {
```
CodeArno: Nit: Consider >=2 to be on the safe side for possible future extensions (also below in 3D).

cartographer/mapping/2d/submap_2d.cc
```diff
@@ -127,22 +127,22 @@ void Submap2D::Finish() {
...
-  // and will create it at the origin in absence of a better choice.
-  AddSubmap(Eigen::Vector2f::Zero());
-}
+    : options_(options), range_data_inserter_(CreateRangeDataInserter()) {}
 
 std::vector<std::shared_ptr<Submap2D>> ActiveSubmaps2D::submaps() const {
   return submaps_;
 }
 
 void ActiveSubmaps2D::InsertRangeData(const sensor::RangeData& range_data) {
```
CodeArno: The general API is a bit strange that the outside class uses the submap() accessor to determine to which submaps these points were added - the caller needs to know implementation details of this class (which is the reason this PR requires changing the trajectory builder implementation).
How about we'd change the interface of this method (and in 3D) to actually return the affected submaps directly? Maybe that's something for another PR though.

cartographer/mapping/internal/2d/pose_graph_2d_test.cc
```diff
@@ -171,17 +171,17 @@ class PoseGraph2DTest : public ::testing::Test {
     const sensor::PointCloud new_point_cloud = sensor::TransformPointCloud(
         point_cloud_,
         transform::Embed3D(current_pose_.inverse().cast<float>()));
-    std::vector<std::shared_ptr<const Submap2D>> insertion_submaps;
-    for (const auto& submap : active_submaps_->submaps()) {
-      insertion_submaps.push_back(submap);
-    }
+
```
CodeArno: Nit: Remove empty line?

cartographer/mapping/internal/3d/local_trajectory_builder_3d.cc
```diff
@@ -332,15 +334,15 @@ LocalTrajectoryBuilder3D::InsertIntoSubmap(
...
+  active_submaps_.InsertRangeData(filtered_range_data_in_local,
+                                  gravity_alignment);
   std::vector<std::shared_ptr<const mapping::Submap3D>> insertion_submaps;
   for (const std::shared_ptr<mapping::Submap3D>& submap :
        active_submaps_.submaps()) {
     insertion_submaps.push_back(submap);
   }
-  active_submaps_.InsertRangeData(filtered_range_data_in_local,
-                                  gravity_alignment);
+
```
CodeArno: Nit: Remove empty line

cartographer/mapping/2d/submap_2d.cc
schwoere: Done

cartographer/mapping/internal/2d/pose_graph_2d_test.cc
schwoere: Done

cartographer/mapping/internal/3d/local_trajectory_builder_3d.cc
schwoere: Done

cartographer/mapping/2d/submap_2d.cc
schwoere: I see your point. I think it's up to the definition: When is a submap considered "active". At least for now, the submap() getter will return those submaps the range data was inserted last time. @wohe do you have any opinion on that?

wohe: Seems reasonable to me. Return the submaps directly from the insertion function. We still need the separate `submaps()` function in this case since it is also used to determine which submap to use for scan matching.

cartographer/mapping/2d/submap_2d_test.cc
```diff
@@ -79,14 +79,18 @@ TEST(Submap2DTest, TheRightNumberOfRangeDataAreInserted) {
...
   }
   EXPECT_EQ(2, submaps.submaps().size());
-  int correct_num_range_data = 0;
+  int correct_num_finished_submaps = 0;
+  int num_unfinished_submaps = 0;
   for (const auto& submap : all_submaps) {
     if (submap->num_range_data() == kNumRangeData * 2) {
-      ++correct_num_range_data;
+      ++correct_num_finished_submaps;
+    } else if (submap->num_range_data() == kNumRangeData) {
```
wohe: Move the condition into an `EXPECT_EQ`? Even if the actual `num_range_data()` is, this would still be an unfinished submap.

cartographer/mapping/3d/submap_3d.cc
```diff
@@ -297,19 +290,24 @@ std::vector<std::shared_ptr<Submap3D>> ActiveSubmaps3D::submaps() const {
...
+    submaps_.front()->Finish();
   }
 }
 
 void ActiveSubmaps3D::AddSubmap(const transform::Rigid3d& local_submap_pose) {
-  if (submaps_.size() > 1) {
-    submaps_.front()->Finish();
+  if (submaps_.size() >= 2) {
+    // This will crop the finished Submap before inserting a new Submap to
+    // reduce peak memory usage a bit.
```
wohe: Can't we add a `CHECK` similar to 2D?

cartographer/mapping/internal/2d/local_trajectory_builder_2d.cc
```diff
@@ -259,13 +262,11 @@ LocalTrajectoryBuilder2D::InsertIntoSubmap(
     return nullptr;
   }
 
-  // Querying the active submaps must be done here before calling
-  // InsertRangeData() since the queried values are valid for next insertion.
+  active_submaps_.InsertRangeData(range_data_in_local);
   std::vector<std::shared_ptr<const Submap2D>> insertion_submaps;
   for (const std::shared_ptr<Submap2D>& submap : active_submaps_.submaps()) {
```
wohe: Can we change `submaps()` to return a ` std::vector<std::shared_ptr<const Submap2D>>`? Similar in 3D.

schwoere: Do you want to move the for-loop into the getter?
There is no `submaps()` getter in 3D that returns `std::vector<std::shared_ptr<const Submap2D>>`.

schwoere: Okay, I will re-construct the vector with `const Submap2D`.

cartographer/mapping/3d/submap_3d.cc
schwoere: done

cartographer/mapping/2d/submap_2d_test.cc
schwoere: Sorry I don't understand the second sentence, but I moved the condition into the `EXPECT_EQ`.

schwoere:
ptal @wohe

wohe:
@wally-the-cartographer merge

wally-the-cartographer:

PR #1258, opened 2018-07-11
[GenericPoseGraph] Add protos for nodes/constraints.

pifon2a:
Does not cover all constraints yet.

### DIFF
--- a/cartographer/pose_graph/proto/constraint.proto
+++ b/cartographer/pose_graph/proto/constraint.proto
@@ -1,0 +1,26 @@
+// Copyright 2018 The Cartographer Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+syntax = "proto3";
+
+import "cartographer/pose_graph/proto/loss_function.proto";
+import "cartographer/pose_graph/proto/cost_function.proto";
+
+package cartographer.pose_graph.proto;
+
+message Constraint {
+  string id = 1;
+  CostFunction cost_function = 2;
+  LossFunction loss_function = 3;
+}
--- a/cartographer/pose_graph/proto/cost_function.proto
+++ b/cartographer/pose_graph/proto/cost_function.proto
@@ -23,7 +23,7 @@ message RelativePose2D {
   NodeId first = 1;
   NodeId second = 2;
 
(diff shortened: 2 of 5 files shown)


cartographer/pose_graph/proto/cost_function.proto
```diff
@@ -0,0 +1,68 @@
...
+import "cartographer/pose_graph/proto/node.proto";
+import "cartographer/transform/proto/transform.proto";
+
+package cartographer.pose_graph.proto;
+
+message RelativePose2D {
+  NodeId first = 1;
+  NodeId second = 2;
+
+  transform.proto.Rigid2d first_from_second = 3;
```
mbokeloh: Is this Cartographers notation?
In Tango we had second_T_first to denote a transformation from first to second; this has the advantage that we can chain these pretty easily:

third_T_first = third_T_second * second_T_first

This notation is also pretty common in robotics.

cartographer/pose_graph/proto/constraint.proto
```diff
@@ -0,0 +1,26 @@
...
+
+import "cartographer/pose_graph/proto/loss_function.proto";
+import "cartographer/pose_graph/proto/cost_function.proto";
+
+package cartographer.pose_graph.proto;
+
+message Constraint {
+  string id = 1;
+  CostFunction cost_function = 2;
+  LossFunction loss_function = 3;
```
mbokeloh: Does this allow loss_function = empty? Or will this always default to Huber loss?

pifon2a: Yes, it allows it to be empty. We can explicitly have an enum value for that inside  `LossFunction`.

cartographer/pose_graph/proto/cost_function.proto
pifon2a: Let's change to `first_T_second`, sure. As long as it can be chained.

cartographer/pose_graph/proto/constraint.proto
pifon2a: plz, take a look at `loss_function.proto`. I do not know yet how to express the default parabola explicity. Empty message?

cartographer/pose_graph/proto/node.proto
```diff
@@ -0,0 +1,63 @@
...
+    Pose2D pose_2d = 1;
+    Pose3D pose_3d = 2;
+    ImuCalibration imu_calibration = 3;
+  }
+}
+
+message Node {
+  NodeId id = 1;
+  bool constant = 2;
+  Parameters parameters = 3;
```
mbokeloh: Instead of having a wrapper type 'Parameters', you could also the oneof {} block directly in the node. I don't have any preference here, just a thought.

cartographer/pose_graph/proto/constraint.proto
mbokeloh: Yes, let's try the empty message 
message QuadraticLoss {}

If that works it would be nice and clean.

mbokeloh:
LGTM

mbokeloh:
@wally-the-cartographer merge

wally-the-cartographer:
Merge requested by authorized user mbokeloh. Merge queue now has a length of 2.

Issue #1250, opened 2018-07-10
Foreground thread is not smooth due to PoseGraph lock contention


danielsievers:
The issue is observed with PoseGraph3D where the real time performance is not smooth.
But it probably affects Posegraph2D in a similar fashion.

Compiling the necessary data for constraint computation loops over all the submaps while holding the lock. Furthermore, this workload can be chunky when submaps are newly finished.
Sometimes (when the work queue was drained) we also process these tasks in the foreground instead of the background.

The lock contention is also discussed in https://github.com/googlecartographer/cartographer/issues/1138

The issue can be fixed by:
- not processing work queue items in the foreground
- having a separate mutex for work_queue_ access
- making the lock in ComputeConstraintsForNode() more fine-grained and not holding PoseGraph::mutex_ while also holding the constraint builder mutex

danielsievers:
I'll work on pushing the necessary changes.

PR #1264, opened 2018-07-11
Take PoseGraph3D mutex inside instead of outside work item

danielsievers:
Take PoseGraph3D mutex inside instead of outside work item
    
This refactors the code but does not alter behavior except
that the shared mutex is separately taken for work_queue_ access,
released, and then re-acquired inside each work item.
    
As a consequence AddWorkItem() also needs to EXCLUDE() the lock and
acquire it internally, because for the case where we run the task
in the foreground we cannot hold the lock and recursively acquire it again
inside the task being run.
    
This prepares for making locking more fine-grained, see issue #1250.


schwoere:
@wohe do you want to have a look now?

cartographer/mapping/internal/3d/pose_graph_3d.cc
```diff
@@ -111,31 +111,33 @@ NodeId PoseGraph3D::AddNode(
...
-  if (data_.submap_data.SizeOfTrajectoryOrZero(trajectory_id) == 0 ||
-      std::prev(data_.submap_data.EndOfTrajectory(trajectory_id))
-              ->data.submap != insertion_submaps.back()) {
-    // We grow 'data_.submap_data' as needed. This code assumes that the first
-    // time we see a new submap is as 'insertion_submaps.back()'.
-    const SubmapId submap_id =
-        data_.submap_data.Append(trajectory_id, InternalSubmapData());
-    data_.submap_data.at(submap_id).submap = insertion_submaps.back();
-    LOG(INFO) << "Inserted submap " << submap_id << ".";
+  NodeId node_id(0, 0);
```
wohe: The constructor for `NodeId` was introduced to prevent uninitialized IDs. This seems not much better. How about moving the code under the mutex lock into a separate function? Then you can take the mutex there and return the `node_id`.

danielsievers: Done.

danielsievers:
Regarding PoseGraph2D: 
Regarding the performance problem I don't know how severe it is in 2D, but regardless it would be nice to keep the code more aligned. That being said, maybe it's worth a refactoring to pull out some common code for this (everything related to work queue processing). Either way, maybe it makes sense to finish all the work for 3D first.

wally-the-cartographer:

wohe:
@gaschler Can you verify?

PR #1267, opened 2018-07-12
Added function GetThreadCpuTimeSeconds in common/time.h

spielawa:
This will be needed for measuring SLAM performance.


cartographer/common/time.h
```diff
@@ -60,6 +60,8 @@ int64 ToUniversal(Time time);
 // For logging and unit tests, outputs the timestamp integer.
 std::ostream& operator<<(std::ostream& os, Time time);
 
+double GetThreadCpuTimeSeconds();
```
wohe: Please add a comment what this returns: CPU time consumed by the thread so far.

cartographer/common/time.cc
```diff
@@ -50,5 +51,12 @@ common::Duration FromMilliseconds(const int64 milliseconds) {
       std::chrono::milliseconds(milliseconds));
 }
 
+double GetThreadCpuTimeSeconds() {
+  struct timespec tm;
+  clock_gettime(CLOCK_THREAD_CPUTIME_ID, &tm);
+  return static_cast<double>(tm.tv_sec) +
```
wohe: The `static_cast<double>` is not really needed and could be removed.

cartographer/common/time.cc
```diff
@@ -50,5 +51,12 @@ common::Duration FromMilliseconds(const int64 milliseconds) {
       std::chrono::milliseconds(milliseconds));
 }
 
+double GetThreadCpuTimeSeconds() {
+  struct timespec tm;
```
wohe: Why `tm`? `thread_cpu_time`?

cartographer/common/time.cc
```diff
@@ -50,5 +51,11 @@ common::Duration FromMilliseconds(const int64 milliseconds) {
       std::chrono::milliseconds(milliseconds));
 }
 
+double GetThreadCpuTimeSeconds() {
+  struct timespec thread_cpu_time;
+  clock_gettime(CLOCK_THREAD_CPUTIME_ID, &thread_cpu_time);
+  return thread_cpu_time.tv_sec + 1e-9 * thread_cpu_time.tv_nsec;
```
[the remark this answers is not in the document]
wohe: `1e-9` is `double`, so this is fine. Single precision would be `1e-9f`.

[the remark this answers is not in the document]
CodeArno: Ah, you're right.

cartographer/common/time.cc
```diff
@@ -15,7 +15,9 @@
  */
 
 #include "cartographer/common/time.h"
+#include "glog/logging.h"
```
danielsievers: +#include <cerrno>
+#include <cstring>

cartographer/common/time.cc
```diff
@@ -50,5 +52,13 @@ common::Duration FromMilliseconds(const int64 milliseconds) {
       std::chrono::milliseconds(milliseconds));
 }
 
+double GetThreadCpuTimeSeconds() {
+  struct timespec thread_cpu_time;
+  if (clock_gettime(CLOCK_THREAD_CPUTIME_ID, &thread_cpu_time) == -1) {
```
CodeArno: There are several error codes, e.g. see: https://linux.die.net/man/3/clock_gettime
Checking "if (!clock_gettime() != 0)" is safer.

spielawa: "clock_gettime(), clock_settime() and clock_getres() return 0 for success, or -1 for failure (in which case errno is set appropriately)."

cartographer/common/time.cc
```diff
@@ -15,7 +15,11 @@
  */
 
 #include "cartographer/common/time.h"
+#include "glog/logging.h"
```
wohe: This belongs after the system includes.

cartographer/common/time.cc
wohe: +1. Let's check `!= 0`which is more common. Also: Use a `CHECK`? Right now we return bogus values on error.

spielawa: ok :) done.

cartographer/common/time.cc
```diff
@@ -50,5 +55,12 @@ common::Duration FromMilliseconds(const int64 milliseconds) {
       std::chrono::milliseconds(milliseconds));
 }
 
+double GetThreadCpuTimeSeconds() {
+  struct timespec thread_cpu_time;
+  CHECK(clock_gettime(CLOCK_THREAD_CPUTIME_ID, &thread_cpu_time) == 0)
+      << "clock_gettime failed: " << std::strerror(errno);
```
wohe: `CHECK` should already print the precise call that failed, does it not? So the first part `"clock_gettime failed: "` is redundant. Remove?

spielawa: Good point. Done.

wohe:
@wally-the-cartographer merge

PR #1270, opened 2018-07-12
Add function to query if metrics are registered.

MichaelGrupp:
Useful to decide whether to run code that's only relevant if metrics
were registered.

### DIFF
--- a/cartographer/mapping/internal/2d/local_trajectory_builder_2d.cc
+++ b/cartographer/mapping/internal/2d/local_trajectory_builder_2d.cc
@@ -21,6 +21,7 @@
 
 #include "cartographer/common/make_unique.h"
 #include "cartographer/metrics/family_factory.h"
+#include "cartographer/metrics/register.h"
 #include "cartographer/sensor/range_data.h"
 
 namespace cartographer {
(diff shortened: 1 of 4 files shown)


cartographer/mapping/internal/2d/local_trajectory_builder_2d.cc
```diff
@@ -88,7 +89,7 @@ std::unique_ptr<transform::Rigid2d> LocalTrajectoryBuilder2D::ScanMatch(
                             filtered_gravity_aligned_point_cloud,
                             *matching_submap->grid(), pose_observation.get(),
                             &summary);
-  if (pose_observation) {
+  if (pose_observation && metrics::MetricsRegistered()) {
```
MichaelGrupp: @gaschler An example in the current code - makes it obvious that the following code block is only for metrics.

MichaelGrupp: I'd prefer it for instrumentation code that is  a bit more involved than just a single Observe() etc.

gaschler: We already have a lot of code paths, so I prefer instrumentation to always execute.
cschuet@ pfion2a@ What do you think?

MichaelGrupp:
Closing this for now.

PR #1277, opened 2018-07-13
Store histogram in submap

schwoere:
The histogram of a submap is now stored in the submap (class and proto) itself. This change allows to accumulate the histogram of a submap in local SLAM by adding up the histogram of each new scan.

The main advantage is that the background thread doesn't have to loop over all `TrajectoryNode`s of a finished submap to compute the submap histogram for the `RotationalScanMatcher`. Instead this chunk of work is moved to the local SLAM thread but is split up into a few computations for each new scan. When running localization, the histogram of a submap can just be read from a map pbstream and does not have to be computed from the nodes.

In summary:
- This change improved the CPU time of offline SLAM by ~7%.
- Increases the readability of the code and performance of the background thread. (see `PoseGraph3D::ComputeConstraint`)
- No negative performance impacts on accuracy or finding loop-closures

However:
- With this change to the submap proto, old maps (pbstreams) are no longer supported and need to be re-created by running offline slam


schwoere:
Please do not review, yet.

cartographer/mapping/3d/submap_3d.h
```diff
@@ -92,9 +101,11 @@ class ActiveSubmaps3D {
   // Inserts 'range_data' into the Submap collection. 'gravity_alignment' is
   // used for the orientation of new submaps so that the z axis approximately
   // aligns with gravity.
-  std::vector<std::shared_ptr<const Submap3D>> InsertRangeData(
+  std::vector<std::shared_ptr<const Submap3D>> InsertData(
       const sensor::RangeData& range_data,
-      const Eigen::Quaterniond& gravity_alignment);
+      const Eigen::Quaterniond& gravity_alignment,
+      const Eigen::VectorXf& rotational_scan_matcher_histogram_in_gravity,
```
schwoere: Do you think it's a better idea to pass the histogram rotated into the local map frame for better readability?

cartographer/mapping/3d/submap_3d.cc
```diff
@@ -211,6 +214,12 @@ Submap3D::Submap3D(const proto::Submap3D& proto)
           common::make_unique<HybridGrid>(proto.low_resolution_hybrid_grid())) {
   set_num_range_data(proto.num_range_data());
   set_finished(proto.finished());
+  rotational_scan_matcher_histogram_.resize(
```
danielsievers: Eigen says 'resizing a matrix to 0 is not supported'. 
Can you either handle this here (not sure if it should or can happen) or CHECK(proto.rotational_scan_matcher_histogram_size())?

cartographer/mapping/3d/submap_3d.cc
```diff
@@ -246,6 +261,14 @@ void Submap3D::UpdateFromProto(const proto::Submap& proto) {
                   submap_3d.low_resolution_hybrid_grid())
             : nullptr;
   }
+
+  rotational_scan_matcher_histogram_.resize(
```
danielsievers: see above... also since it's the same code maybe a small helper?

cartographer/mapping/3d/submap_3d.cc
```diff
@@ -272,6 +297,15 @@ void Submap3D::InsertRangeData(const sensor::RangeData& range_data,
   range_data_inserter.Insert(transformed_range_data,
                              low_resolution_hybrid_grid_.get());
   set_num_range_data(num_range_data() + 1);
+  if (rotational_scan_matcher_histogram_.rows() == 0) {
+    rotational_scan_matcher_histogram_.resize(scan_histogram_in_gravity.size());
+    rotational_scan_matcher_histogram_.setZero();
+  }
+  const float yaw_to_submap_from_gravity =
```
danielsievers: i think the name is a bit confusing, since it's the (gravity-aligned) yaw for the scan, relative to the submap's orientation, right? yaw_in_submap?

cartographer/mapping/3d/submap_3d.h
danielsievers: Would that avoid having to pass in local_from_gravity? Then yes.

cartographer/mapping/internal/3d/scan_matching/fast_correlative_scan_matcher_3d_test.cc
```diff
@@ -120,7 +116,11 @@ class FastCorrelativeScanMatcher3DTest : public ::testing::Test {
                                 {},
                                 point_cloud_,
                                 low_resolution_point_cloud,
-                                Eigen::VectorXf::Zero(10)};
+                                GetRotationalScanMatcherHstogram()};
+  }
+
+  static Eigen::VectorXf GetRotationalScanMatcherHstogram() {
```
danielsievers: typo: Hstogram

cartographer/mapping/internal/3d/scan_matching/fast_correlative_scan_matcher_3d_test.cc
```diff
@@ -120,7 +116,11 @@ class FastCorrelativeScanMatcher3DTest : public ::testing::Test {
...
+    return Eigen::VectorXf::Zero(10);
```
danielsievers: how do the tests work if the histogram used to build the matcher only has zeros? (i know you didn't change that, just wondering)

cartographer/mapping/3d/submap_3d.cc
```diff
@@ -225,6 +234,12 @@ void Submap3D::ToProto(proto::Submap* const proto,
     *submap_3d->mutable_low_resolution_hybrid_grid() =
         low_resolution_hybrid_grid().ToProto();
   }
+  submap_3d->clear_rotational_scan_matcher_histogram();
```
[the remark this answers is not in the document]
schwoere: @danielsievers @wohe . Initially, I planned to fix the call site after this PR, but now I feel it is much better to first fix the call site and interface and then update this PR (see [#1294](https://github.com/googlecartographer/cartographer/pull/1294)). This changes the interface of `Submap::ToProto` such that the Proto is now returned.

cartographer/mapping/internal/constraints/constraint_builder_3d.cc
```diff
@@ -180,12 +175,13 @@ ConstraintBuilder3D::DispatchScanMatcherConstruction(
...
   auto scan_matcher_task = common::make_unique<common::Task>();
+  const auto& histogram = submap->rotational_scan_matcher_histogram();
   scan_matcher_task->SetWorkItem(
-      [&submap_scan_matcher, &scan_matcher_options, submap_nodes]() {
+      [&submap_scan_matcher, &scan_matcher_options, histogram]() {
         submap_scan_matcher.fast_correlative_scan_matcher =
             common::make_unique<scan_matching::FastCorrelativeScanMatcher3D>(
                 *submap_scan_matcher.high_resolution_hybrid_grid,
-                submap_scan_matcher.low_resolution_hybrid_grid, submap_nodes,
+                submap_scan_matcher.low_resolution_hybrid_grid, histogram,
```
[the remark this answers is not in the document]
schwoere: I considered storing the histogram inside in `SubmapScanMatcher`, but this struct is only used for the call to `fast_correlative_scan_matcher->Match`. This call does not require the histogram, as it's already stored inside the matcher.
However, I now pass the histogram as pointer to follow the Google Style guide.

cartographer/mapping/internal/3d/scan_matching/rotational_scan_matcher.h
```diff
@@ -28,16 +28,17 @@ namespace scan_matching {
 
 class RotationalScanMatcher {
  public:
+  // Rotates the given 'histogram' by the given 'angle'. This might lead to
```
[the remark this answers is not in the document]
schwoere: It's not possible to extract everything but I extracted a part now in #1316.

cartographer/mapping/3d/submap_3d.h
schwoere: okay done

cartographer/mapping/internal/constraints/constraint_builder_3d.cc
```diff
@@ -180,12 +175,13 @@ ConstraintBuilder3D::DispatchScanMatcherConstruction(
   auto& scan_matcher_options =
       options_.fast_correlative_scan_matcher_options_3d();
   auto scan_matcher_task = common::make_unique<common::Task>();
+  const auto& histogram = submap->rotational_scan_matcher_histogram();
   scan_matcher_task->SetWorkItem(
-      [&submap_scan_matcher, &scan_matcher_options, submap_nodes]() {
+      [&submap_scan_matcher, &scan_matcher_options, histogram]() {
```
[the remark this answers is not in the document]
schwoere: The `histogram` is now passed as a pointer, copying the pointer should be fine.

cartographer/mapping/3d/submap_3d.h
```diff
@@ -92,9 +101,11 @@ class ActiveSubmaps3D {
   // Inserts 'range_data' into the Submap collection. 'gravity_alignment' is
```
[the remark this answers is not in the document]
schwoere: done

cartographer/mapping/3d/submap_3d.h
```diff
@@ -43,7 +43,9 @@ proto::SubmapsOptions3D CreateSubmapsOptions3D(
 class Submap3D : public Submap {
  public:
   Submap3D(float high_resolution, float low_resolution,
-           const transform::Rigid3d& local_submap_pose);
+           const transform::Rigid3d& local_submap_pose,
+           const Eigen::VectorXf& rotational_scan_matcher_histogram =
+               Eigen::VectorXf());
```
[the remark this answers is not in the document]
schwoere: I changed the constructor. Now the default argument is not requirement.

cartographer/mapping/3d/submap_3d.cc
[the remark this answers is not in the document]
schwoere: With the API change in #1294 this is no longer required.

cartographer/mapping/internal/3d/scan_matching/fast_correlative_scan_matcher_3d_test.cc
schwoere: In these tests the `FastCorrelativeScanMatcher` is initialized with a histogram containing just zeros. The tests call `CreateConstantData` which creates a `TrajectoryNode` with a histogram containing just zeros. So matching the histograms is quite trivial in this case.

schwoere:
@cschuet I would like to hear your opinion on this change, as this would change the map pbstream.

cschuet:
@schwoere Is it possible to make this change backwards-compatible by adding login in pbstream deserialization that calculates the histogram if it is not found in the pbstream?

schwoere:
@cschuet  yes, that would be possible.

wohe:
@cschuet Do you think bumping the version number and adding it to the migration tool is enough? Or do you think we should retain backwards compatibility? If we do, when do we plan to drop it again? Cartographer 2.0 or sooner?

cartographer/mapping/3d/submap_3d.cc
```diff
@@ -196,12 +197,20 @@ proto::SubmapsOptions3D CreateSubmapsOptions3D(
 }
 
 Submap3D::Submap3D(const float high_resolution, const float low_resolution,
-                   const transform::Rigid3d& local_submap_pose)
+                   const transform::Rigid3d& local_submap_pose,
+                   const int rotational_scan_matcher_histogram_size)
+    : Submap3D(high_resolution, low_resolution, local_submap_pose,
+               Eigen::VectorXf::Zero(rotational_scan_matcher_histogram_size)) {}
```
wohe: No need for this convenience constructor. Call it like this at the call sites of this constructor instead?

cschuet:
@wohe If the effort to maintain backwards-compatibility is limited I would go for it (and here it seems like it really is the same effort whether you stick it into the load routine or the migration tool). There are companies that use Cartographer in production and have a large number of maps that they would need to migrate. Typically maps go stale quickly so I think removing it after 2 month or so should be fine. WDYT? Do you think that is a reasonable trade-off between trying to soften impact on the ecosystem and you having to go through extra hoops?

wohe:
@cschuet Sounds good to me.

sebastianklose:
@cschuet +1 to updating the version number and supporting reading both and writing only the newest.
The migration tool was meant to overcome breaking changes (like the introduction of the header). It seems this change can be handled by additional reader logic.

cartographer/mapping/internal/3d/local_trajectory_builder_3d.cc
```diff
@@ -352,19 +352,25 @@ LocalTrajectoryBuilder3D::InsertIntoSubmap(
...
       scan_matching::RotationalScanMatcher::ComputeHistogram(
           sensor::TransformPointCloud(
               filtered_range_data_in_tracking.returns,
               transform::Rigid3f::Rotation(gravity_alignment.cast<float>())),
           options_.rotational_histogram_size());
+
+  const auto local_from_gravity_aligned =
+      pose_estimate.rotation() * gravity_alignment.inverse();
+  const float local_yaw_from_gravity =
+      transform::GetYaw(local_from_gravity_aligned);
```
wohe: This is potentially numerically unstable. A better approach would be to forward the histogram unchanged along with the transform to local frame, combine all transforms in the end (this is then from an approximately gravity aligned frame to a different approximately gravity aligned frame, so mostly a rotation in yaw with small roll/pitch component, thus numerically stable) and then rotate the histogram by the yaw from this transform.

cartographer/mapping/3d/submap_3d.h
```diff
@@ -97,14 +102,18 @@ class ActiveSubmaps3D {
   // Inserts 'range_data_in_local' into the Submap collection.
   // 'local_from_gravity_aligned' is used for the orientation of new submaps so
   // that the z axis approximately aligns with gravity.
-  std::vector<std::shared_ptr<const Submap3D>> InsertRangeData(
+  // 'rotational_scan_matcher_histogram_in_local' will be accumulated in all
```
wohe: This comment needs to be fixed: `rotational_scan_matcher_histogram_in_gravity`

cartographer/mapping/internal/3d/local_trajectory_builder_3d.cc
```diff
@@ -352,19 +352,19 @@ LocalTrajectoryBuilder3D::InsertIntoSubmap(
...
-
-  const Eigen::VectorXf rotational_scan_matcher_histogram =
+  const Eigen::VectorXf rotational_scan_matcher_histogram_in_gravity =
       scan_matching::RotationalScanMatcher::ComputeHistogram(
           sensor::TransformPointCloud(
               filtered_range_data_in_tracking.returns,
               transform::Rigid3f::Rotation(gravity_alignment.cast<float>())),
           options_.rotational_histogram_size());
+
+  const auto local_from_gravity_aligned =
```
wohe: Isn't `auto` a potential bug here? The type is an expression template and depending on the temporaries constructed in evaluating it, this might be use after free. See https://eigen.tuxfamily.org/dox/TopicPitfalls.html

cartographer/mapping/internal/3d/local_trajectory_builder_3d.cc
schwoere: Good catch! I fixed it.

cartographer/mapping/3d/submap_3d.h
schwoere: done

PR #1278, opened 2018-07-13
[GenericPoseGraph] Add IMU Calibration and Pose3D node types.

pifon2a:


pifon2a:
@wally-the-cartographer merge

cartographer/pose_graph/node/imu_calibration.cc
```diff
@@ -0,0 +1,49 @@
...
+
+  // TODO(pifon): Use a common method to convert from Eigen::Quaterniond to
+  // proto. Probably, the one defined in transform.h.
+  auto* orientation = imu_calibration->mutable_orientation();
+  orientation->set_w(orientation_[0]);
+  orientation->set_x(orientation_[1]);
+  orientation->set_y(orientation_[2]);
+  orientation->set_z(orientation_[3]);
+
+  std::cout << parameters.DebugString();
```
mbokeloh: Do you want that to be in there?

cartographer/pose_graph/node/nodes.h
```diff
@@ -0,0 +1,39 @@
...
+#include "cartographer/pose_graph/node/pose_2d.h"
+#include "cartographer/pose_graph/node/pose_3d.h"
+
+#include <map>
+
+namespace cartographer {
+namespace pose_graph {
+
+struct Nodes {
+  // TODO(pifon): Should it really be an std::map or smth else?
```
mbokeloh: Let's first try it with map but do an experiment once everything is done with an std::unordered_map to see if there is a difference.

cartographer/pose_graph/node/imu_calibration.cc
```diff
@@ -0,0 +1,47 @@
...
+
+namespace cartographer {
+namespace pose_graph {
+
+ImuCalibration::ImuCalibration(const NodeId& node_id, bool constant,
+                               double gravity_constant,
+                               const Eigen::Quaterniond& orientation)
+    : Node(node_id, constant),
+      gravity_constant_(gravity_constant),
+      orientation_{{orientation.w(), orientation.x(), orientation.y(),
```
mbokeloh: I assume that the order is important but it comes at a surprise. Could you add a comment somewhere why that is? Alternatively, could we keep the Eigen::Quaterniond format?

Also the same below in Pose3D. Maybe we can have a converter function?

cartographer/pose_graph/node/nodes.h
pifon2a: Agreed.

cartographer/pose_graph/node/imu_calibration.cc
pifon2a: I just kept Cartographer's format. But I agree, let's keep Eigen's.

pifon2a:
@wally-the-cartographer merge

wally-the-cartographer:

PR #1284, opened 2018-07-16
Introduce separate mutex for PoseGraph3D work queue access

danielsievers:
Introduce separate mutex for PoseGraph3D work queue access.

This reduces lock contention, see issue #1250.
In particular higher frequency call to AddWorkItem
don't need to take the main mutex from the froground
thread anymore.

### DIFF
--- a/cartographer/mapping/internal/3d/pose_graph_3d.cc
+++ b/cartographer/mapping/internal/3d/pose_graph_3d.cc
@@ -553,10 +553,10 @@ void PoseGraph3D::WaitForAllComputations() {
       std::ostringstream progress_info;
       progress_info << "Optimizing: " << std::fixed << std::setprecision(1)
                     << 100. *
-                         (constraint_builder_.GetNumFinishedNodes() -
-                          num_finished_nodes_at_start) /
-                         (num_trajectory_nodes - num_finished_nodes_at_start)
-                  << "%...";
+                           (constraint_builder_.GetNumFinishedNodes() -
+                            num_finished_nodes_at_start) /
+                           (num_trajectory_nodes - num_finished_nodes_at_start)
+                    << "%...";
       std::cout << "\r\x1b[K" << progress_info.str() << std::flush;
     }
   }
--- a/cartographer/mapping/internal/3d/pose_graph_3d.cc
+++ b/cartographer/mapping/internal/3d/pose_graph_3d.cc
@@ -522,27 +522,28 @@ void PoseGraph3D::WaitForAllComputations() {
   // a WhenDone() callback.
   {
     common::MutexLocker locker(&work_queue_mutex_);
-    locker.Await(
-        [this]() REQUIRES(work_queue_mutex_) { return work_queue_ == nullptr; });
+    while (!locker.AwaitWithTimeout(
+        [this]() REQUIRES(work_queue_mutex_) { return work_queue_ == nullptr; },
+        common::FromSeconds(1.))) {
+      std::ostringstream progress_info;
+      progress_info << "Processing work queue: "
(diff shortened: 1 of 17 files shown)


cartographer/mapping/internal/3d/pose_graph_3d.cc
```diff
@@ -509,40 +509,59 @@ void PoseGraph3D::HandleWorkQueue(
...
+  }
+
   const int num_finished_nodes_at_start =
       constraint_builder_.GetNumFinishedNodes();
+
+  // First wait for the work queue to drain so that it's safe to schedule
+  // a WhenDone() callback.
+  {
+    common::MutexLocker locker(&work_queue_mutex_);
+    locker.Await([this]() REQUIRES(work_queue_mutex_) {
```
wohe: This can take a long time. Can we switch this to a while loop with `AwaitWithTimeout()` similar to below? The functionality to output the progress can be deduplicated extracting it into a local function with a lambda.

cartographer/mapping/internal/3d/pose_graph_3d.cc
```diff
@@ -509,40 +509,59 @@ void PoseGraph3D::HandleWorkQueue(
...
+      return work_queue_ == nullptr;
+    });
+  }
+
+  // Now wait for any pending constraint computations to finish.
+  common::Mutex notification_mutex;
```
wohe: Why not reuse the `mutex_`? Seems this complicates the code quite a bit without a perceivable performance benefit.

danielsievers: Done.

cartographer/mapping/internal/3d/pose_graph_3d.cc
danielsievers: Done.

cartographer/mapping/internal/3d/pose_graph_3d.cc
```diff
@@ -522,27 +522,29 @@ void PoseGraph3D::WaitForAllComputations() {
   // a WhenDone() callback.
   {
     common::MutexLocker locker(&work_queue_mutex_);
-    locker.Await(
-        [this]() REQUIRES(work_queue_mutex_) { return work_queue_ == nullptr; });
+    while (!locker.AwaitWithTimeout(
+        [this]() REQUIRES(work_queue_mutex_) { return work_queue_ == nullptr; },
+        common::FromSeconds(1.))) {
+      std::ostringstream progress_info;
```
wohe: I thought we could use the same progress report as below, i.e. percentage of finished nodes. No need to switch to a different style here.

danielsievers: done. ptal.

cartographer/mapping/internal/3d/pose_graph_3d.cc
```diff
@@ -518,46 +518,48 @@ void PoseGraph3D::WaitForAllComputations() {
...
+      [this,
+       &notification](const constraints::ConstraintBuilder3D::Result& result)
+          EXCLUDES(mutex_) {
+            common::MutexLocker locker(&mutex_);
+            data_.constraints.insert(data_.constraints.end(), result.begin(),
+                                     result.end());
+            notification = true;
           });
   while (!locker.AwaitWithTimeout(
       [&notification]() REQUIRES(notification_mutex) { return notification; },
```
wohe: This should be `mutex_`.

wohe:
@wally-the-cartographer merge

wally-the-cartographer:

PR #1285, opened 2018-07-16
Don't process PoseGraph3D work items in foreground

danielsievers:
This helps the foreground thread in keeping up real time.
Also see issue #1250.

### DIFF
--- a/cartographer/mapping/internal/3d/pose_graph_3d.cc
+++ b/cartographer/mapping/internal/3d/pose_graph_3d.cc
@@ -47,7 +47,8 @@ PoseGraph3D::PoseGraph3D(
     common::ThreadPool* thread_pool)
     : options_(options),
       optimization_problem_(std::move(optimization_problem)),
-      constraint_builder_(options_.constraint_builder_options(), thread_pool) {}
+      constraint_builder_(options_.constraint_builder_options(), thread_pool),
+      thread_pool_(thread_pool) {}
 
 PoseGraph3D::~PoseGraph3D() {
   WaitForAllComputations();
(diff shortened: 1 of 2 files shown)


danielsievers:
ptal, rebased.

cartographer/mapping/internal/3d/pose_graph_3d.cc
```diff
@@ -152,26 +153,19 @@ NodeId PoseGraph3D::AddNode(
...
+  common::MutexLocker locker(&work_queue_mutex_);
+  if (work_queue_ == nullptr) {
+    work_queue_ = common::make_unique<WorkQueue>();
+    auto task = common::make_unique<common::Task>();
+    task->SetWorkItem([this]() { DrainWorkQueue(); });
+    thread_pool_->Schedule(std::move(task));
   }
+  const auto now = std::chrono::steady_clock::now();
+  work_queue_->push_back({now, work_item});
+  kWorkQueueDelayMetric->Set(
```
schwoere: compared to before we now call `Set` in both cases `work_queue_ == nullptr`  and `work_queue_ != nullptr`, but for me this change looks still fine.

cartographer/mapping/internal/3d/pose_graph_3d.h
```diff
@@ -199,6 +199,10 @@ class PoseGraph3D : public PoseGraph {
   void HandleWorkQueue(const constraints::ConstraintBuilder3D::Result& result)
       EXCLUDES(mutex_) EXCLUDES(work_queue_mutex_);
 
+  // Process pending tasks in the work queue on the calling thread, until the
+  // queue is empty.
+  void DrainWorkQueue() EXCLUDES(mutex_) EXCLUDES(work_queue_mutex_);
```
schwoere: DrainWorkQueue() does not exclude `mutex_` anymore, right?

cartographer/mapping/internal/3d/pose_graph_3d.cc
danielsievers: Yes, because before we would basically say that when the work queue was idle and we run a single task immediately in the foreground, that this does not constitute a delay.

cartographer/mapping/internal/3d/pose_graph_3d.h
danielsievers: It has to exclude it because the work items will internally take `mutex_`.

danielsievers: I did update the comment however since it was slightly inaccurate.

wohe:
@wally-the-cartographer merge

wally-the-cartographer:

PR #1292, opened 2018-07-17
Fix Clang warnings.

pifon2a:
1) cartographer/cloud/internal/client_server_test.cc:56:16: warning: unused variable 'kUnsupportedSerializationHeaderProtoString' 
constexpr char kUnsupportedSerializationHeaderProtoString[] 

2) cartographer/mapping/3d/submap_3d.h:105:7: warning: private field 'matching_submap_index_' is not used [-Wunused-private-field]

3) pose_graph_3d.cc:549:21: warning: 'guarded_by' attribute only applies to fields and global variables
  bool notification GUARDED_BY(mutex_) = false;

### DIFF
--- a/cartographer/mapping/internal/3d/pose_graph_3d.cc
+++ b/cartographer/mapping/internal/3d/pose_graph_3d.cc
@@ -546,7 +546,7 @@ void PoseGraph3D::WaitForAllComputations() {
 
   // Now wait for any pending constraint computations to finish.
   common::MutexLocker locker(&mutex_);
-  bool notification GUARDED_BY(mutex_) = false;
+  bool notification = false;
   constraint_builder_.WhenDone(
       [this,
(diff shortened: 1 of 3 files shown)


cartographer/mapping/internal/3d/pose_graph_3d.cc
```diff
@@ -546,7 +546,7 @@ void PoseGraph3D::WaitForAllComputations() {
 
   // Now wait for any pending constraint computations to finish.
   common::MutexLocker locker(&mutex_);
-  bool notification GUARDED_BY(mutex_) = false;
+  bool notification = false;
```
cschuet: @wohe I don't think this does anything useful here does it?

wohe: I agree. Let's remove it.

PR #1300, opened 2018-07-19
[GenericPoseGraph] Add CeresOptimizer.

pifon2a:


cartographer/pose_graph/optimizer/optimizer.h
```diff
@@ -0,0 +1,39 @@
...
+
+class Optimizer {
+ public:
+  Optimizer() = default;
+  virtual ~Optimizer() = default;
+
+  Optimizer(const Optimizer&) = delete;
+  Optimizer& operator=(const Optimizer&) = delete;
+
+  virtual ceres::Solver::Summary Solve(PoseGraphData* data) const = 0;
```
mbokeloh: I think we might have to abstract here from ceres since this base class is intended for using it with other optimizers than ceres, if I'm not mistaken. 
How about an enum class returning whether it converged or not?

pifon2a:
@wally-the-cartographer

PR #1304, opened 2018-07-19
Small readability improvement

ojura:


ojura:
The same time variable was defined twice. Also add a missing const.

PR #1316, opened 2018-07-23
Preparation for #1277: refactor rotation scan matcher

schwoere:
Added new constructor for `RotationalScanMatcher` and exposed `RotateHistogram`.

This PR prepares for PR #1277 where the constructor 
`RotationalScanMatcher(const std::vector<std::pair<Eigen::VectorXf, float>>& histograms_at_angles)`
will be removed and only the new constructor
`RotationalScanMatcher(const Eigen::VectorXf& histogram)`
will remain. The unit tests will be updated to use the new constructor in #1277.

### DIFF
--- a/cartographer/mapping/internal/3d/scan_matching/rotational_scan_matcher_test.cc
+++ b/cartographer/mapping/internal/3d/scan_matching/rotational_scan_matcher_test.cc
@@ -28,7 +28,9 @@ namespace {
 TEST(RotationalScanMatcher3DTest, OnlySameHistogramIsScoreOne) {
   Eigen::VectorXf histogram(7);
   histogram << 1.f, 43.f, 0.5f, 0.3123f, 23.f, 42.f, 0.f;
-  RotationalScanMatcher matcher({{histogram, 0.f}});
+  const std::vector<std::pair<Eigen::VectorXf, float>> histogram_at_angle = {
+      {histogram, 0.f}};
+  RotationalScanMatcher matcher(histogram_at_angle);
   const auto scores = matcher.Match(histogram, 0.f, {0.f, 1.f});
   ASSERT_EQ(2, scores.size());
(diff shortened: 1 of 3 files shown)


cartographer/mapping/internal/3d/scan_matching/rotational_scan_matcher_test.cc
```diff
@@ -28,7 +28,8 @@ namespace {
 TEST(RotationalScanMatcher3DTest, OnlySameHistogramIsScoreOne) {
   Eigen::VectorXf histogram(7);
   histogram << 1.f, 43.f, 0.5f, 0.3123f, 23.f, 42.f, 0.f;
-  RotationalScanMatcher matcher({{histogram, 0.f}});
+  RotationalScanMatcher matcher(
```
spielawa: Why this change?

spielawa: oh, ic, that's because you'll eventually remove this constructor. I'd move this change to the other PR then, where you actually remove the constructor.

schwoere: Okay, I reverted the change. Now it has the same semantics as before, but I still have to update the tests, otherwise the compiler complains about ambiguous constructors for `RotationalScanMatcher`.

schwoere:
ptal @spielawa

schwoere:
Thanks @spielawa . Can you now have a look as well, @wohe ?

wohe:
@schwoere The test no longer is changed to use the new constructor. Change the PR description to reflect this?

schwoere:
@wohe thanks for the hint, done!

schwoere:
@wally-the-cartographer merge

wohe:
@wally-the-cartographer merge

PR #1341, opened 2018-07-26
Don't process PoseGraph2D work items in foreground.

### LINKED ISSUE #1250: Foreground thread is not smooth due to PoseGraph lock contention

wohe:
This is #1285 for 2D.
Fixes #1250.

### DIFF
--- a/cartographer/mapping/internal/2d/pose_graph_2d.cc
+++ b/cartographer/mapping/internal/2d/pose_graph_2d.cc
@@ -48,7 +48,8 @@ PoseGraph2D::PoseGraph2D(
     common::ThreadPool* thread_pool)
     : options_(options),
       optimization_problem_(std::move(optimization_problem)),
-      constraint_builder_(options_.constraint_builder_options(), thread_pool) {}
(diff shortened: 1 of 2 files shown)


schwoere:
@wally-the-cartographer merge

wally-the-cartographer:
Merge requested by authorized user schwoere. Merge queue now has a length of 1.

PR #1344, opened 2018-07-27
fix dead code in UpdateFromProto

schwoere:
Removes an additional condition that is not required.


wohe:
@wally-the-cartographer merge

wally-the-cartographer:

PR #1350, opened 2018-07-28
Fix Abseil build command

ojura:
@pifon2a I've noticed the `BUILD_COMMAND` is broken. You're setting `ABSL_BUILD_COMMAND`, but using `ABSEIL_BUILD_COMMAND` (effectively empty). Somehow this doesn't confuse newer CMakes, but the one you install on Trusty gets confused and interprets this as a no-op build step. This causes the indigo pipeline in cartographer_ros to fail.

Use the cross-platform build command instead: 
https://cmake.org/cmake/help/latest/module/ExternalProject.html#obtaining-project-properties (just above the linked paragraph)

### DIFF
--- a/cmake/modules/FindAbseil.cmake
+++ b/cmake/modules/FindAbseil.cmake
@@ -57,19 +57,12 @@ if(NOT TARGET standalone_absl)
     "${ABSEIL_PROJECT_BUILD_DIR}/absl/utility/${prefix}absl_utility${suffix}"
     )
 
-   if(CMAKE_GENERATOR MATCHES "Ninja")
-     set (ABSL_BUILD_COMMAND "ninja")
-   else()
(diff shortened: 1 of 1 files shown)


pifon2a:
@ojura You are the hero of the day. Thank you a lot!

ojura:
You're welcome! I did warn you, though :D https://github.com/googlecartographer/cartographer/pull/1301#issuecomment-406353293

pifon2a:
@ojura yeah, `cmake` is weird.  :)

pifon2a:
@wally-the-cartographer merge

wally-the-cartographer:

ojura:
@pifon2a also WDYT about throwing in a `find_package(Abseil REQUIRED)` into `cartographer-config.cmake.in` so Cartographer users don't have to it by themselves?

PR #1351, opened 2018-07-28
Make standalone_absl imported target global.

ojura:
@pifon2a there's one more thing (TM): CMake has odd visibility rules for imported library targets. They are only visible in the directory they have been `add_library()`-ed in. This means that if we `find_package(Abseil)` from multiple places in a single project, the previously added `standalone_absl` target won't be seen, so we will enter the if in `FindAbseil.cmake` multiple times. However, `ExternalProject_Add` will crap out because it already added a project called `abseil` (which _is_ globally visible).

Solution: add [`GLOBAL`](https://cmake.org/cmake/help/v3.12/command/add_library.html#imported-libraries) to `add_library(standalone_absl IMPORTED)`.

While I was at it, I added `find_package(Abseil)` to `cartographer-config.cmake.in`, so libcartographer users don't have to do it, and did a small cleanup of quiet/required stuff. For `FindAbseil.cmake` to be visible from `cartographer-config.cmake`, I had to add the installed Cartographer cmake modules to CMAKE_MODULE_PATH.


pifon2a:
@ojura thank you again

PR #1353, opened 2018-07-30
Fill submaps slices

gaschler:
This implements a frequently used function (twice in cartographer_ros).
Also, it corrects a wrong repeated proto field access,
so we can fix cartographer_ros/issues/944.

PR #1356, opened 2018-07-31
Use absl::Mutex instead of custom implementation

CodeArno:
Makes `common::Mutex` an alias for `absl::Mutex` and replaces the logic inside `common::MutexLocker` with `absl::MutexLock`.

A future PR can then remove these classes entirely and replace references to it across the Cartographer code base with the Abseil classes directly.


CodeArno:
@wally-the-cartographer merge

wally-the-cartographer:
Merge requested by authorized user CodeArno. Merge queue now has a length of 2.

wally-the-cartographer:
Error: exit status 1

Log:
~~~
=> handle work item: PR 1356 on googlecartographer/cartographer.
 => Switching to branch master.
 $ git reset --hard HEAD [data/cartographer]
HEAD is now at c1fbb6bb Fill submaps slices (#1353)
 $ git clean -fd [data/cartographer]
 $ git checkout master [data/cartographer]
Already on 'master'
Your branch is up to date with 'origin/master'.
 $ git pull [data/cartographer]
Your configuration specifies to merge with the ref 'refs/heads/master'
from the remote, but no such ref was fetched.

~~~

pifon2a:
@CodeArno please, rebase

CodeArno:
@pifon2a Do you mean rebase instead of merge or did you just mean that I should update this before submitting again?

I've finally managed to make this compile with CMake. @wohe gave me the hint that the CMake dependencies are order dependent which is why it failed. Submitting now.

CodeArno:
@wally-the-cartographer merge

wally-the-cartographer:
Merge requested by authorized user CodeArno. Merge queue now has a length of 1.

pifon2a:
@CodeArno I meant any of those. Rebase or merge, whatever gets it submitted :). Thanks for fixing the CMake build.

ojura:
> CMake dependencies are order dependent

This is incorrect. When you link against a CMake target, all its transitive dependencies get linked in the right order, so the order of `target_link_libraries()` does not matter.

However, in `FindAbseil.cmake`, we define an imported target which is just a bunch of externally built static libraries. Since these libraries are built as an external CMake project, Cartographer's CMake does not know how they depend one on another. When doing linking in general, the order matters, since the linker only looks for the currently missing symbols. E.g. if `a.o` depends on `b.o` which depends on `c.o`, you would call `your_favorite_linker a.o b.o c.o`.

When CMake knows the dependency DAG, it takes care of linking in the correct order for you. However, if you just throw a bunch of static libraries at it like we do, you have to specify them in the correct linking order, just as if you were linking them manually.

Issue #947, opened 2018-03-01
Wrong point cloud origin when accumulating data of different sensors


fabln:
Hi,
I'm going to map with four LIDARs which have some position separation. The occupancy grid (see screenshot) then gets some free-space lines drawn between e.g. sensor 1 (the upper right) and the orange data points measured by sensor 4 (the sensor most left). I think this shouldn't be the case, i.e. only free-space lines between each sensor and its own measured point cloud. 

This is observed with setting `TRAJECTORY_BUILDER_2D.num_accumulated_range_data >=1`. With `TRAJECTORY_BUILDER_2D.num_accumulated_range_data = 1`, the behavior is correct, however scan matching quality is lowered then. I tested only the 2D version.
![multi-lidar-map](https://user-images.githubusercontent.com/36953371/36839583-c2b52d84-1d42-11e8-993f-685fb362d0a8.png)

gaschler:
I believe your urdf config is incorrect.
After you fixed the urdf config, you should accumulate as many range data sets as you have sensors (or even multiple thereof if the packets to do contain a full scanner revolution.)

fabln:
Well, the sensor positions are fairly OK as set in the urdf (I did not yet calibrate the sensor positions. It seems that point clouds of multiple measurements are accumulated together and then joined with the map assuming that all data are originating from the same origin which is not the case.

gaschler:
No, cartographer_ros queries tf and transforms the data to the tracking frame.

For more help, please give more details on your config following the issue template https://raw.githubusercontent.com/googlecartographer/cartographer_ros/master/.github/ISSUE_TEMPLATE

fabln:
I cloned the master branch on 2018-02-28 and executed with the default values of your backpack 2D demo. Running the `cartographer_rosbag_validate` returned only timing issues not important so far. 

With the 4 LIDARs (in this case, onboard a car just standing in the garage), I would expect a map like this (couldn't tune to get this, so I photoshopped): 
![multi-lidar-map-expect](https://user-images.githubusercontent.com/36953371/36847559-007aa3da-1d5f-11e8-8e81-0de999aabfe3.png)

gaschler:
Cartographer makes the occupancy grid insertions by rays from the return point to the single tracking frame origin.
(In 3D, the insertion ray starts at the return point and is very short.)
So, what you are describing I would call a feature request to insert rays to their respective sensor positions.
[...]

PR #1357, opened 2018-07-31
Introduce [Timed]RangefinderPoint.

ojura:
RFC: https://github.com/googlecartographer/rfcs/pull/36


ojura:
So, this PR is about getting rid of direct uses of `Vector3/4` throughout the Cartographer codebase and using a new class `[Timed]RangefinderPoint` instead. Also, I made sure not to interrupt the data flow of points - wherever we recalculate the position inside `RangefinderPoint`, the structure itself is preserved. This enables libcartographer users to extend `RangefinderPoint` with their own stuff and have that additional metadata survive the entire local SLAM process (so we get the same `RangefinderPoint`s inside a `LocalSlamResult`).

cartographer/sensor/range_data.cc
```diff
@@ -58,30 +58,25 @@ proto::RangeData ToProto(const RangeData& range_data) {
...
-      static_cast<Eigen::Vector3f (*)(const transform::proto::Vector3f&)>(
-          transform::ToEigen));
+  std::copy(proto.returns().begin(), proto.returns().end(),
+            std::back_inserter(returns));
   PointCloud misses;
   misses.reserve(proto.misses().size());
-  std::transform(
-      proto.misses().begin(), proto.misses().end(), std::back_inserter(misses),
-      static_cast<Eigen::Vector3f (*)(const transform::proto::Vector3f&)>(
-          transform::ToEigen));
```
ojura: I'd say that this monstrosity disappearing is the highlight of this PR

gaschler:
@ojura Please consider presenting your design in the open house (today or in two weeks) to make sure implementation goes in the right direction.

ojura:
To address some concerns from yesterday:
Performance: 
30 runs of `roslaunch cartographer_ros offline_backpack_2d.launch bag_filenames:=b2-2016-04-05-14-44-52.bag no_rviz:=true`
with `[Timed]RangeFinderPoint`: 64.64 s, σ = 0.2
master: 65.06 s, σ = 0.31

I believe this change does not increase complexity and that it makes code more readable. It is made clear that a rangefinder point is being passed, and not just any `Vector3f`. Also, I think `.time()` is more readable than accessing the fourth element (`[3]`) of a `Vector4f`.

ojura:
@pifon2a @gaschler @cschuet friendly ping, PTAL?

cartographer/io/points_batch.h
```diff
@@ -52,7 +53,7 @@ struct PointsBatch {
   int trajectory_id;
 
   // Geometry of the points in the map frame.
-  std::vector<Eigen::Vector3f> points;
+  cartographer::sensor::PointCloud points;
```
pifon2a: nit: just `sensor::PointCloud`.

cartographer/sensor/point_cloud.h
```diff
@@ -29,14 +30,14 @@ namespace sensor {
 
 // Stores 3D positions of points.
 // For 2D points, the third entry is 0.f.
-typedef std::vector<Eigen::Vector3f> PointCloud;
+typedef std::vector<RangefinderPoint> PointCloud;
```
pifon2a: nit:
`using PointCloud = std::...` here and below

cartographer/sensor/proto/sensor.proto
```diff
@@ -20,6 +20,15 @@ option java_outer_classname = "Sensor";
 
 import "cartographer/transform/proto/transform.proto";
 
+message RangefinderPoint {
+  transform.proto.Vector3f position = 1;
```
pifon2a: I am not sure I like 'position' name here. In ROS it was called just a [point](http://docs.ros.org/melodic/api/sensor_msgs/html/msg/PointCloud.html).

cartographer/sensor/proto/sensor.proto
```diff
@@ -20,6 +20,15 @@ option java_outer_classname = "Sensor";
...
+}
+
+message TimedRangefinderPoint {
+  RangefinderPoint rangefinder_point = 1;
```
pifon2a: inline the type: `transform.proto.Vector3f point = 1;` . It can happen that `RangefinderPoint` has some metadata that we do not really want in `TimedRangefinderPoint`.

cartographer/sensor/proto/sensor.proto
```diff
@@ -30,14 +39,14 @@ message CompressedPointCloud {
 message TimedPointCloudData {
   int64 timestamp = 1;
   transform.proto.Vector3f origin = 2;
-  repeated transform.proto.Vector4f point_data = 3;
+  repeated TimedRangefinderPoint point_data = 3;
```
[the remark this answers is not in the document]
pifon2a: yes, please do that. Let's not break it for now. We will have plenty of opportunities to do so :)

cartographer/sensor/range_data.cc
pifon2a: I cannot stop being surprised that we have such code in the first place... :)

cartographer/sensor/timed_point_cloud_data.h
```diff
@@ -32,7 +32,7 @@ struct TimedPointCloudData {
 
 struct TimedPointCloudOriginData {
   struct RangeMeasurement {
-    Eigen::Vector4f point_time;
+    sensor::TimedRangefinderPoint point_time;
```
pifon2a: omit `sensor::`.

cartographer/sensor/rangefinder_point.h
```diff
@@ -0,0 +1,98 @@
...
+
+#include "Eigen/Core"
+#include "cartographer/sensor/proto/sensor.pb.h"
+#include "cartographer/transform/rigid_transform.h"
+#include "glog/logging.h"
+
+namespace cartographer {
+namespace sensor {
+
+// Stores 3D position of a point observed by a rangefinder sensor.
```
pifon2a: 1. Please, use composition here. 
2. All data members should be in `private` section of the class. 
3. Try to get rid of friend functions.

cartographer/sensor/proto/sensor.proto
ojura: Okay, should the C++ implementation follow this as well? I.e. `TimedRangefinderPoint` contains a `Vector3f`?

cartographer/sensor/rangefinder_point.h
ojura: About the friend `operator*`: we typically apply a multiplication with a `Rigid3` pose estimate from the left. Unfortunately, a member `operator*` only allows `this` as the left operand. So I see two solutions: either an overloaded friend `operator*` like this, or a member function e.g. `RangefinderPoint::ToLocal(Rigid3 pose_estimate)` which would do a left multiplication.

cartographer/sensor/proto/sensor.proto
ojura: Sure. We're not bound by ROS conventions here, though. A `RangefinderPoint` having a `point` member sounds kind of odd to me.

cartographer/sensor/rangefinder_point.h
ojura: Also, about composition: you mentioned you would inline the `Vector3f` in `TimedRangefinderPoint` proto. Does this mean that the C++ class should be the same?

Slicing a `TimedRangefinderPoint` into a `RangefinderPoint` then means a little bit of more work. Should I put it in a `TimedRangefinderPoint::ToRangefinderPoint` member?

cartographer/sensor/proto/sensor.proto
pifon2a: When I see `position` I always confuse it with `pose` which also has orientation. What about `x_y_z` or `coordinates`? Or leave it as `position`.

cartographer/sensor/proto/sensor.proto
pifon2a: Yep

cartographer/sensor/rangefinder_point.h
pifon2a: Make them structs. As in proto. 

```
struct TimedRangefinderPoint {
  proto::RangefinderPoint ToProto() const;
  Eigen::Vector3f position_;
  float time_;
};

struct RangefinderPoint {
  proto::RangefinderPoint ToProto() const;
  Eigen::Vector3f position_;
};
```

ojura: Isn't `_` only for private members? Also, I think structs shouldn't have methods, IIRC `ToProto/FromProto` are implemented as functions outside structs.

pifon2a: In structs, you do not need `_`, sorry, i just copied. Structs should not have complicated logic, yes. But serialization/deserialization can be put there. Up to you.

cartographer/sensor/proto/sensor.proto
ojura: At the robotics classes at my uni, we were taught the following definition of a robot [pose](https://en.wikipedia.org/wiki/Pose_(computer_vision)): _The combination of position and orientation is referred to as the pose of an object_.

So if that's okay with you, I'd leave it as position :-)

pifon2a: Yeah, it is ok. :) I did not have robotics classes :)

cartographer/sensor/rangefinder_point.h
ojura: Okay, cool. I'm only missing one more detail: where to implement discarding time? Member `TimedRangefinderPoint::ToRangefinderPoint` or standalone function?

ojura:
Okay, the nits should be fixed. I updated the two new classes to be structs. `[Range|TimedPointCloud]Data::FromProto` should be able handle old .pbstreams.

cartographer/sensor/range_data.cc
```diff
@@ -58,30 +58,37 @@ proto::RangeData ToProto(const RangeData& range_data) {
...
-      proto.returns().begin(), proto.returns().end(),
-      std::back_inserter(returns),
-      static_cast<Eigen::Vector3f (*)(const transform::proto::Vector3f&)>(
-          transform::ToEigen));
+  if (proto.returns_size() > 0) {
+    returns.reserve(proto.returns().size());
+    for (const auto& point_proto : proto.returns())
+      returns.push_back(FromProto(point_proto));
+  } else {
+    returns.reserve(proto.returns_legacy().size());
```
ojura: This should take care of deserializing old .pbstreams

cartographer/sensor/rangefinder_point.h
```diff
@@ -0,0 +1,89 @@
...
+                       const RangefinderPoint& rhs) {
+  return lhs.position == rhs.position;
+}
+
+inline bool operator==(const TimedRangefinderPoint& lhs,
+                       const TimedRangefinderPoint& rhs) {
+  return lhs.position == rhs.position && lhs.time == rhs.time;
+}
+
+RangefinderPoint FromProto(
```
ojura: @pifon2a WDYT about inlining all of these? Point processing functions seem like they could cause a bottleneck if not inlined.

cartographer/sensor/rangefinder_point.h
```diff
@@ -0,0 +1,108 @@
...
+};
+
+// Stores 3D position of a point with its relative measurement time.
+// See point_cloud.h for more details.
+struct TimedRangefinderPoint {
+  Eigen::Vector3f position;
+  float time;
+};
+
+template <class T>
```
pifon2a: since there is no metadata yet, I doubt that the operators*() are needed at all.

cartographer/sensor/rangefinder_point.h
```diff
@@ -0,0 +1,108 @@
...
+inline proto::TimedRangefinderPoint ToProto(
+    const TimedRangefinderPoint& timed_rangefinder_point) {
+  proto::TimedRangefinderPoint proto;
+  *proto.mutable_position() =
+      transform::ToProto(timed_rangefinder_point.position);
+  proto.set_time(timed_rangefinder_point.time);
+  return proto;
+}
+
+inline RangefinderPoint ToRangefinderPoint(
```
pifon2a: ToRangefinderPoint/ToTimedRangefinderPoint are also not needed.

cartographer/sensor/timed_point_cloud_data.cc
```diff
@@ -28,20 +28,25 @@ proto::TimedPointCloudData ToProto(
...
-  timed_point_cloud.reserve(proto.point_data().size());
-  std::transform(
-      proto.point_data().begin(), proto.point_data().end(),
-      std::back_inserter(timed_point_cloud),
-      static_cast<Eigen::Vector4f (*)(const transform::proto::Vector4f&)>(
-          transform::ToEigen));
+  if (proto.point_data().size() > 0) {
+    timed_point_cloud.reserve(proto.point_data().size());
+    for (const auto& timed_point_proto : proto.point_data())
+      timed_point_cloud.push_back(FromProto(timed_point_proto));
```
pifon2a: nit: unfortunately, style guide requires {} here.

cartographer/mapping/internal/2d/local_trajectory_builder_2d.cc
```diff
@@ -166,20 +166,23 @@ LocalTrajectoryBuilder2D::AddRangeData(
...
   for (size_t i = 0; i < synchronized_data.ranges.size(); ++i) {
-    const Eigen::Vector4f& hit = synchronized_data.ranges[i].point_time;
+    const sensor::TimedRangefinderPoint& hit =
+        synchronized_data.ranges[i].point_time;
     const Eigen::Vector3f origin_in_local =
         range_data_poses[i] *
         synchronized_data.origins.at(synchronized_data.ranges[i].origin_index);
-    const Eigen::Vector3f hit_in_local = range_data_poses[i] * hit.head<3>();
-    const Eigen::Vector3f delta = hit_in_local - origin_in_local;
+    sensor::RangefinderPoint hit_in_local =
```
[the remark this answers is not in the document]
ojura: I disagree with this and the comment below. The whole point of this PR is to enable adding metadata by changing only `rangefinder_point.h`. The rest of Cartographer should be written with an assumption that there _is_ some hidden metadata in `RangefinderPoint`. This metadata should be preserved throughout the SLAM pipeline. By constructing a completely new point here, and not using `operator*`, we would discard this data.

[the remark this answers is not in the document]
pifon2a: Could you provide an example of such metadata? And when it will be added to Cartographer.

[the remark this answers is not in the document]
ojura: Please see the comment below. My primary motivation for this is keeping the laser ID (i.e. which laser of the 16/32/etc) observed the point. Currently, I do not plan on upstreaming this, but I would like to be able to do this with minimal intervetions to Cartographer, hence this PR.

Another possible metadata that we might want to add is the sensor ID, to help resolve issues like #947.

ojura:
@pifon2a Have I managed to convince you? Ready to merge?

pifon2a:
Juraj, I want to take another look and then we merge today. Let me finish
some stuff first.
--
Cheers,
Alex

Issue #1360, opened 2018-08-02
Constraint search segfaults in uplink server


gaschler:
How to reproduce:

Run a distributed system where an uplink grpc server does constraint search.
The segfault is more likely when constraint search is tuned up, the trajectory is short, and the work queue is short.
Constraint search will segfault in the scan matcher creation (because it accesses a nullptr grid).

Root cause:

The "active" submaps in the uplink don't have a grid but participate in constraint search.

PR #1362, opened 2018-08-02
Skip constraint search against unfinished submaps.

gaschler:
The uplink server only receives the grid content of a submap after
that submap is finished for efficiency. Therefore, constraint
searches against that submap need to be skipped.
Also add checks to avoid this in the future.

FIXES=#1360


gaschler:
@MichaelGrupp This should fix #1360, could you please verify?

MichaelGrupp:
Will test.

SirVer:
Defer to @wohe. Not doing SLAM anymore these days.

gaschler:
@wohe Do you see any unexpected side effects on the constraint search when ignoring submaps with incomplete range data insertion? (I believe this didn't happen in single-process SLAM, so there shouldn't be side effects.)

gaschler:
@wally-the-cartographer merge

wally-the-cartographer:

PR #1369, opened 2018-08-06
[ABSL] Purge common/mutex.h.

pifon2a:


CodeArno:
Thanks! I was about to do the same actually, but then shied away from it for now because other projects (including `cartographer_ros`) depend on `mutex.h` in Cartographer. Have you adapted `cartographer_ros` as well and are you aware of any other projects like that?

pifon2a:
I will adapt cartographer_ros. and submit two CL simultaneously.

PR #1379, opened 2018-08-09
Support gRPC requests for loading unfrozen states.

MichaelGrupp:


cartographer/mapping/internal/optimization/optimization_problem_2d.cc
```diff
@@ -256,7 +256,8 @@ void OptimizationProblem2D::Solve(
   }
   // Add cost functions for landmarks.
   AddLandmarkCostFunctions(landmark_nodes, freeze_landmarks, node_data_,
-                           &C_nodes, &C_landmarks, &problem, options_.huber_scale());
```
MichaelGrupp: No idea why this is formatted just now ... not related to this PR

MichaelGrupp:
@cschuet @gaschler

MichaelGrupp:
There's a `LoadStateAndDelete` test in `cloud/internal/client_server_test.cc

pifon2a:
Yes, does it test your change?

MichaelGrupp:
Added. If there's anything, I will be back in 2 weeks ;)

cartographer/cloud/internal/client_server_test.cc
```diff
@@ -636,6 +636,56 @@ TEST_P(ClientServerTestByGridType, LoadStateAndDelete) {
...
+      ProtoReaderFromStrings(kSerializationHeaderProtoString,
+                             {
+                                 kPoseGraphProtoString,
+                                 kAllTrajectoryBuilderOptionsProtoString,
+                                 kSubmapProtoString,
+                                 kNodeProtoString,
+                                 kTrajectoryDataProtoString,
+                                 kImuDataProtoString,
+                                 kOdometryDataProtoString,
+                                 // TODO: not implemented by 2D pose graph.
```
MichaelGrupp: Not sure about this one. Loading unfrozen state with fixed frame pose data leads to a check fail "not implemented":
```
E0809 11:43:14.039064  3494 pose_graph_2d.cc:694] not implemented
F0809 11:43:14.039091  3494 pose_graph_2d.cc:216] Not yet implemented for 2D.
```

cschuet: Fixed frame pose is not yet implemented in 2D. So we would never produce a proto in 2D that constrains that data. Same thing with the kTrajectoryDataProtoString: no proto produced by the 2D pipeline would ever contain such an item. I would therefore just leave these out in the test.

MichaelGrupp: Ok. They are however also used in the other LoadStateAndDelete test and there it doesn't crash.

Feel free to remove it if you want to merge, otherwise I'll do it in 2 weeks

PR #1382, opened 2018-08-10
[GenericPoseGraph] Add a macro to find nodes.

pifon2a:


cartographer/pose_graph/constraint/constraint_utils.h
```diff
@@ -25,6 +25,21 @@
 namespace cartographer {
 namespace pose_graph {
 
+#define FIND_NODE(node_name, node_id, map, log_message)                   \
```
mbokeloh: Maybe rename to FIND_NODE_OR_RETURN

cartographer/pose_graph/constraint/constraint_utils.h
```diff
@@ -25,6 +25,21 @@
...
+  FIND_NODE_IMPL(STATUS_MACROS_IMPL_CONCAT_(__node, __LINE__), node_name, \
+                 node_id, map, log_message)
+
+#define FIND_NODE_IMPL(node_item, node_name, node_id, map, log_message) \
+  auto node_item = common::FindOrNull(map, node_id);                    \
```
mbokeloh: why not direclty?:
auto node_name = common::FindOrNull(map, node_id);

Then you don't need these concatenations and extra macros

cartographer/pose_graph/constraint/constraint_utils.h
mbokeloh: You could also pull the type out, so that it's clearer form the outside what is happening:
FIND_NODE_OR_RETURN(auto my_node, ...);

cartographer/pose_graph/constraint/constraint_utils.h
```diff
@@ -31,6 +31,13 @@ void AddPose3D(Pose3D* pose, ceres::Problem* problem);
 
 void AddImuCalibration(ImuCalibration* pose, ceres::Problem* problem);
 
+#define FIND_NODE_OR_RETURN(node_name, node_id, map, log_message) \
```
ojura: Since this is defined in a public header, maybe prefix the macro with `CARTOGRAPHER_`?

This is the first time that Cartographer defines a macro for something, though. Are you sure the readability is worth doing this?

pifon2a: Later (end of q3), when absl has statusor lib, we will migrate and remove this code.

PR #1384, opened 2018-08-10
Rename submap finished flags

gaschler:
A submap has two separate bool finished states:
1. 'Submap::finished()' whether all range data has been inserted.
2. 'SubmapState', used by 'PoseGraph', which becomes true slightly
later after pose graph elements have been created and the submap
should take part in constraint search.

This is confusing. Serialization knows only one such state.

So, we rename
1. finished => insertion_finished
2. kActive => kNoConstraintSearch


pifon2a:
@wally-the-cartographer merge

wally-the-cartographer:

ojura:
CI failures due to Debian servers being unavailable are pretty common these days. Maybe Wally should try re-running CI if that happens?

PR #1387, opened 2018-08-11
Use FetchContent to download and configure Abseil at configure time

ojura:
I stumbled across a cool CMake module called `FetchContent`:
https://cmake.org/cmake/help/git-stage/module/FetchContent.html

It is very similar to `ExternalProject`, but it handles downloading dependencies at CMake configure time. Thus we can `add_subdirectory()` Abseil [just like the official documentation recommends](https://github.com/abseil/abseil-cpp/blob/master/CMake/README.md). CMake then knows about all Abseil targets and how they depend on each other since they're built in the same project, and we can link against them in a sane manner without having to enumerate all those static libraries in the right order. This also works much better with multi-configuration build systems such as MSVC since we properly propagate the build type.

The only downside is that this module was introduced very recently, in CMake 3.11. However, it's trivial to backport it to older CMake versions by just providing a copy, which I did in this PR.

@pifon2a

### DIFF
--- a/cartographer/io/probability_grid_points_processor.cc
+++ b/cartographer/io/probability_grid_points_processor.cc
@@ -36,14 +36,15 @@ void DrawTrajectoriesIntoImage(
     const std::vector<mapping::proto::Trajectory>& trajectories,
     cairo_surface_t* cairo_surface) {
   for (size_t i = 0; i < trajectories.size(); ++i) {
-    DrawTrajectory(trajectories[i], GetColor(i),
-                   [&probability_grid,
-                    &offset](const transform::Rigid3d& pose) -> Eigen::Array2i {
-                     return probability_grid.limits().GetCellIndex(
-                                pose.cast<float>().translation().head<2>()) -
-                            offset;
-                   },
-                   cairo_surface);
+    DrawTrajectory(
+        trajectories[i], GetColor(i),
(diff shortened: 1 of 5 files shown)


ojura:
@pifon2a WDYT?

MichaelGrupp:
@ojura  @pifon2a This looks like an useful addition, any updates on this? Besides removing the warnings, I think you need to declare the dependency for the flat hash maps that were added recently to make it compile without errors.

pifon2a:
@MichaelGrupp yep, https://github.com/Mizux/abseil-cpp/commit/f18b05de3a148c4c1c8d46703ad13f33e7c76cfd

ojura:
@MichaelGrupp thanks for the hint. I added linking against `absl::hash` and it builds. But it looks like that @pifon2a is waiting for Abseil to become installable :)

wohe:
Closing in favor of #1570.

ojura:
@wohe if we’re not providing support for building abseil in Cartographer, we will require installing it as a dependency?

wohe:
@ojura The change leaves it to the user to decide how they want to satisfy the dependency. You can for example use stow and run the script. Another possibility with ROS is to add a catkin-aware version of Abseil and let catkin build it.

PR #1398, opened 2018-08-24
Fix gravity alignment of 3D submaps in local SLAM

danielsievers:
We were passing the gravity estimate of the current tracking frame
to intialize the local submap pose. Fixing this improves the alignment
of submaps in the global (and approx. gravity-aligned) frame.

### DIFF
--- a/cartographer/mapping/3d/submap_3d.cc
+++ b/cartographer/mapping/3d/submap_3d.cc
@@ -253,12 +253,13 @@ void Submap3D::ToResponseProto(
                     response->add_textures());
 }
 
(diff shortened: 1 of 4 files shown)


danielsievers:
Here is some data from running the offline node over the museum 3d data (b3-2016-04-05-14-14-00.bag) and then looking at the angle between unit z and the submap pose's z-axis in the global frame after final optimization for all submaps (values in degrees):

before:
Count: 136  Min: 2.665955  Max: 20.354956  Mean: 9.984844

after:
Count: 136  Min: 0.103457  Max: 2.117294  Mean: 0.874563

PR #1402, opened 2018-08-28
Configurable CC/CXX compiler toolchain in Dockerfiles.

MichaelGrupp:
Allows to build Docker images with Clang.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,14 +22,15 @@ cache:
     - /home/travis/docker/
 
 env:
-  - LSB_RELEASE=trusty DOCKER_CACHE_FILE=/home/travis/docker/trusty-cache.tar.gz
-  - LSB_RELEASE=xenial DOCKER_CACHE_FILE=/home/travis/docker/xenial-cache.tar.gz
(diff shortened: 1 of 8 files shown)


MichaelGrupp:
Todo: `-Wthread-safety` is not yet set as it seems?

MichaelGrupp:
* can build using Clang and without libc++
* building with Clang and libc++ leads to ABI compatibility issues with Boost packages (and probably more)

pifon2a:
@MichaelGrupp btw, boost packages can and should be removed from libcartographer. Only `boost//:iostreams` package is used to get gzip.

pifon2a:
@wally-the-cartographer merge

wally-the-cartographer:

wally-the-cartographer:
Refusing to merge, since the combined checks state is failure

MichaelGrupp:
I set the default compiler for Travis builds to GCC again, Clang seems to be a bit brittle with some dependencies sometimes and I don't want to break the merge queue.

But we can leave the compiler option for testing.

pifon2a:
@wally-the-cartographer merge

PR #1403, opened 2018-08-29
Enable -Wthread-safety for Clang and CMake.

MichaelGrupp:


MichaelGrupp:
Related to, but independent from #1402 

The CMake compiler constant is documented here: https://cmake.org/cmake/help/v3.2/variable/CMAKE_LANG_COMPILER_ID.html

MichaelGrupp:
> This is also called by cartographer_ros, have you tested that it also compiles there?

It compiles, and found an error:
https://github.com/googlecartographer/cartographer_ros/pull/1068

PR #1405, opened 2018-08-29
Store histogram in submaps: prepare for backward compatiblity

schwoere:
This makes the map builder backwards compatible to the current map pbstream (version 1). The PR prepares for #1277, where pbstream version 2 will be introduced. Backwards compatibility was discussed in #1277.

When a map with pbstream version 1 is loaded, a rotational scan matcher histogram is generated for each submap using the histograms of all nodes that were inserted to the submap during local SLAM. Once this backwards compatibility is in place, I would like to introduce the new format with #1277.

### DIFF
--- a/cartographer/io/serialization_format_migration.cc
+++ b/cartographer/io/serialization_format_migration.cc
@@ -309,7 +309,7 @@ MigrateSubmapFormatVersion1ToVersion2(
     return submap_id_to_submap;
   }
 
-  std::map<SubmapId, std::unique_ptr<Submap3D>> deserialized_submaps;
+  MapById<SubmapId, proto::Submap> migrated_submaps = submap_id_to_submap;
   for (const proto::PoseGraph::Constraint& constraint_proto :
        pose_graph_proto.constraint()) {
     if (constraint_proto.tag() == proto::PoseGraph::Constraint::INTRA_SUBMAP) {
@@ -321,30 +321,40 @@ MigrateSubmapFormatVersion1ToVersion2(
       const Eigen::VectorXf& rotational_scan_matcher_histogram_in_gravity =
           node_data.rotational_scan_matcher_histogram;
 
-      const float local_yaw_from_gravity =
-          transform::GetYaw(node_data.local_pose.rotation() *
-                            node_data.gravity_alignment.inverse());
-      const Eigen::VectorXf rotational_scan_matcher_histogram_in_local =
-          scan_matching::RotationalScanMatcher::RotateHistogram(
-              rotational_scan_matcher_histogram_in_gravity,
-              local_yaw_from_gravity);
       SubmapId submap_id{constraint_proto.submap_id().trajectory_id(),
                          constraint_proto.submap_id().submap_index()};
-      CHECK(submap_id_to_submap.Contains(submap_id));
-      const proto::Submap& submap_proto = submap_id_to_submap.at(submap_id);
-      CHECK(submap_proto.has_submap_3d());
-      if (deserialized_submaps.find(submap_id) == deserialized_submaps.end()) {
-        deserialized_submaps[submap_id] =
-            absl::make_unique<Submap3D>(submap_proto.submap_3d());
+      CHECK(migrated_submaps.Contains(submap_id));
+      proto::Submap& migrated_submap_proto = migrated_submaps.at(submap_id);
+      CHECK(migrated_submap_proto.has_submap_3d());
+
+      mapping::proto::Submap3D* submap_3d_proto =
+          migrated_submap_proto.mutable_submap_3d();
+
+      const double submap_yaw_from_gravity = transform::GetYaw(
+          transform::ToRigid3(submap_3d_proto->local_pose()).rotation() *
+          node_data.local_pose.rotation() *
+          node_data.gravity_alignment.inverse());
+      const Eigen::VectorXf rotational_scan_matcher_histogram_in_submap =
+          scan_matching::RotationalScanMatcher::RotateHistogram(
+              rotational_scan_matcher_histogram_in_gravity,
+              submap_yaw_from_gravity);
+
+      if (submap_3d_proto->rotational_scan_matcher_histogram_size() == 0) {
+        for (Eigen::VectorXf::Index i = 0;
+             i != rotational_scan_matcher_histogram_in_submap.size(); ++i) {
+          submap_3d_proto->add_rotational_scan_matcher_histogram(
+              rotational_scan_matcher_histogram_in_submap(i));
+        }
+      } else {
+        auto submap_histogram =
+            submap_3d_proto->mutable_rotational_scan_matcher_histogram();
+        for (Eigen::VectorXf::Index i = 0;
+             i != rotational_scan_matcher_histogram_in_submap.size(); ++i) {
+          *submap_histogram->Mutable(i) +=
+              rotational_scan_matcher_histogram_in_submap(i);
+        }
       }
-      Submap3D* const submap_3d = deserialized_submaps.at(submap_id).get();
-      submap_3d->AddScanHistogram(rotational_scan_matcher_histogram_in_local);
     }
-  }
-  MapById<SubmapId, proto::Submap> migrated_submaps = submap_id_to_submap;
-  for (const auto& submap : deserialized_submaps) {
-    migrated_submaps.at(submap.first) =
-        submap.second->ToProto(submap.second->insertion_finished());
   }
   return migrated_submaps;
 }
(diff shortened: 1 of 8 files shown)


cartographer/io/serialization_format_migration_test.cc
```diff
@@ -100,6 +102,63 @@ class MigrationTest : public ::testing::Test {
...
+class SubmapHistogramMigrationTest : public ::testing::Test {
+ protected:
+  void SetUp() {
+    CreateNode();
+    CreateSubmap();
+    CreatePoseGraphWithNodeToSubmapConstraint();
+  }
+
+ private:
+  void CreateSubmap() {
```
danielsievers: Can you use CreateFakeSubmap3D() from mapping/internal/testing/test_helpers.h instead of adding this?
There is also CreateFakeNode(). Or is that header internal to mapping/*?

cartographer/io/serialization_format_migration_test.cc
```diff
@@ -175,6 +234,41 @@ TEST_F(MigrationTest, SerializedDataOrderAfterGridMigrationIsCorrect) {
...
+                          node_.node_id().node_index());
+  node_id_to_node.Insert(node_id, node_);
+
+  const auto submap_id_to_submap_migrated =
+      MigrateSubmapFormatVersion1ToVersion2(submap_id_to_submap,
+                                            node_id_to_node, pose_graph_);
+
+  EXPECT_EQ(submap_id_to_submap_migrated.size(), submap_id_to_submap.size());
+  const mapping::proto::Submap& migrated_submap =
+      submap_id_to_submap_migrated.at(submap_id);
```
danielsievers: Would the code from below this line be simpler if we just did something like:

mapping::proto::Submap expected_submap = submap_;
// then set the histogram field accordingly
EXPECT_THAT(migrated_submap, EqualsProto(expected_submap));

This would also test that no other fields got screwed up.
wdyt?

cartographer/io/serialization_format_migration.cc
```diff
@@ -295,5 +297,56 @@ void MigrateStreamFormatToVersion1(
...
+    return submap_id_to_submap;
+  }
+
+  std::map<SubmapId, std::unique_ptr<Submap3D>> deserialized_submaps;
+  for (const proto::PoseGraph::Constraint& constraint_proto :
+       pose_graph_proto.constraint()) {
+    if (constraint_proto.tag() == proto::PoseGraph::Constraint::INTRA_SUBMAP) {
+      NodeId node_id{constraint_proto.node_id().trajectory_id(),
+                     constraint_proto.node_id().node_index()};
+      const TrajectoryNode::Data& node_data =
```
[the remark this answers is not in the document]
schwoere: yes, done!

cartographer/io/serialization_format_migration.cc
```diff
@@ -295,5 +297,56 @@ void MigrateStreamFormatToVersion1(
...
+      const float local_yaw_from_gravity =
+          transform::GetYaw(node_data.local_pose.rotation() *
+                            node_data.gravity_alignment.inverse());
+      const Eigen::VectorXf rotational_scan_matcher_histogram_in_local =
+          scan_matching::RotationalScanMatcher::RotateHistogram(
+              rotational_scan_matcher_histogram_in_gravity,
+              local_yaw_from_gravity);
+
+      SubmapId submap_id{constraint_proto.submap_id().trajectory_id(),
+                         constraint_proto.submap_id().submap_index()};
```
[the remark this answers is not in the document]
schwoere: yes, done!

cartographer/io/serialization_format_migration.cc
```diff
@@ -295,5 +297,56 @@ void MigrateStreamFormatToVersion1(
...
+      const proto::Submap& submap = submap_id_to_submap.at(submap_id);
```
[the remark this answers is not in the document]
schwoere: yes, done!

cartographer/io/serialization_format_migration.cc
```diff
@@ -295,5 +297,56 @@ void MigrateStreamFormatToVersion1(
...
+      CHECK(submap.has_submap_3d());
+      if (deserialized_submaps.find(submap_id) == deserialized_submaps.end()) {
+        std::unique_ptr<Submap3D> submap3d =
+            absl::make_unique<Submap3D>(submap.submap_3d());
+        deserialized_submaps[submap_id] = std::move(submap3d);
```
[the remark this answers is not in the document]
schwoere: yes, good point, thanks!
I added this, but then decided to remove the call to `Submap3D::AddScanHistogram`. Therefore, this step is not required anymore, as I am working only on the submap proto now.

cartographer/io/serialization_format_migration_test.cc
```diff
@@ -100,6 +102,63 @@ class MigrationTest : public ::testing::Test {
...
+    mapping::proto::PoseGraph::Constraint* constraint =
+        pose_graph_.add_constraint();
+    constraint->set_tag(mapping::proto::PoseGraph::Constraint::INTRA_SUBMAP);
+    *constraint->mutable_node_id() = node_.node_id();
+    *constraint->mutable_submap_id() = submap_.submap_id();
+    *constraint->mutable_relative_pose() =
+        transform::ToProto(transform::Rigid3d::Identity());
+
+    constraint = pose_graph_.add_constraint();
+    constraint->set_tag(mapping::proto::PoseGraph::Constraint::INTER_SUBMAP);
```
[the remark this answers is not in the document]
schwoere: okay, done!

cartographer/io/serialization_format_migration_test.cc
schwoere: Thanks for this hint! This allowed me to remove most of this code.

cartographer/io/serialization_format_migration_test.cc
schwoere: `EqualsProto` seemed to only work on string formatted protos. (see [relative_pose_constraint_3d_test.cc](https://github.com/googlecartographer/cartographer/blob/master/cartographer/pose_graph/constraint/relative_pose_constraint_3d_test.cc))

schwoere:
@cschuet what do you think about adding the backwards compatibility method of this PR, call it from map builder in a follow up PR and then introducing the version format as in #1277.

schwoere:
@danielsievers I added the migration of submaps to the map builder and addressed your comments. Could you have a look again?

cartographer/mapping/map_builder.cc
```diff
@@ -360,6 +364,21 @@ std::map<int, int> MapBuilder::LoadState(
     }
   }
 
+  // TODO(schwoere): Remove backwards compatibility once the pbstream format
+  // version 2 is established.
+  constexpr uint32_t kFormatVersionWithoutSubmapHistograms = 1;
+  if (deserializer.header().format_version() ==
```
danielsievers: You probably saw already, but looks like once you add version 2 we will have to modify IsVersionSupported() in ProtoStreamDeserializer to:
bool IsVersionSupported(const mapping::proto::SerializationHeader& header) {
  return header.format_version() == kFormatVersionWithoutSubmapHistograms ||
  return header.format_version() == kMappingStateSerializationFormatVersion;
 }

So should kFormatVersionWithoutSubmapHistograms move to mapping_state_serialization.h?
For now you can set it to kMappingStateSerializationFormatVersion.

schwoere: Yes, makes sense to me. I moved `kFormatVersionWithoutSubmapHistograms` to  `mapping_state_serialization.h.`

cartographer/io/serialization_format_migration.cc
```diff
@@ -295,5 +297,68 @@ void MigrateStreamFormatToVersion1(
                             output);
 }
 
+mapping::MapById<mapping::SubmapId, mapping::proto::Submap>
+MigrateSubmapFormatVersion1ToVersion2(
+    const mapping::MapById<mapping::SubmapId, mapping::proto::Submap>&
+        submap_id_to_submap,
+    mapping::MapById<mapping::NodeId, mapping::proto::Node>& node_id_to_node,
+    const mapping::proto::PoseGraph& pose_graph_proto) {
+  using namespace cartographer::mapping;
```
wohe: `using mapping;` if you want to be short, or `using ::cartographer::mapping;` for the full namespace?

cartographer/io/serialization_format_migration_test.cc
```diff
@@ -100,6 +103,44 @@ class MigrationTest : public ::testing::Test {
   static constexpr int kNumOriginalMessages = 9;
 };
 
+class SubmapHistogramMigrationTest : public ::testing::Test {
+ protected:
+  void SetUp() {
```
wohe: `override`?

schwoere:
@cschuet I am merging now. If there are any concerns for cloud parts, we can revert this PR or fix it in a follow up PR.

schwoere:
@wally-the-cartographer merge

wally-the-cartographer:
Merge requested by authorized user schwoere. Merge queue now has a length of 1.

PR #1410, opened 2018-09-04
Include all trajectory nodes when serializing without unfinished submaps.

MichaelGrupp:
Fixes a mismatch with the number of nodes of the trajectory node field
added in pbstream files.

Also, loading the serialized pose graph fails with an `out_of_range` exception
if "orphaned" nodes are not present.


MichaelGrupp:
TODO: check test

PR #1413, opened 2018-09-06
Return submap ids in InsertionResult along with pointers

ojura:
Returning just submap pointers in `InsertionResult` is not very useful for libcartographer users. You can't really tell which submap IDs changed when insertion happened. Therefore, have the pose graph return the assigned submap IDs just like it does for the inserted node ID.

Also, move `GetSubmapData` to the pose graph interface, so the user can query the pose graph just for these submap IDs (instead of having to use `GetAllSubmapData`).

### DIFF
--- a/cartographer/mapping/3d/hybrid_grid.h
+++ b/cartographer/mapping/3d/hybrid_grid.h
@@ -544,6 +544,32 @@ class HybridGrid : public HybridGridBase<uint16> {
   std::vector<ValueType*> update_indices_;
 };
 
+struct AverageIntensityData {
+  float sum = 0.f;
+  int count = 0;
+};
+
+class IntensityHybridGrid : public HybridGridBase<AverageIntensityData> {
+ public:
+  explicit IntensityHybridGrid(const float resolution)
+      : HybridGridBase<AverageIntensityData>(resolution) {}
+
+  void AddIntensity(const Eigen::Array3i& index, const float intensity) {
+    AverageIntensityData* const cell = mutable_value(index);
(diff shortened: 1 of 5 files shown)


ojura:
Ping?

ojura:
@pifon2a since we're on a roll, if I could choose a PR review to get as a Christmas present :), it would be this one. I have encountered multiple times so far a need to know which submaps exactly changed with a recent insertion, and I think this would be a really useful extension of the API.

pifon2a:
@ojura nice try :). PR review is not really a good Christmas present. I hope you get a normal present instead. Merry Christmas!

ojura:
Damn :) Merry Christmas and happy new year to y'all as well!

ojura:
Another month has passed, so here's another friendly ping for this PR.

wohe:
@ojura Are you still interested in getting this merged? If so, please let me know and I will review. Also, it seems this PR is doing two completely separate things. Would it be possible to split off the `GetSubmapData` change to a separate PR? Should be faster to get it merged then. For the other part: can you give some context how it is used, for example the use case you have in mind?

ojura:
Hi @wohe. One application is vertex coloring of scan matched point clouds according to the inserted submap ID (e.g. of the older one from the insertion pair). If you use that Golden ratio HSV scheme, I think it’s a nice way to provide false coloring for checking the cloud quality.

Sure, I can split the PR.

ojura:
@wohe here's a picture of coloring submap ID:
image

PR #1415, opened 2018-09-10
Fix LocalTrajectoryUploader recovery

gaschler:
Make sure no data is uploaded to an interrupted uplink trajectory,
but create a new uplink trajectory to upload to that one.

### DIFF
--- a/cartographer/cloud/internal/local_trajectory_uploader.cc
+++ b/cartographer/cloud/internal/local_trajectory_uploader.cc
@@ -183,6 +183,14 @@ void LocalTrajectoryUploader::TryRecovery() {
     }
   }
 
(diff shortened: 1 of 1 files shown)


cartographer/cloud/internal/local_trajectory_uploader.cc
```diff
@@ -183,6 +183,14 @@ void LocalTrajectoryUploader::TryRecovery() {
     }
   }
 
+  // Because the trajectories may be interrupted on the uplink side, we can no
```
MichaelGrupp: What do you mean by "interrupted" here?

gaschler: Interrupted means that the uplink does not have all sensor data (because the connection was interrupted)

PR #1417, opened 2018-09-10
Replace token file with Google auth in MapBuilder.

stevewolter:
This allows Cartographer to use normal Application Default Credentials
(ADC) instead of a hand-generated format.

Google-internally, we'll only use ADC going forward. At your option, I
can keep the old code for the token files around if you think it would
be generally useful. Many OSS programs follow the same approach.

### DIFF
--- a/cartographer/cloud/internal/local_trajectory_uploader.cc
+++ b/cartographer/cloud/internal/local_trajectory_uploader.cc
@@ -112,9 +112,9 @@ LocalTrajectoryUploader::LocalTrajectoryUploader(
   auto channel_creds =
       enable_google_auth
           ? grpc::GoogleDefaultCredentials()
-          : enable_ssl_encryption
-                ? ::grpc::SslCredentials(::grpc::SslCredentialsOptions())
-                : ::grpc::InsecureChannelCredentials();
+          : (enable_ssl_encryption
+                 ? ::grpc::SslCredentials(::grpc::SslCredentialsOptions())
+                 : ::grpc::InsecureChannelCredentials());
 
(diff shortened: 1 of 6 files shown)


cartographer/cloud/internal/local_trajectory_uploader.cc
```diff
@@ -108,19 +107,15 @@ class LocalTrajectoryUploader : public LocalTrajectoryUploaderInterface {
...
-    auto call_creds = async_grpc::TokenFileCredentials(
-        token_file_path, std::chrono::seconds(kTokenRefreshIntervalInSeconds));
-    channel_creds =
-        grpc::CompositeChannelCredentials(channel_creds, call_creds);
-  }
+      enable_google_auth
+          ? grpc::GoogleDefaultCredentials()
+          : enable_ssl_encryption
+                ? ::grpc::SslCredentials(::grpc::SslCredentialsOptions())
+                : ::grpc::InsecureChannelCredentials();
```
cschuet: Wow :) Can you put some brackets to clarify the evaluation order?

cschuet:
@wally-the-cartographer merge

stevewolter:
Sure, I sent you a PR for removing TokenFileCredentials in async_grpc.

cartographer/cloud/internal/local_trajectory_uploader.cc
stevewolter: Done. It's only one pair of parenthesis, but maybe it makes things easier to read :).

wally-the-cartographer:

PR #1426, opened 2018-09-14
Increase gRPC timeout for AddTrajectory

gaschler:
A common use case is that a gRPC server and a gRPC--ROS bridge node
are started at the same time.
Setting the timeout of the stub higher ensures that these can connect
even when the gRPC server starts up slowly.


gaschler:
Good point, I reduced the number of retries to 5 so it only tries for ~30 seconds.

gaschler:
@wally-the-cartographer merge

wally-the-cartographer:

Issue #1422, opened 2018-09-12
Allow snapshotting of large pose graphs from cartographer server


cschuet:
Currently when snapshotting the cartographer server state using MapBuilderService::WriteState, we stream back a sequence of SerializedData protos. One of them contains the PoseGraphProto which can easily grow beyoned the maximum supported protobuf message size of gRPC (currently 10 MB).

I would be better to add a WriteStateToFile() gRPC method which allows the client to ask the server to directly write the pbstream to a local file.

This method should also be invoked from the cartographer_grpc_node when the ROS service for saving state is called.

cschuet:
@MichaelGrupp FYI

MichaelGrupp:
Resolved, you can close this.

cschuet:
Resolved by https://github.com/googlecartographer/cartographer/pull/1424.

PR #1428, opened 2018-09-14
Add WriteStateToFileResponse.

MichaelGrupp:
Changes the public map builder API, but is required to not break the ROS API.
See also https://github.com/googlecartographer/cartographer_ros/pull/1014 and #1422

### DIFF
--- a/cartographer/cloud/client/map_builder_stub.cc
+++ b/cartographer/cloud/client/map_builder_stub.cc
@@ -166,8 +166,8 @@ bool MapBuilderStub::SerializeStateToFile(bool include_unfinished_submaps,
       client_channel_);
   if (!client.Write(request, &status)) {
     LOG(ERROR) << "WriteStateToFileRequest failed - "
-    << "code: " << status.error_code()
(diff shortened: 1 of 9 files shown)


cartographer/cloud/client/map_builder_stub.cc
```diff
@@ -164,9 +164,12 @@ void MapBuilderStub::SerializeStateToFile(bool include_unfinished_submaps,
   ::grpc::Status status;
   async_grpc::Client<handlers::WriteStateToFileSignature> client(
       client_channel_);
-  CHECK(client.Write(request, &status))
-      << "code: " << status.error_code()
-      << " reason: " << status.error_message();
+  if (!client.Write(request, &status)) {
+    LOG(ERROR) << "WriteStateToFileRequest failed - "
```
MichaelGrupp: Removing this CHECK - e.g. an empty filename from a ROS service request could crash the client otherwise.

cschuet:
@wally-the-cartographer merge

PR #1430, opened 2018-09-17
Stop using broken async_grpc::Client retry strategy

cschuet:


cschuet:
@MichaelGrupp FYI

MichaelGrupp:
Please wait.

We should try using https://github.com/grpc/grpc/blob/master/doc/keepalive.md

pifon2a:
@cschuet , @MichaelGrupp is this PR still relevant?

MichaelGrupp:
This was and is a hack to catch physical network outage, which is not always caught in the gRPC channel state. The `async_grpc` retry strategies didn't work well in this situation, so we did a workaround that recreates the client periodically (reconnecting every time, catching bad connections).

As I wrote in my previous comment, a heartbeat/keepalive for the channel is probably a cleaner way to go. Also, a channel state notification (as discussed internally, see my mail: "If you are not sending RPCs and not subscribed for channel state notifications, you might not see the channel going disconnected.")

MichaelGrupp:
I think we can close this, but should continue tracking it in an issue.

I can open one.

PR #1442, opened 2018-10-08
[ABSL] Use flat_hash_set instead of unordered_set.

pifon2a:


sebastianklose:
@wally-the-cartographer merge

PR #1446, opened 2018-10-15
Allow immediate destruction of pose graph

ojura:
If the pose graph work queue gets clogged up, libcartographer users are forced to wait until it gets fully processed when they are destroying Cartographer. Change this to drop all queued tasks if the user is destructing.
If the user wishes to wait until everything gets processed, they can call `RunFinalOptimization`. Alternatively, we could expose `WaitForAllComputations` for achieving the same thing without running that additional optimization.

I have made the change only for 3D in this PR, I can replicate when (if? :-)) it's approved.


cartographer/mapping/internal/3d/pose_graph_3d.cc
```diff
@@ -560,7 +564,7 @@ void PoseGraph3D::WaitForAllComputations() {
                                   absl::FromChrono(common::FromSeconds(1.)))) {
     report_progress();
   }
-  CHECK_EQ(constraint_builder_.GetNumFinishedNodes(), num_trajectory_nodes);
+  // CHECK_EQ(constraint_builder_.GetNumFinishedNodes(), num_trajectory_nodes);
```
ojura: Maybe this `CHECK` can be moved into `RunFinalOptimization`.

damienrg:
I think that you should retain the old behavior by default as some users may depend on the current behavior.

cartographer/sensor/compressed_point_cloud.cc
```diff
@@ -61,8 +61,8 @@ RangefinderPoint CompressedPointCloud::ConstIterator::operator*() const {
   return {current_point_};
 }
 
-CompressedPointCloud::ConstIterator& CompressedPointCloud::ConstIterator::
-operator++() {
+CompressedPointCloud::ConstIterator&
+CompressedPointCloud::ConstIterator::operator++() {
```
damienrg: Revert this change as it is not related to the PR?

ojura:
> I think that you should retain the old behavior by default as some users may depend on the current behavior.

Can you give a concrete example? In cartographer_ros, the offline node calls `RunFinalOptimization`, so it is not affected. 

I do not see a case where users could even theoretically depend on the results of work queue being fully processed - if you are destructing, you cannot access the data in the pose graph anyway, so effects of everything in the work queue being processed are not observable to you.

damienrg:
> Can you give a concrete example? In cartographer_ros, the offline node calls RunFinalOptimization, so it is not affected.

If someone relies on `LocalSlamResultCallback` and/or `GlobalSlamOptimizationCallback` being called even if I think that is unlikely.

cartographer/mapping/internal/3d/pose_graph_3d.cc
damienrg: I see this as a warning as it could mean that implications of the change are not well understood and that it was developed in this fashion:
1. change the code to clear the queue
2. test
3. got the assertion
4. comment the assertion
5. test
6. ok

This is just to let you know how it can be interpreted as I think that you are aware of implications of your code change.

ojura:
Such users can call `RunFinalOptimization` explicitly. I think the benefit of others not having to `SIGKILL` Cartographer when the work queue is clogged is more important.

cartographer/mapping/internal/3d/pose_graph_3d.cc
ojura: You are right it is the case that I stumbled into the assertion after I implemented clearing the work queue, and that this was "developed" in such a way. This does not preclude understanding (or trying to) what is going on and trying to come up with the ideal solution :-) This was more intended to be an RFC (hence my "maybe we could move this..." comment) rather than the definite proposal.

I am also not quite happy with this because even with this change, we can still end up waiting for some time because the constraint builder has its own set of scheduled tasks that can be blocking destruction.

wohe: @ojura Code in comments is bad and should rather be removed completely. That said, if you would like this merged it would need to be a PR intended for review, not an RFC as PR. That is you should have thought through the implications and be able to explain why this is fine the way you implemented it. Please let me know if you want to do this. Otherwise, I would close the PR.

ojura: @wohe in that case the PR should be closed because it's not ready for review prior to merge; and an issue or RFC opened for this, perhaps referencing the changes here as a starting point.

ojura: @wohe closed and opened #1721 referencing this.

PR #1453, opened 2018-10-24
Fix virtual destructors

gaschler:
This is necessary because otherwise derived classes
are potentially deleted incompletely, using the destructor
of the interface.


wally-the-cartographer:
Merge requested by authorized user pifon2a. Merge queue now has a length of 3.

PR #1454, opened 2018-10-24
Fix locking while modifying counter

gaschler:
Clang with thread safety does not compile because
num_nodes_since_last_loop_closure_
is modified without holding a mutex.
This fixes it.

### DIFF
--- a/cartographer/mapping/internal/3d/pose_graph_3d.cc
+++ b/cartographer/mapping/internal/3d/pose_graph_3d.cc
@@ -371,6 +371,7 @@ WorkItem::Result PoseGraph3D::ComputeConstraintsForNode(
     }
   }
   constraint_builder_.NotifyEndOfNode();
+  absl::MutexLock locker(&mutex_);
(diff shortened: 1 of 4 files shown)


cartographer/mapping/internal/3d/pose_graph_3d.cc
```diff
@@ -371,6 +371,7 @@ WorkItem::Result PoseGraph3D::ComputeConstraintsForNode(
     }
   }
   constraint_builder_.NotifyEndOfNode();
+  absl::MutexLock locker(&mutex_);
```
danielsievers: Doesn't look like you strictly need the mutex for num_nodes_since_last_loop_closure_ since it's only accessed from the work queue, so you could also remove the annotation instead. But this seems fine too, as not much happens here after taking the mutex and until giving it up.

gaschler:
@wally-the-cartographer merge

gaschler:
@wally-the-cartographer merge

PR #1462, opened 2018-11-13
Add git dependency to package.xml (for Abseil build).

MichaelGrupp:
Otherwise, building on a ROS buildfarm (with bloom) fails with:
`error: could not find git for clone of abseil`

### DIFF
--- a/package.xml
+++ b/package.xml
@@ -36,6 +36,7 @@
 
   <buildtool_depend>catkin</buildtool_depend>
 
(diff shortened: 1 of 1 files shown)


gaschler:
@wally-the-cartographer merge

MichaelGrupp:
The same also here. https://github.com/googlecartographer/cartographer_ros/pull/1098

PR #1465, opened 2018-11-19
Add gauge metrics for inter-submap constraints.

MichaelGrupp:
Since the constraint builder metrics are ever-increasing, it makes sense
to add gauges that monitor the current counts of constraints in the pose
graph. The main advantages are that they also take trimming into account
and distinguish constraints within or across trajectories.

### DIFF
--- a/cartographer/mapping/internal/2d/pose_graph_2d.cc
+++ b/cartographer/mapping/internal/2d/pose_graph_2d.cc
@@ -42,6 +42,8 @@ namespace cartographer {
 namespace mapping {
 
 static auto* kWorkQueueDelayMetric = metrics::Gauge::Null();
+static auto* kConstraintsSameTrajectoryMetric = metrics::Gauge::Null();
+static auto* kConstraintsDifferentTrajectoryMetric = metrics::Gauge::Null();
 
 PoseGraph2D::PoseGraph2D(
     const proto::PoseGraphOptions& options,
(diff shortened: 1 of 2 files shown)


MichaelGrupp:
Adding intra constraints would be easy, but I didn't do it so far because they don't have useful information for me.

cartographer/mapping/internal/2d/pose_graph_2d.cc
```diff
@@ -480,6 +482,25 @@ void PoseGraph2D::HandleWorkQueue(
     num_nodes_since_last_loop_closure_ = 0;
   }
 
+  // Update the gauges that count the current number of constraints.
+  double inter_constraints_same_trajectory = 0;
+  double inter_constraints_different_trajectory = 0;
+  for (const auto& constraint : constraints()) {
```
gaschler: I think you need to hold the mutex while reading constraints, so move this block inside the block above.

cartographer/mapping/internal/2d/pose_graph_2d.cc
```diff
@@ -480,6 +482,25 @@ void PoseGraph2D::HandleWorkQueue(
     num_nodes_since_last_loop_closure_ = 0;
   }
 
+  // Update the gauges that count the current number of constraints.
```
gaschler: Could be an extra function (I leave this to you)

cartographer/mapping/internal/2d/pose_graph_2d.cc
MichaelGrupp: `constraints()` uses a mutex: https://github.com/googlecartographer/cartographer/blob/7d69ca9fa0393a0be2ae6dd5781d7a8ee9f4dcb1/cartographer/mapping/internal/2d/pose_graph_2d.cc#L946

gaschler: Then it is still preferable not to drop and take the same mutex but iterate over data_ instead while holding it.

cartographer/mapping/internal/2d/pose_graph_2d.cc
MichaelGrupp: As long as it's in `HandleWorkQueue` I don't see the need to call it somewhere else.

gaschler:
@wally-the-cartographer merge

MichaelGrupp:
Is Wally running?

wally-the-cartographer:

PR #1475, opened 2018-12-06
Allow introduction of new landmarks into loaded (frozen) map

### LINKED ISSUE #1470: Add landmarks to existing map for pure localization mode
In our application we have a known (mapped) environment and operate in pure localization mode.
Since we have poor localization performance in certain parts of the map and possibly also to accelerate initialization, we'd like to add distinct landmarks (QR-Tags or similar) at certain places.

Is there a way to define the landmark poses with respect to the map at this stage? We already have a pbstream of the map, but don't want to rebuild the map because it takes too much time considering that we will need to try different landmark configurations.

johuber:
In the current implementation, as soon as one trajectory is frozen it is not possible to use new landmarks anymore.
In this PR a suggest a change, that allows introducing new landmarks also in the presence of a frozen trajectory. This is done with the following steps:
- Store set of frozen landmarks
- Landmarks that are loaded from a pbstream, with the flag ` load_frozen_state=true` are inserted into the set of frozen landmarks via `setLandmarkPose` function
- frozen landmarks are set constant in the optimization problem
- new landmarks can be introduced and will be optimized in the optimization problem

Motivated by and closes #1470


cartographer/mapping/pose_graph_data.h
```diff
@@ -74,6 +74,7 @@ struct PoseGraphData {
   // Global landmark poses with all observations.
   std::map<std::string /* landmark ID */, PoseGraphInterface::LandmarkNode>
       landmark_nodes;
+  std::set<std::string /* landmark ID */> frozen_landmarks;
```
pifon2a: I like this change and it makes perfect sense to have control over what pose graph nodes are constant or not. Thank you for making this. 

Just one suggestion: instead of having a separate `set<string> frozen_landmarks` in `PoseGraphData`, please, add a field `bool frozen` to `PoseGraphInterface::LandmarkNode`. It should simplify the code and keep things together.

johuber: Yes, I was thinking about that solution too. I can make the change eventually; however, it might take me a couple of weeks until I have time

johuber:
changes made as suggested. Now storing `bool frozen` in `LandmarkNode` instead of keeping set of frozen landmarks. Otherwise it remains the same.

cartographer/mapping/internal/3d/pose_graph_3d.cc
```diff
@@ -952,6 +952,7 @@ void PoseGraph3D::SetLandmarkPose(const std::string& landmark_id,
   AddWorkItem([=]() LOCKS_EXCLUDED(mutex_) {
     absl::MutexLock locker(&mutex_);
     data_.landmark_nodes[landmark_id].global_landmark_pose = global_pose;
+    data_.landmark_nodes[landmark_id].frozen = true;
```
pifon2a: This is an interesting moment. What if someone uses `SetLandmarkPose` to specify  starting point for optimization? Should it be always frozen then? Same in 2d case.

cartographer/mapping/internal/optimization/optimization_problem_3d.cc
```diff
@@ -167,7 +163,8 @@ void AddLandmarkCostFunctions(
             CeresPose(starting_point, nullptr /* translation_parametrization */,
                       absl::make_unique<ceres::QuaternionParameterization>(),
                       problem));
-        if (freeze_landmarks) {
+        // Set landmark constant if it is frozen
```
pifon2a: nit: period after 'frozen'.

cartographer/mapping/internal/3d/pose_graph_3d.cc
johuber: So far I did not know that somebody would use it like this. But we could easily deal with it by adding a `bool frozen=false` to the function arguments.

I think whenever they are read from a .pbstream file they should be frozen.

We could also think about what should happen to the landmarks when  `freezeTrajectory(..)` is called. Personally, since landmarks are not directly associated to a trajectory, I think this call should not affect the landmarks.

If somebody ever needs to freeze individual landmarks a method `freezeLandmark(..)` should be implemented.

pifon2a: Sometimes we load a pbstream to continue optimizing jointly everything we have. Then it should not be frozen.  Please, add `bool frozen=false` to the function arguments.

I agree with `FreezeLandmark(...)`.

johuber: okay, so should I in this case also add an option `--freeze_landmarks` similar to `--freeze_trajectory`? Or should I keep the setting in `map_builder.cc`` to `frozen = true`, which finally corresponds to the original state?

pifon2a: for now i would leave the flags as they are. At least for this PR.

johuber: I made the change to the function signature. I'm setting `frozen = true` in `map_builder.cc` but otherwise the function arguments defaults to false.

I did not fully look into the _cloud_ methods, they still work as expected, but I don't understand it well enough to reason how frozen should be set there or how it should be added to the protobuf requests.

If this is necessary, does somebody else have time to do it?

pifon2a:
@johuber One build is failing.

johuber:
> @johuber One build is failing.

yep I'm slowly getting there to fix it!

johuber:
tests passed

pifon2a:
@wally-the-cartographer merge

pifon2a:
@johuber thanks!

cartographer/mapping/map_builder.cc
```diff
@@ -279,7 +279,8 @@ std::map<int, int> MapBuilder::LoadState(
   // Set global poses of landmarks.
   for (const auto& landmark : pose_graph_proto.landmark_poses()) {
     pose_graph_->SetLandmarkPose(landmark.landmark_id(),
-                                 transform::ToRigid3(landmark.global_pose()));
+                                 transform::ToRigid3(landmark.global_pose()),
+                                 true);
```
rgreid: Here, `SetLandmarkPose()` is being called with `frozen = true`.  Shouldn't `MapBuilder::LoadState` apply the `load_frozen_state` flag to landmarks also?

johuber: I guess it could, what I did here was just replicating the behavior that it had before.

YangAndI:
I use landmarks during running pure localization, but the scan_to_map is wrong when standing still for more than five minutes

YangAndI:
But the error is  disappeared time lasted an hour when i close frozen

johuber:
> I use landmarks during running pure localization, but the scan_to_map is wrong when standing still for more than five minutes

are you using this PR or how is this related?
This PR should not affect anything when you are doing pure localization.

YangAndI:
These landmarks not working when doing pure localization?

johuber:
> These landmarks not working when doing pure localization?

They do if you mapped the landmarks in your original map.

YangAndI:
> > These landmarks not working when doing pure localization?
> 
> They do if you mapped the landmarks in your original map.

I mapped the landmarks in my map and publish /landmark during doing pure localization,but these are no constraints with landmarks.And it away lose pose in  a similar environment. What should i pay attention to when using landmarks

Issue #1481, opened 2018-12-15
Build error, can not found protobuf which is already installed


jinfagang:
```
[138/261] Building CXX object CMakeFiles/cartographer.mapping.intern...artographer/mapping/internal/trajectory_connectivity_state_test.cc.o
FAILED: cartographer_autogenerate_ground_truth 
: && /usr/bin/c++  -O3 -DNDEBUG  -rdynamic CMakeFiles/cartographer_autogenerate_ground_truth.dir/cartographer/ground_truth/autogenerate_ground_truth_main.cc.o  -o cartographer_autogenerate_ground_truth  -Wl,-rpath,/usr/local/lib: libcartographer.a /usr/local/lib/libceres.a -lglog /usr/lib/x86_64-linux-gnu/libgflags.so.2.2.1 -lspqr -ltbbmalloc -ltbb -lcholmod -lccolamd -lcamd -lcolamd -lamd -lopenblas -lsuitesparseconfig -lrt -lcxsparse -lopenblas -lsuitesparseconfig -lrt -lcxsparse -llua5.2 -lm -lboost_iostreams -lboost_regex -lglog /usr/lib/x86_64-linux-gnu/libgflags.so.2.2.1 -lpthread -lcairo /usr/local/lib/libprotobuf.so -lpthread && :
libcartographer.a(relations.pb.cc.o): In function `cartographer::ground_truth::proto::Relation::GetMetadata() const':
relations.pb.cc:(.text+0x81): undefined reference to `google::protobuf::GoogleOnceInitImpl(long*, google::protobuf::Closure*)'
libcartographer.a(relations.pb.cc.o): In function `cartographer::ground_truth::proto::Relation::SerializeWithCachedSizes(google::protobuf::io::CodedOutputStream*) const':
relations.pb.cc:(.text+0x133): undefined reference to `google::protobuf::internal::proto3_preserve_unknown_'
libcartographer.a(relations.pb.cc.o): In function `cartographer::ground_truth::proto::GroundTruth::SerializeWithCachedSizes(google::protobuf::io::CodedOutputStream*) const':
relations.pb.cc:(.text+0x1eb): undefined reference to `google::protobuf::internal::proto3_preserve_unknown_'
[...]

PR #1491, opened 2019-01-02
Fix regression in Protobuf find_package logic.

MichaelGrupp:
The introduced changes are not compatible with CMake < 3.6 if the
FindProtobuf.cmake bundled with CMake is used.
Since our minimum required version is 3.2, I consider this to be a bug.
Issues reported in #1481 are probably related.

- camel-case variables names weren't used before 3.6
  (upper-case is lowered in CMake > 3.6 for backwards compatibility)
- the `CONFIG` search added for the Windows build logs a warning even if the
`REQUIRED` search succeeds, should be quiet

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -41,8 +41,8 @@ find_package(Eigen3 REQUIRED)
 find_package(LuaGoogle REQUIRED)
 # On Windows, Protobuf is incorrectly found by the bundled CMake module. Prefer native CMake config, if possible.
 set(protobuf_MODULE_COMPATIBLE TRUE CACHE INTERNAL "")
-find_package(Protobuf 3.0.0 CONFIG)
-if (NOT Protobuf_FOUND)
+find_package(Protobuf 3.0.0 CONFIG QUIET)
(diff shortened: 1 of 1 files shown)


MichaelGrupp:
https://cmake.org/cmake/help/v3.2/module/FindProtobuf.html

Backwards compatibility: https://github.com/Kitware/CMake/commit/a7b09e7f43fda0a459145053653a614330bf1fe1#diff-a77489da88045e361833b7dfe293c07fR215

I guess the CI pipeline didn't complain because it probably doesn't use the CMake-bundled file.

MichaelGrupp:
Bazel build fails (also on master) because of some unrelated problems.

ojura:
Sorry about that!

ojura:
Windows build is also happy: https://dev.azure.com/larics/cartographer/_build/results?buildId=194

It looks like I wrongly concluded that we need to switch to the camelcase variable to use Protobuf in config mode.

mrgransky:
You may have a look at [here](https://github.com/googlecartographer/cartographer/issues/1481#issuecomment-568904908)!

PR #1495, opened 2019-01-09
Add serialization for TimestampedTransform.

wohe:


cartographer/transform/timestamped_transform_test.cc
```diff
@@ -0,0 +1,39 @@
...
+namespace {
+
+TEST(TimestampedTransformTest, ToProtoAndBack) {
+  const TimestampedTransform expected{
+      common::FromUniversal(12345678),
+      Rigid3d(Eigen::Vector3d(1., 2., 3.),
+              Eigen::Quaterniond(1., 2., 3., 4.).normalized())};
+  const TimestampedTransform actual = FromProto(ToProto(expected));
+  EXPECT_EQ(expected.time, actual.time);
+  EXPECT_THAT(actual.transform, IsNearly(expected.transform, 1e-6));
```
danielsievers: Should be exactly the same right?

wohe: If we assume IEEE-754, I think so. But `IsNearly` is readily available while `operator==` is not, and it gives us good enough coverage.

wohe:
@wally-the-cartographer merge

wally-the-cartographer:

PR #1500, opened 2019-01-15
Fix TSD converter include location

alexmillane:
The TSDF2D class included a header dependency on TSDValueConverter (an internal component). This breaks the build for cartographer clients wanting to make use of TSDF2D objects. I have moved TSDValueConverter out of the internal space to solve this problem.


alexmillane:
Signed (my first born children away).

PR #1504, opened 2019-01-17
Save map builder options in options proto. Rename it to BuilderOptions.

ojura:
RFC=https://github.com/googlecartographer/rfcs/pull/38

### DIFF
--- a/cartographer/io/proto_stream_deserializer.cc
+++ b/cartographer/io/proto_stream_deserializer.cc
@@ -57,17 +57,17 @@ ProtoStreamDeserializer::ProtoStreamDeserializer(
          "`SerializationHeader`, but got field tag "
       << pose_graph_.data_case();
 
-  CHECK(ReadNextSerializedData(&all_trajectory_builder_options_))
-      << "Serialized stream misses `AllTrajectoryBuilderOptions`.";
-  CHECK(all_trajectory_builder_options_.has_all_trajectory_builder_options())
+  CHECK(ReadNextSerializedData(&builder_options_))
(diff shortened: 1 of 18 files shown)


ojura:
I verified that the pbstream info tool correctly prints out the map builder options for .pbstreams produced with this version of Cartographer. For older pbstreams it just prints out an empty string for map builder options.

cartographer/io/proto_stream_deserializer.cc
```diff
@@ -57,17 +57,17 @@ ProtoStreamDeserializer::ProtoStreamDeserializer(
          "`SerializationHeader`, but got field tag "
       << pose_graph_.data_case();
 
-  CHECK(ReadNextSerializedData(&all_trajectory_builder_options_))
-      << "Serialized stream misses `AllTrajectoryBuilderOptions`.";
-  CHECK(all_trajectory_builder_options_.has_all_trajectory_builder_options())
+  CHECK(ReadNextSerializedData(&builder_options_))
```
MichaelGrupp: Does this break compatibility?

ojura: It is the same proto, just renamed to builder options (instead of all trajectory builder options), since it now contains the map builder options as well.

MichaelGrupp: Oh yes, true!

daddivan69 (approved):

bochen87:
@ojura this was reviewed and approved a long time ago but somehow not merged. It is now outdated and has many conflicts. Do you still have time to resolve the conflict?

PR #1529, opened 2019-03-06
Do not use native CMake config for Protobuf on Linux

ojura:
On Ubuntu 16.04, we have to install Protobuf 3. Unfortunately, `find_package()`ing Protobuf using native installed CMake `CONFIG` is problematic in case if there is a system Protobuf 2 installed, while the bundled CMake module works correctly and manages to avoid interference with system-installed Protobuf 2.


pifon2a:
@wally-the-cartographer merge

wally-the-cartographer:

PR #1558, opened 2019-04-02
Check for empty basename in GetFileContentOrDie()

MichaelGrupp:
An empty basename would mean a directory path is returned by
GetFullPathOrDie(), which then leads to a cryptic exception when trying
to read from it using the istreambuf_iterator:

"basic_filebuf::underflow error reading the file: iostream error"


sebastianklose:
@wally-the-cartographer merge

wally-the-cartographer:
Merge requested by authorized user sebastianklose. Merge queue now has a length of 1.

PR #1563, opened 2019-04-09
Use fixed precision for string histogram buckets.

MichaelGrupp:
`absl::StrCat/StrAppend` convert numeric values to strings with inconsistent
decimal precision, which can lead to ugly formatting of the histogram.

This can be fixed by using `StrAppendFormat` with `%f`, which uses 6
decimals (printf default and as it was when we had `std::to_string`).


gaschler:
@MichaelGrupp LGTM, but please force push again to undo what the formatting bot did

gaschler:
@wally-the-cartographer merge

PR #1570, opened 2019-04-17
Use installed Abseil.

pifon2a:


nindanaoto:
I hope that this PR will be merged as soon as possible.
This PR helped me a lot to build latest commit.

wohe:
@pifon2a Thanks for the PR. Could you please consent to the googlebot? I think this PR can be merged. I verified that with a small set of follow up fixes, cartographer_ros can be built with a separate abseil package.

wohe:
@ojura @barcode I am in favor of merging this PR. What do you think? It seems the cleanest to me as it moves building Abseil out of this library, so everyone can pick their favorite method. Regarding ROS: it means we need to build Abseil there but this seems pretty straightforward.

barcode:
I agree this PR would be the best to merge. Since this one was unresolved for one year and it seemed, there were some concerns because code was changes in several places, I created a PR with minimal changes. I did not touch any ros stuff, since i do not use ros.

I just have to be able to build abseil by myself. Also the old code caused projects depending on this one to build abseil again (due to the find package script).

So +1 for merging this.

ojura:
@wohe I agree, using installed Abseil makes sense (unless you are planning to use bleeding edge features from Abseil master and want to strictly control the Abseil version). 

I think this is orthogonal to improving the case when (if?) we want to build Abseil with Cartographer. For that case, I still think #1387 is the way to go.

barcode:
If you are using a local install of abseil, you can use the bleeding edge features of it anyway.
(In general i would never do a system install of abseil, cartographer or any other lib i am using for research, since a local install works just as well.)

ojura:
@barcode sure, but installing an appropriate build of Abseil would then be an extra step when building Cartographer. Which might be fine (although I am extremely wary of installing anything to the /usr/local prefix - I saw stuff breaking lots of times when doing that because of conflicts with system libs, and undoing a bad install is a giant PITA).

I like the current method of having CMake do it for us if necessary. IMHO I would use the installed version if it is recent enough, and switch to `FetchContent` if we need to build it with Cartographer. (`add_subdirectory()` is the preferred way of using Abseil in their docs, or at least was at some point.)

barcode:
@ojura with local install i meant a user install in the user directory (`CMAKE_INSTALL_PREFIX`). That's why i said, i would never do a system install (e.g.: to `/usr` or `/opt`). My reason is the same as yours: One time I broke my setup by installing a lib that had a broken install target.

PR #1581, opened 2019-05-23
Implement fixed size transform interpolation buffer.

MichaelGrupp:
While using the TransformInterpolationBuffer in private code I noticed
that it seems to be incomplete. Neither does it have a buffer size
limit, nor does it "remove the oldest transform if the buffer size limit
is exceeded" as stated in the header.

The changes don't break compatibility with existing code:
unless explicitly set, the buffer size is unlimited.


cartographer/transform/transform_interpolation_buffer.h
```diff
@@ -17,7 +17,8 @@
 #ifndef CARTOGRAPHER_TRANSFORM_TRANSFORM_INTERPOLATION_BUFFER_H_
 #define CARTOGRAPHER_TRANSFORM_TRANSFORM_INTERPOLATION_BUFFER_H_
 
-#include <vector>
+#include <deque>
+#include <numeric>
```
wohe, line 21: Typo? `#include <limits>` contains `std::numeric_limits`.

cartographer/transform/transform_interpolation_buffer.h
```diff
@@ -27,18 +28,29 @@
...
+constexpr size_t kUnlimitedBufferSize = std::numeric_limits<size_t>::max();
+
 // A time-ordered buffer of transforms that supports interpolated lookups.
+// Unless explicitly set, the buffer size is unlimited.
 class TransformInterpolationBuffer {
  public:
   TransformInterpolationBuffer() = default;
   explicit TransformInterpolationBuffer(
       const mapping::proto::Trajectory& trajectory);
+  explicit TransformInterpolationBuffer(const size_t buffer_size_limit);
```
wohe, line 40: Not sure this constructor provides much value. Convenience is not necessarily bad, but here it seems to also harm readability as you would call the constructor with just an integer. Calling `SetSizeLimit()` directly after construction makes it much clearer what is happening. Remove?

MichaelGrupp: Sounds good.

MichaelGrupp:
I do have a use case for that in an internal package that uses libcartographer. Besides that, it could be also useful for others.

cartographer/transform/transform_interpolation_buffer.h
```diff
@@ -27,18 +28,28 @@
...
+// Unless explicitly set, the buffer size is unlimited.
 class TransformInterpolationBuffer {
  public:
   TransformInterpolationBuffer() = default;
   explicit TransformInterpolationBuffer(
       const mapping::proto::Trajectory& trajectory);
 
+  // Sets the transform buffer size limit and removes old transforms
+  // if it is exceeded.
+  void SetSizeLimit(const size_t buffer_size_limit);
```
wohe, line 43: Nit: the `const` should be removed as it has no meaning here. It is only useful at the implementation.

wohe: done

PR #1583, opened 2019-05-29
Instead of crashing when timestamps out of order, log and continue.

rotu:
When using some ROS2 QoS settings, data can arrive out of order or multiple times. Instead of treating this as a fatal error, log it as a warning and ignore the delayed data.

### DIFF
--- a/cartographer/sensor/map_by_time.h
+++ b/cartographer/sensor/map_by_time.h
@@ -40,7 +40,13 @@ class MapByTime {
     CHECK_GE(trajectory_id, 0);
     auto& trajectory = data_[trajectory_id];
     if (!trajectory.empty()) {
-      CHECK_GT(data.time, std::prev(trajectory.end())->first);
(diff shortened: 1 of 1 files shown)


rotu:
I signed it!

ojura:
This was an issue discussed many times and the authors (note that I am not one of them so I can‘t speak on their behalf) have always taken a stance that the Cartographer __library__ must be provided valid data e.g. no NaN range measurements and always strictly increasing timestamps. This is strictly enforced everywhere with crash-inducing checks.

Maybe ignoring out of order messages could be added as an option to the ROS wrapper (I think that would be more feasible than changing the library itself). But be prepared for a response such as „fix your data, not Cartographer“ :)

rotu:
Thanks, @ojura. I'm coming to this with fresh eyes, and I appreciate your patient explanation! I'll redirect my efforts to fix this in the ROS layer.

PR #1580, opened 2019-05-23
FixedFramePose in optimization 2d

jihoonl:
Most of the code is copied from optimization_3d except the cost function replacement.

PR #1591, opened 2019-06-12
Added fixed frame pose data to the optimization problem and pose graph

Conhokis:
I tried to do what was suggested here [https://github.com/googlecartographer/cartographer_ros/issues/835#issuecomment-383504222](https://github.com/googlecartographer/cartographer_ros/issues/835#issuecomment-383504222).

The recommended compilation test was passed and messages when running cartographer_ros appear which set the initial coordinates. I'm not at all sure that my solutions is correct, I would wish some feedback since I am currently using this in a college project.


cartographer/mapping/internal/optimization/optimization_problem_2d.cc
```diff
@@ -301,6 +350,64 @@ void OptimizationProblem2D::Solve(
...
+                nullptr,
+                absl::make_unique<ceres::AutoDiffLocalParameterization<
+                    YawOnlyQuaternionPlus, 4, 1>>(),
+                &problem));
+        fixed_frame_pose_initialized = true;
+      }
+
+      problem.AddResidualBlock(CreateAutoDiffSpaCostFunction(constraint_pose),
+                               nullptr,
+                               C_fixed_frames.at(trajectory_id).translation(),
```
wohe, line 406: This is not quite right I think: Translation is 3 values for x, y, z, but in 2D we expect 3 value for x, y, yaw. This does not necessarily lead to bad results, but means that the optimized value for yaw is not in the correct place in the output.

cartographer/mapping/internal/optimization/optimization_problem_2d.cc
```diff
@@ -301,6 +350,64 @@ void OptimizationProblem2D::Solve(
     }
   }
 
+  // Add fixed frame pose constraints.
+  std::map<int, CeresPose> C_fixed_frames;
```
wohe, line 354: Using a `CeresPose` which is 3D here is inelegant. The rotation part is not used in the optimization below.

wohe:
Hi, thanks a lot for the PR. Did you see that #1580 implemented something very similar? I'm closing this in favor of the other PR. If you think this decision is wrong, please let me know.

PR #1616, opened 2019-07-23
Fix MapById::lower_bound function and implement test to reproduce the bug

mgladkova:
This fixes the bug that has been faced during localization with an initial pose and a trimmed map (OverlappingSubmapsTrimmer2D has been used). A small test is added that reproduces the problem (an infinite loop) where a trimmed trajectory (MapById instance) is used.

### DIFF
--- a/cartographer/mapping/id.h
+++ b/cartographer/mapping/id.h
@@ -383,22 +383,26 @@ class MapById {
       return EndOfTrajectory(trajectory_id);
     }
 
-    size_t left = 0;
-    size_t right = trajectory.size() - 1;
+    auto left = trajectory.begin();
+    auto right = std::prev(trajectory.end());
     while (left != right) {
-      const int middle_index = (left + right) / 2;
-      auto middle = trajectory.begin();
-      std::advance(middle, middle_index);
-
-      if (internal::GetTime(middle->second) < time) {
-        left = middle_index + 1;
+      // This is never 'right' which is important to guarantee progress.
+      const int middle = left->first + (right->first - left->first) / 2;
+      // This could be 'right' in the presence of gaps, so we need to use the
+      // previous element in this case.
+      auto lower_bound_middle = trajectory.lower_bound(middle);
+      if (lower_bound_middle->first > middle) {
+        CHECK(lower_bound_middle != left);
+        lower_bound_middle = std::prev(lower_bound_middle);
+      }
+      if (internal::GetTime(lower_bound_middle->second) < time) {
+        left = std::next(lower_bound_middle);
       } else {
-        right = middle_index;
+        right = lower_bound_middle;
       }
     }
-    auto result = trajectory.begin();
-    std::advance(result, left);
-    return ConstIterator(*this, IdType{trajectory_id, result->first});
+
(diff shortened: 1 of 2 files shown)


mgladkova:
@googlebot I consent.

cartographer/mapping/id_test.cc
```diff
@@ -254,6 +254,53 @@ TEST(IdTest, LowerBoundFuzz) {
...
+
+  auto trim_segment_start = map_by_id.begin();
+  std::advance(trim_segment_start, trim_segment_start_index);
+
+  auto trim_segment_end = map_by_id.begin();
+  std::advance(trim_segment_end,
+               trim_segment_start_index + trim_segment_length);
+
+  for (auto it = trim_segment_start; it != trim_segment_end;) {
+    const auto this_it = it;
```
wohe, line 288: This is necessary since after trimming, `it` no longer is a valid iterator, so `++it` after trimming cannot be done and we have to do it before.

cartographer/mapping/id.h
```diff
@@ -383,22 +383,26 @@ class MapById {
...
     }
 
-    size_t left = 0;
-    size_t right = trajectory.size() - 1;
+    auto left = trajectory.begin();
+    auto right = std::prev(trajectory.end());
     while (left != right) {
-      const int middle_index = (left + right) / 2;
-      auto middle = trajectory.begin();
-      std::advance(middle, middle_index);
```
wohe: @mgladkova Thanks a lot for pointing out the bug. I changed the fix a bit, since this would change asymptotic behavior: this would take linear time for a `std::map`.

wohe:
@mgladkova Sorry that I just went in and changed your PR. I was not expecting a quick reaction from your side after this PR has been sitting around for a year. Do my proposed changes look good to you? If they do, and if CI passes, I would merge.

MichaelGrupp:
Hi Wolfgang, thanks for the review. We have been using this patch on our robots since a year and haven't experienced a crash due to this anymore. So I'd say let's get this merged (@mgladkova wrote this as an intern at Magazino, so I'm speaking on behalf of her since she's not working on this project anymore)

---

Is this repo now alive again? I know that it's not maintained by the ex Google team anymore and the last time I talked to Christoph last summer it was unclear how to proceed. Are there any plans yet? Also, regarding cartographer_ros.

We have since been working exclusively with our forks and it would be good to know if it's worth to put efforts in pull requests again. Would be also good to communicate to others: #1675

wohe:
@MichaelGrupp Okay, merged.

You might have noticed that the project is now at https://github.com/cartographer-project. Together with https://www.openrobotics.org/ we are currently figuring out how to best proceed. Hopefully I can share more soon. If you want to be involved, just let me know.

Regarding PRs: yes, please send them. I expect things can be a bit slow at least for the next weeks.

MichaelGrupp:
Good to hear that. I would be interested in participating, e.g. doing some kind of co-maintainership of cartographer_ros or other things. A community effort is the way to go on here IMO.

Feel free to contact me via e-mail for more: grupp<at>magazino.eu

mgladkova:
thank you, @wohe , for your corrections and the long-awaited review!

PR #1637, opened 2019-10-10
MinMaxRangeFiteringPointsProcessor: use squaredNorm() instead of norm()

facontidavide:
Passes all the tests...


facontidavide:
Shall we also correct the typo in the name (Fitering)?

PR #1693, opened 2020-06-02
Fix build status on front page.

wohe:
Changes "googlecartographer" to "cartographer-project"
for references to CI and GitHub.

### DIFF
--- a/package.xml
+++ b/package.xml
@@ -28,7 +28,7 @@
   </maintainer>
   <license>Apache 2.0</license>
 
-  <url>https://github.com/googlecartographer/cartographer</url>
+  <url>https://github.com/cartographer-project/cartographer</url>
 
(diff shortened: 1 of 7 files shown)


package.xml
```diff
@@ -28,7 +28,7 @@
   </maintainer>
   <license>Apache 2.0</license>
 
-  <url>https://github.com/googlecartographer/cartographer</url>
+  <url>https://github.com/cartographer-project/cartographer</url>
 
   <author email="google-cartographer@googlegroups.com">
```
feuerste, line 33: Can we also create an alias `cartographer` or `cartographer-project` for `google-cartographer` for the google group?

wohe: We could, but I am not sure what to do there right now. For example, it is unclear to me whether anyone reads messages sent to the google group. I'll leave it for a future PR.

PR #1696, opened 2020-06-04
Introduce CARTOGRAPHER_CONFIG_DIR for relocatable

seanyen:
This pull request is motivated by that on Windows, usually the binaries will be redistributed to a different location than where it was built. However, the current setting of cartographer will look for a [location](https://github.com/cartographer-project/cartographer/blob/2bd987ffb40e2d6667682c5c96432173f371094f/cartographer/common/config.h.cmake#L24) determined at compiled time.

This change is to introduce an environment variable, `CARTOGRAPHER_CONFIG_DIR`, which gives the developers an option to override the default setting, so that makes the binaries more relocatable.

### DIFF
--- a/cartographer/common/configuration_file_resolver.cc
+++ b/cartographer/common/configuration_file_resolver.cc
@@ -19,6 +19,7 @@
 #include <fstream>
 #include <iostream>
 #include <streambuf>
+#include <cstdlib>
 
 #include "cartographer/common/config.h"
 #include "glog/logging.h"
@@ -29,7 +30,11 @@ namespace common {
 ConfigurationFileResolver::ConfigurationFileResolver(
     const std::vector<std::string>& configuration_files_directories)
     : configuration_files_directories_(configuration_files_directories) {
-  configuration_files_directories_.push_back(kConfigurationFilesDirectory);
+  if(const char* config_dir = std::getenv("CARTOGRAPHER_CONFIG_DIR")) {
+    configuration_files_directories_.push_back(config_dir);
+  } else {
+    configuration_files_directories_.push_back(kConfigurationFilesDirectory);
+  }
 }
 
 std::string ConfigurationFileResolver::GetFullPathOrDie(


seanyen:
@googlebot I signed it!

seanyen:
@wohe Hope I am pinging the correct maintainer. This is ready for review and merge. Thanks!

cartographer/common/configuration_file_resolver.cc
```diff
@@ -29,7 +30,11 @@ namespace common {
 ConfigurationFileResolver::ConfigurationFileResolver(
     const std::vector<std::string>& configuration_files_directories)
     : configuration_files_directories_(configuration_files_directories) {
-  configuration_files_directories_.push_back(kConfigurationFilesDirectory);
+  if(const char* config_dir = std::getenv("CARTOGRAPHER_CONFIG_DIR")) {
```
wohe, line 33: I think this at least needs a comment with the rationale why this was added.

Also, it is unclear to me:
1. Is this how libraries which need to load configuration usually handle this in Windows?
2. Should this also be done for non-Windows builds or put behind an `IFDEF`?
3. @ojura I think you did some work on Windows support before. What do you think?

ojura: Hi @wohe! Happy to see you back! It started to look like that the project was going towards a heat death which was a bit sad for me. People at my university are using it on land vehicles and UAVs.

Regarding the pull request, I don't think this is a Windows-specific issue. It makes sense to be able to specify the configuration directory when deploying Cartographer as part of a bigger standalone app.

I am not sure about the environment variable approach, though. Libcartographers's `ConfigurationFileResolver` happily accepts a vector of possible paths for the clients to specify. In my case, I changed the API in the ROS wrapper (https://github.com/cartographer-project/cartographer_ros/blob/29105294468e0cc81def94fd68a540eef7533f46/cartographer_ros/cartographer_ros/node_options.cc#L54) to also accept a vector of directories. Then, when calling `cartographer_ros::LoadOptions`, I can provide all the correct paths.

I think that introducing a gflag in the wrapper for this might be more in line with the rest of the codebas., instead of an environment variable this deeply in the library (we don't have any other env variables AFAIK).

wohe: @seanyen Does the solution suggested by @ojura work for you?

wohe: @seanyen I will close this PR under the assumption that the suggestion above works best for you. Please re-open if you think otherwise.

PR #1697, opened 2020-06-04
Add Ubuntu 18.04 to the install scripts, CI, docs.

wohe:


scripts/install_debs_cmake.sh
```diff
@@ -50,3 +50,9 @@ sudo apt-get install -y \
     libsuitesparse-dev \
     ninja-build \
     python-sphinx
+
+# Install Ceres Solver on Ubuntu Bionic. No need to build it ourselves.
+if [[ "$(lsb_release -sc)" = "bionic" ]]
+then
+  sudo apt-get install -y libceres-dev
```
feuerste, line 57: Just for my understanding: Will the system `libceres` be picked up automatically by the `cartographer` configuration step?

wohe: Correct. Xenial already had a version of libceres-dev but it was missing some files which are needed to use it with CMake. For this reason we were building it ourselves before, but this is no longer necessary.

PR #1700, opened 2020-06-10
Allow using TolerantLoss for INS in pose graph optimization

spielawa:
# Summary

This PR adds the possibility to use the TolerantLoss function for INS in pose graph optimization. This is currently switched off, and so there's not functional change. 

The ceres TolerantLoss function (see [description](http://ceres-solver.org/nnls_modeling.html), and [implementation](https://github.com/ceres-solver/ceres-solver/blob/master/internal/ceres/loss_function.cc)) has the following property

- for large values of x it approaches a quadratic loss ("null loss") with the specified weight (fixed_frame_pose_translation_weight)
- for small values of x it approaches a quadratic loss with a smaller weight
- there's a crossover at some value x_c. The function is convex everywhere.

# Tests and commands

`cd build; cmake .. -G Ninja; ninja; CTEST_OUTPUT_ON_FAILURE=1 ninja test; cd ..`


cartographer/mapping/internal/optimization/optimization_problem_3d.cc
```diff
@@ -548,7 +548,11 @@ void OptimizationProblem3D::Solve(
 
       problem.AddResidualBlock(
           SpaCostFunction3D::CreateAutoDiffCostFunction(constraint_pose),
-          nullptr /* loss function */,
+          /* loss function */
```
wohe, line 551: Can we remove this comment? Seems clear, now that we add loss function. It was only there before for readability since `nullptr` could be anything.

cartographer/mapping/proto/pose_graph/optimization_problem_options.proto
```diff
@@ -18,8 +18,10 @@ package cartographer.mapping.optimization.proto;
 
 import "cartographer/common/proto/ceres_solver_options.proto";
 
-// NEXT ID: 19
+// NEXT ID: 26
+// IDs 20-22 are for visual constraints
```
wohe, line 22: Seems redundant with line 24 below. Remove?

cartographer/mapping/internal/optimization/optimization_problem_3d.cc
spielawa: done

cartographer/mapping/proto/pose_graph/optimization_problem_options.proto
spielawa: Done.

PR #1701, opened 2020-06-11
ROS 2 Cartographer port

clalancette:
This PR is a "port" of the Cartographer code to ROS 2.  I put the words port in quotes because the port doesn't require very much; it is mostly changes to compile on Ubuntu 20.04, plus some changes to the package.xml to make it a pure cmake package rather than a catkin one.

As-is, this PR doesn't apply to master.  That's because this "port" was originally done against the 1.0.0 tag of this repository. However, given the rather simple nature of this PR, it should be pretty easy to move forward.

I'm opening this PR as a starting point for the discussion on how to move this code into the upstream.  Some steps that seem obvious to me:

1.  Update this to apply against master.
1.  Create a ros2 branch for this to target

Some steps that are less obvious, and/or need discussion:
1.  Do we want to try to support ROS 1 and ROS 2 on the same branch?  It *might* be possible here, but I'm not sure.
1.  Is the cartographer project willing to take over maintenance of this branch and release it into newer ROS 2 versions as they come along?

Finally, I'll note that I don't have too much of time to devote to this.  I'm doing this on discretionary time, so while I'm glad to do some simple work here, I won't have the time to do any major refactoring.

Comments, questions, thoughts all welcome!

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -2,6 +2,10 @@
 Changelog for package cartographer
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
+1.0.0-RC1 (2018-05-31)
+----------------------
+https://github.com/googlecartographer/cartographer/compare/0.3.0...1.0.0-RC1
+
 0.3.0 (2017-11-23)
 ------------------
 https://github.com/googlecartographer/cartographer/compare/0.2.0...0.3.0
--- a/cartographer/common/lockless_queue.h
+++ b/cartographer/common/lockless_queue.h
@@ -64,7 +64,7 @@ class LocklessQueue {
       data_list_head_ = data_list_head_->next;
       std::unique_ptr<T> data = std::move(node->data);
       PushNodeToList(&free_list_head_, node);
-      return std::move(data);
+      return data;
     }
     return nullptr;
   }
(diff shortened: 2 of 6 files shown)


wohe:
@clalancette I opened #1704 for the remaining focal related fixes. Some of the changes in this PR have already been fixed. I'll send a PR which adds focal to CI later.

Since this is the library without a dependency on ROS it should be possible to have one version for both ROS 1 and 2. The only ROS related file is the `package.xml` which to my understanding allows configuring dependencies differently based on the ROS version. Should we do this? Can you have a look? I dont't think anything would speak against merging a PR which just fixes up the `package.xml` for ROS 2 support.

clalancette:
> @clalancette I opened #1704 for the remaining focal related fixes. Some of the changes in this PR have already been fixed. I'll send a PR which adds focal to CI later.

Great, thanks!

> Since this is the library without a dependency on ROS it should be possible to have one version for both ROS 1 and 2. The only ROS related file is the `package.xml` which to my understanding allows configuring dependencies differently based on the ROS version. Should we do this? Can you have a look? I dont't think anything would speak against merging a PR which just fixes up the `package.xml` for ROS 2 support.

Yep, in theory we should be able to do that.  I'm going to rebase this on top of master, and then see if I can get that working.

clalancette:
It turns out that I made a mistake here.  The `ros2` branch on the `ros2` fork is our actively-released branch, so I can't really make modifications to it without doing a re-release into the ROS world.  Instead of doing that, I'm going to close this PR and open a new one from a separate branch which I can make modifications to at will.  Sorry for the noise.

clalancette:
New PR is #1705

PR #1703, opened 2020-06-14
Use installed Abseil if available (otherwise download it)

barcode:
This PR addresses the same issue as #1570. It enables a user to use an installed version of abseil. The main difference: If there is no installed package of Abseil, the repo will behave as before and download and build Abseil.
Thus, this PR should not mess with existing setups while allowing the user to use an installed version of Abseil.
I hope this prevents the concerns stopping  #1570 from getting merged.

### DIFF
--- a/cmake/modules/FindAbseil.cmake
+++ b/cmake/modules/FindAbseil.cmake
@@ -14,7 +14,7 @@
 
 cmake_minimum_required(VERSION 3.2)
 
-if(NOT TARGET standalone_absl)
+if(NOT Abseil_FOUND)
   set(Abseil_FOUND ON)
   set(ABSEIL_FOUND ON)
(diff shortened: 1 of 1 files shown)


barcode:
@googlebot I signed it!

ojura:
There is also #1387 which I believe is a superior method for downloading and configuring Abseil as 
a Cartographer subproject, since this is done during configure-time. This voids the need to have a list of Abseil targets inside `FindAbseil.cmake`, since all are available natively because `add_subdirectory()` is used instead of the (IMHO) awful `ExternalProject_Add`.

barcode:
The goal of this PR was to change as little as possible while allowing to provide your own version of abseil.
I hope this or any other PR achieving the same effect gets merged soon, since it is bothersome to work on a fork.

wohe:
Closing in favor of #1570.

PR #1704, opened 2020-06-16
Small fixes needed for focal support.

wohe:


ojura:
Wait, you are allowed to return a `unique_ptr` by value without moving from it? I remember seeing somewhere that in general `std::move` on return is a bad thing because it prevents the compiler to do copy elision. But I thought it was necessary with non-copyable types like `unique_ptr`.

wohe:
@ojura See https://en.cppreference.com/w/cpp/language/return or https://isocpp.org/blog/2013/02/no-really-moving-a-return-value-is-easy-stackoverflow: Removing the `std::move` here does not prevent moving.

anchuanxu:
thanks!

PR #1705, opened 2020-06-17
Minor fixes for Cartographer from ROS

clalancette:
This PR is a "port" of the Cartographer code to ROS 2. I put the words port in quotes because the port doesn't require very much; it is mostly changes to compile on Ubuntu 20.04, plus some changes to the package.xml to make it a pure cmake package rather than a catkin one.

As-is, this PR doesn't apply to master. That's because this "port" was originally done against the 1.0.0 tag of this repository. However, given the rather simple nature of this PR, it should be pretty easy to move forward.

I'm opening this PR as a starting point for the discussion on how to move this code into the upstream. Some steps that seem obvious to me:

* ~~Update this to apply against master.~~ This is now done.
* ~~Create a ros2 branch for this to target~~ We decided to try to merge this all into master.

Some steps that are less obvious, and/or need discussion:

* ~~Do we want to try to support ROS 1 and ROS 2 on the same branch? It might be possible here, but I'm not sure.~~  We decided we do.
* Is the cartographer project willing to take over maintenance of this branch and release it into newer ROS 2 versions as they come along?

Finally, I'll note that I don't have too much of time to devote to this. I'm doing this on discretionary time, so while I'm glad to do some simple work here, I won't have the time to do any major refactoring.

Comments, questions, thoughts all welcome!

This was originally PR #1701, but I had to close that one and move it to a new source branch.  Below are comments from that PR.

---

@wohe said:

@clalancette I opened #1704 for the remaining focal related fixes. Some of the changes in this PR have already been fixed. I'll send a PR which adds focal to CI later.

Since this is the library without a dependency on ROS it should be possible to have one version for both ROS 1 and 2. The only ROS related file is the package.xml which to my understanding allows configuring dependencies differently based on the ROS version. Should we do this? Can you have a look? I dont't think anything would speak against merging a PR which just fixes up the package.xml for ROS 2 support.

---

@clalancette said:

>    @clalancette I opened #1704 for the remaining focal related fixes. Some of the changes in this PR have already been fixed. I'll send a PR which adds focal to CI later.

Great, thanks!

>    Since this is the library without a dependency on ROS it should be possible to have one version for both ROS 1 and 2. The only ROS related file is the package.xml which to my understanding allows configuring dependencies differently based on the ROS version. Should we do this? Can you have a look? I dont't think anything would speak against merging a PR which just fixes up the package.xml for ROS 2 support.

Yep, in theory we should be able to do that. I'm going to rebase this on top of master, and then see if I can get that working.

### DIFF
--- a/cmake/modules/FindGMock.cmake
+++ b/cmake/modules/FindGMock.cmake
@@ -77,6 +77,7 @@ if(NOT GMock_FOUND)
           EXCLUDE_FROM_ALL)
       endif()
       set(GMOCK_LIBRARIES gmock_main)
+      set(GMOCK_INCLUDE_DIRS ${GMOCK_SRC_DIR}/include)
     endif()
   endif()
 
--- a/package.xml
+++ b/package.xml
@@ -38,6 +38,7 @@
 
   <build_depend>git</build_depend>
   <build_depend>google-mock</build_depend>
+  <build_depend>gtest</build_depend>
   <build_depend>python3-sphinx</build_depend>
 
   <depend>libboost-iostreams-dev</depend>
--- a/package.xml
+++ b/package.xml
@@ -40,7 +40,7 @@
   <build_depend>google-mock</build_depend>
   <build_depend>python3-sphinx</build_depend>
 
-  <depend>boost</depend>
+  <depend>libboost-iostreams-dev</depend>
   <depend>eigen</depend>
   <depend>libcairo2-dev</depend>
   <depend>libceres-dev</depend>


clalancette:
I've now rebased this on top of master, and done some additional fixes to make this compatible with both ROS 1 and ROS 2.  At this point, I think it is ready for review.

wohe:
Could you rebase again? I added some changes for Ubuntu 20.04 support in #1706 and it is now also included in CI.

clalancette:
Rebased again now.

cmake/functions.cmake
```diff
@@ -139,5 +139,11 @@ endmacro()
 
 macro(google_enable_testing)
   enable_testing()
+  # For Ubuntu 20.04, finding gmock always throws a warning about
```
wohe, line 142: Are these changes still needed? On CI it works without them when installing the `libgmock-dev` package.

clalancette: Without this particular change it does work, it just always throws a warning.  I was trying to quiet that down, but I can remove it if you don't think we should have it.

package.xml
```diff
@@ -38,12 +38,14 @@
 
   <build_depend>git</build_depend>
   <build_depend>google-mock</build_depend>
-  <build_depend>python-sphinx</build_depend>
+  <build_depend>gtest</build_depend>
+  <build_depend condition="$ROS_PYTHON_VERSION == 2">python-sphinx</build_depend>
+  <build_depend condition="$ROS_PYTHON_VERSION == 3">python3-sphinx</build_depend>
```
MichaelGrupp, line 43: If this goes to master, does it work with ROS 1?

clalancette: > If this goes to master, does it work with ROS 1?

Yeah, it should.  Both Melodic and Noetic support package format 3, and thus support `build_depend` with a condition.  At least in my testing on Melodic, I can both build the package and install the (Python 2) dependencies via `rosdep`.

MichaelGrupp: cool, I just haven't seen this condition thing yet.

MichaelGrupp:
Can we just use `python3-sphinx` in all cases? It's available also for previous Ubuntu distros and we just need its executable as a build dependency for documentation. It doesn't interfere with the actual ROS stuff in Cartographer.

clalancette:
> Can we just use `python3-sphinx` in all cases? It's available also for previous Ubuntu distros and we just need its executable as a build dependency for documentation. It doesn't interfere with the actual ROS stuff in Cartographer.

Sure, that works for me.  See e1e5212

cmake/modules/FindGMock.cmake
```diff
@@ -77,6 +77,7 @@ if(NOT GMock_FOUND)
           EXCLUDE_FROM_ALL)
       endif()
       set(GMOCK_LIBRARIES gmock_main)
+      set(GMOCK_INCLUDE_DIRS ${GMOCK_SRC_DIR}/include)
```
MichaelGrupp, line 80: +1
I'm working on migrating our fork to Noetic, this was needed there too.

cmake/functions.cmake
wohe: Yes, I think this needs be removed.

cmake/modules/FindGMock.cmake
wohe: Did you try installing the `libgmock-dev` package? I added Ubuntu Focal (which Noetic is primarily targeted at) to CI recently and it passes.

package.xml
```diff
@@ -38,12 +38,13 @@
...
-  <build_depend>python-sphinx</build_depend>
+  <build_depend>gtest</build_depend>
+  <build_depend>python3-sphinx</build_depend>
 
-  <depend>boost</depend>
-  <depend>ceres-solver</depend>
+  <depend>libboost-iostreams-dev</depend>
   <depend>eigen</depend>
   <depend>libcairo2-dev</depend>
+  <depend>libceres-dev</depend>
```
wohe, line 47: This does not work. For example Kinetic is still supported until April 2021 which does not work with this change. How about making this change only for ROS2 and keeping the `ceres-solver` for ROS1?

clalancette: This isn't actually a ROS 1 vs. ROS 2 thing.  It's more of a Linux distribution thing.

For what it is worth, the `libceres-dev` rosdep key is available for Kinetic (Ubuntu distribution Xenial): https://github.com/ros/rosdistro/blob/d8a1a79b5958aaac468d665cb99eb4c3d68c4b09/rosdep/base.yaml#L1950

If we look in Ubuntu, we see that that maps to this package version: https://packages.ubuntu.com/xenial/libceres-dev .

What I don't know off-hand is whether libceres-dev 1.11.0 is new enough to compile the latest cartographer against.  Do you happen to know?

cmake/modules/FindGMock.cmake
clalancette: So installing `libgmock-dev` does make it work.  But I actually still think this is a bug, and installing `libgmock-dev` just works around it.

When you install `libgmock-dev`, it means you actually don't come in the `if(NOT GMOCK_LIBRARIES)` block at all, and hence don't build gmock from source.  If for some reason you *do* need to build gmock from source, then this block still doesn't export the GMOCK_INCLUDE_DIRS properly.

Thus, I think we should either keep this fix, or we should remove the fallback to build from source completely.

package.xml
wohe: The reason was that `libceres-dev` in Xenial was packaged incorrectly and does not contain the CMake config files. Later Ubuntu versions have this, so we use `libceres-dev` starting at Bionic.

cmake/functions.cmake
clalancette: All right, removed in 864edf6.

package.xml
clalancette: All right, that makes sense.

The bit that still doesn't quite make sense to me is the `ceres-solver` key, though.  How is that key being satisfied for Xenial?  There is no entry in https://github.com/ros/rosdistro/blob/master/rosdep/base.yaml or https://github.com/ros/rosdistro/blob/master/kinetic/distribution.yaml for it.

bochen87: Xenial is anyways EOL in 7 months (April 2021) - maybe we just drop support early for Xenial and for anybody who is interested in Xenial we make a [tag](https://github.com/cartographer-project/cartographer/tags) prior to merging this?

wohe: Let's not do this:
1. Ubuntu Xenial and ROS Kinetic are still supported. I think the most important CI this needs to pass is the one in cartographer_ros for the various ROS1 versions still supported.
2. What is the rationale behind this change? We build Ceres for compatibility with old versions for now, but is this actually causing any issues in ROS2? I think the focus of this PR should strictly be adding support for ROS2 in this PR. Other changes should go to separate PRs.

clalancette: > * What is the rationale behind this change? We build Ceres for compatibility with old versions for now, but is this actually causing any issues in ROS2? I think the focus of this PR should strictly be adding support for ROS2 in this PR. Other changes should go to separate PRs.

This is causing problems in ROS, period.  As it is, we cannot release cartographer into ROS 1 or ROS 2.  The reason is that the ROS buildfarm requires that all dependency keys be resolvable to actual package names (either from the system packages, or from elsewhere in the ROS ecosystem).  Since `ceres-solver` doesn't resolve to anything in ROS 1 or ROS 2, we can't release it.

We get around this currently by patching the ceres-solver key out when doing releases for ROS 1 (see https://github.com/ros-gbp/cartographer-release/blob/master/kinetic/package.xml for the package.xml we "overlay" on the sources), and by having a forked cartographer for ROS 2 (https://github.com/ros2/cartographer, the fork we are currently trying to get rid of).  However, the best way to solve this is to make the key resolvable in the upstream, which is what this patch is trying to do.

>     1. Ubuntu Xenial and ROS Kinetic are still supported. I think the most important CI this needs to pass is the one in cartographer_ros for the various ROS1 versions still supported.

I looked into how this is working for Kinetic a bit more.  We are hosting an upgraded version of libceres on our apt repository: http://packages.ros.org/ros-testing/ubuntu/pool/main/c/ceres-solver/ .  With that in place, we are actually able to resolve the `libceres-dev` key to a proper package name on Xenial, and hence we can release cartographer into Kinetic.  Thus, I believe that CI will continue to work with Xenial even after this change (travis seems to agree with me here).

wohe: Thanks for the context, this really helps everyone understand what we are looking into here. I verified that we can install the ROS-provided libceres-dev to make it work. Unfortunately when I start a docker build for kinetic, I get an error. Could it be that the current package.xml is missing a dependency? The error unfortunately does not mention which file is missing:
```
 OSError: Failed command '['/opt/cartographer_ros/env.sh', 'cmake', '/catkin_ws/src/cartographer_ros/cartographer_ros_msgs', '-DCATKIN_DEVEL_PREFIX=/catkin_ws/devel_isolated/cartographer_ros_msgs', '-DCMAKE_INSTALL_PREFIX=/opt/cartographer_ros', '-G', 'Ninja']': [Errno 2] No such file or directory
```

I think we need to investigate more. Once this is resolved, I think it is okay to break existing users of cartographer_ros with such a change as we then know how to fix it quickly in cartographer_ros. Without a follow up PR already tested and ready to go, I think we would let down existing users.

Different topic: I might be mistaken, but it seems to me that you are frustrated about the slow progress ("This is causing problems in ROS, period.") but given that everyone here is contributing voluntarily, this will only proceed when someone is willing to put in the effort to move it forward, as I just did the last hour in trying it out to see whether it works. I think this is the only blocker really and to my knowledge everyone would love to see ROS2 supported as soon as we can.

wohe: My investigation so far mostly for Kinetic:
1. All the Catkin packages in our ROS CI use `/opt/cartographer_ros/env.sh`. Not sure where this is configured/determined and why. Does someone know?
2. If we build Ceres first which is a non-Catkin package, it install it, so everything works.
3. If we do not build Ceres we fail, but if we build Cartographer first it would fix it.
4. Not sure whether this would fix all issues, especially when building outside of CI. After all we do not want to fix CI by hiding existing issues. Any one understands enough of Catkin to know what is going on?
5. Melodic/Noetic: also stops working, same problem. What I do not understand and have not investigated yet: why do  they work before but not with the change? It seems before they somehow know not to use the non-existing `env.sh`. 

Otherwise, my suggestion would be the following: on Monday I create a PR that fixes Cartographer ROS by reordering the package builds and verify it works. Then we merge the Cartographer changes, quickly followed by the Cartographer ROS PR.

clalancette: > Thanks for the context, this really helps everyone understand what we are looking into here.

You are welcome, happy to provide it.

> Could it be that the current package.xml is missing a dependency? The error unfortunately does not mention which file is missing:

Hm, that seems like it is either missing cmake, or the initial `/opt/cartographer_ros/env.sh` file.  Not sure what is going on there.

> I think we need to investigate more. Once this is resolved, I think it is okay to break existing users of cartographer_ros with such a change as we then know how to fix it quickly in cartographer_ros. Without a follow up PR already tested and ready to go, I think we would let down existing users.

Sorry, I don't think I understand this bit.  I don't *think* that making the change in this PR will cause any problems for downstream users, but maybe I'm misunderstanding.

> Different topic: I might be mistaken, but it seems to me that you are frustrated about the slow progress ("This is causing problems in ROS, period.")

No, I'm not frustrated.  I was just emphasizing that we need this change in both ROS 1 and ROS 2.  No worries, I am doing this on volunteer time too :).

wohe: Okay. I think I understand it now:
1. We build package explicitly in a specific order in our cartographer_ros CI to allow Docker to cache results. This has the side effect that if the order which Catkin chooses changes, we have to adapt CI.
2. When we remove the Ceres packages Catkin changes the order in which packages are built which breaks CI. This is even true for the newer ROS versions where the package is just present but was not actually compiled by CI.
3. Especially we have to use the ordering Catkin uses, not just any ordering which would work according to the dependencies. In our case, we must built cartographer before cartographer_ros_msgs even though the latter does not depend on the former.

clalancette:
Thanks for pushing forward with this.  I've now rebased this branch.  This still has a few fixes that would be good to get in, so I'll still suggest that we look into merging this as well.

clalancette:
Friendly ping on this one; can we get the fixes in this PR reviewed?  Thanks.

package.xml
MichaelGrupp: Resolved - we depend on libceres-dev now (hence no diff anymore)

MichaelGrupp (approved):
So, we now have 3 lines left to review. One big discussion about libceres-dev is already resolved. Everything else looks fine to me, so I would prefer to merge. Since he was more involved into the discussion, I would ask @wohe to take a look please.

MichaelGrupp:
Please rebase / update branch and then let's get this merged. This patch is also needed for users that are building ROS1 from source, see my experiment in https://github.com/cartographer-project/cartographer_ros/issues/1531. As we can see from the reply I got there it really affects users.

So unless @wohe has strong objections I would proceed to get things fixed.

wohe:
@MichaelGrupp I will have a look today. We need to make sure that this does not break anything ROS1 related.

clalancette:
> Please rebase / update branch and then let's get this merged. This patch is also needed for users that are building ROS1 from source, see my experiment in [cartographer-project/cartographer_ros#1531](https://github.com/cartographer-project/cartographer_ros/issues/1531). As we can see from the reply I got there it really affects users.

I've rebased onto the latest.

> @MichaelGrupp I will have a look today. We need to make sure that this does not break anything ROS1 related.

Agreed.  I think that CI should show this, however.

wohe (approved):

wohe:
@clalancette Unfortunately, CI for this repo is unaffected by the file you changed. All ROS related code is in cartographer_ros. There, CI of course would fail when we introduce issues, but only after we merged the PR. I now ran this CI manually with your branch and it all succeeded.

clalancette:
> @clalancette Unfortunately, CI for this repo is unaffected by the file you changed. All ROS related code is in cartographer_ros. There, CI of course would fail when we introduce issues, but only after we merged the PR. I now ran this CI manually with your branch and it all succeeded.

Ah, good call.  Thanks for kicking off the tests and merging!

PR #1706, opened 2020-06-18
Add Ubuntu 20.04 to the install scripts, CI, docs.

wohe:

### DIFF
--- a/scripts/install_debs_cmake.sh
+++ b/scripts/install_debs_cmake.sh
@@ -45,11 +45,23 @@ sudo apt-get install -y \
     libgoogle-glog-dev \
     liblua5.2-dev \
     libsuitesparse-dev \
-    ninja-build \
-    python-sphinx
+    ninja-build
+
+if [[ "$(lsb_release -sc)" = "focal" ]]
+then
+    sudo apt-get install -y python3-sphinx libgmock-dev
+else
+    sudo apt-get install -y python-sphinx
+fi
 
 # Install Ceres Solver on Ubuntu Bionic. No need to build it ourselves.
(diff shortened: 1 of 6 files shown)


scripts/install_debs_cmake.sh
```diff
@@ -45,11 +45,23 @@ sudo apt-get install -y \
...
 if [[ "$(lsb_release -sc)" = "bionic" ]]
 then
   sudo apt-get install -y libceres-dev
 fi
+
+# Install Ceres Solver and Protocol Buffers support on Ubuntu Focal. No need to build it ourselves.
+if [[ "$(lsb_release -sc)" = "focal" ]]
+then
+  sudo apt-get install -y libceres-dev protobuf-compiler
+fi
```
feuerste, line 67: How about adding this to line 52?

wohe: My intention was to only list the two libraries which influence the later build steps here, i.e. starting from Focal, we also do not need to compile protobuf support. This replaces these lines from older Dockerfiles:
```
COPY scripts/install_ceres.sh cartographer/scripts/
RUN cartographer/scripts/install_ceres.sh && rm -rf ceres-solver
COPY scripts/install_proto3.sh cartographer/scripts/
RUN cartographer/scripts/install_proto3.sh && rm -rf protobuf
```

Line 52 has no such impact. Does that make sense?

feuerste: It's fine with me to keep them separate, too. When first reading over it I was just confused by the fact that we had an `if` for `focal` twice...

Issue #1709, opened 2020-06-23
Remove legacy serialization format migration code?


MichaelGrupp:
There's a bunch of code that was useful for migrating old pbstream files to the format that was established ~2 years ago with v1.0. ([serialization_format_migration](https://github.com/cartographer-project/cartographer/blob/master/cartographer/io/serialization_format_migration.h), `pbstream migrate` CLI, protos...).

We can consider if it makes sense to keep all of this or if we can clean up. I assume the amount of people who still rely on pre-1.0 pbstream files should be low.

wohe:
I looked into this and it seems to be two different things:

1. There is a legacy format and a tool `pbstream_main migrate`. My understanding is that old maps cannot be loaded before migration. I think it is safe to assume that everyone already converted all interesting maps, and if not they can use any version of Cartographer from the last year to do it. I would say it is safe to remove this now.

2. There is the histogram backwards compatibility from #1710 which seems to be fixed on the fly and _only_ there, with no migration tool available. Which means if someone is using old maps they just work right now and are migrated each time they are loaded. If we remove this support we would break people and there is no migration tool. We might still want to do it for the reasons mentioned in the PR, but we should at least be prepared to add a migration tool if needed.

@schwoere FYI, in case you have more info.

wohe:
FYI I will have a look at changing the migration tool as explained above: remove the legacy code to migrate to the new stream format, add new code to migrate to submaps with histograms.

PR #1710, opened 2020-06-23
Deactivate memory-critical histogram backwards compatibility.

MichaelGrupp:
It's essentially leading to buffering all submaps twice, which
is a high (~2x) peak in memory consumption during deserialization.

Don't convert anymore since this affects _all_ pbstreams and not only
the (probably few) old pbstreams that require the conversion.

### DIFF
--- a/cartographer/mapping/map_builder.cc
+++ b/cartographer/mapping/map_builder.cc
@@ -285,8 +285,12 @@ std::map<int, int> MapBuilder::LoadState(
                                  true);
   }
 
-  MapById<SubmapId, mapping::proto::Submap> submap_id_to_submap;
-  MapById<NodeId, mapping::proto::Node> node_id_to_node;
+  CHECK_NE(deserializer.header().format_version(),
+           io::kFormatVersionWithoutSubmapHistograms)
+      << "The pbstream file contains submaps without rotational histograms. "
+         "This can be converted, but the conversion was removed here due to "
+         "memory concerns.";
+
   SerializedData proto;
   while (deserializer.ReadNextSerializedData(&proto)) {
     switch (proto.data_case()) {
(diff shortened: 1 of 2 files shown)


MichaelGrupp:
We used this patch since a year, it has a big impact. Here's a heap profiling I did back then:

### before
image

### after
image

cartographer/mapping/map_builder.cc
```diff
@@ -285,8 +285,12 @@ std::map<int, int> MapBuilder::LoadState(
                                  true);
   }
 
-  MapById<SubmapId, mapping::proto::Submap> submap_id_to_submap;
-  MapById<NodeId, mapping::proto::Node> node_id_to_node;
+  CHECK_NE(deserializer.header().format_version(),
+           io::kFormatVersionWithoutSubmapHistograms)
+      << "The pbstream file contains submaps without rotational histograms. "
+         "This can be converted, but the conversion was removed here due to "
+         "memory concerns.";
```
MichaelGrupp, line 292: This check message would need some changes probably.

cartographer/mapping/map_builder.cc
```diff
@@ -361,20 +365,6 @@ std::map<int, int> MapBuilder::LoadState(
     }
   }
 
-  // TODO(schwoere): Remove backwards compatibility once the pbstream format
-  // version 2 is established.
-  if (deserializer.header().format_version() ==
-      io::kFormatVersionWithoutSubmapHistograms) {
-    submap_id_to_submap =
-        cartographer::io::MigrateSubmapFormatVersion1ToVersion2(
```
MichaelGrupp, line 369: Do we still need this conversion in the library? I also created issue #1709 regarding this.

cartographer/mapping/map_builder.cc
wohe: Might be best to mention the migration tool here. Also: could we change this to only fail for 3D data? As far as I understand nothing has changed for 2D, so old maps should just continue to work.

MichaelGrupp: done

cartographer/mapping/map_builder.cc
```diff
@@ -303,10 +309,11 @@ std::map<int, int> MapBuilder::LoadState(
         proto.mutable_submap()->mutable_submap_id()->set_trajectory_id(
             trajectory_remapping.at(
                 proto.submap().submap_id().trajectory_id()));
-        submap_id_to_submap.Insert(
+        const auto submap_id =
```
wohe, line 312: nit: Isn't it more readable to write `const SubmapId submap_id( ... )` similar to line 322?

MichaelGrupp:
done

PR #1711, opened 2020-06-24
Small Abseil fixes for dependents.

wohe:
Enable PIC so that shared libraries can use the Cartographer
library.

Fixes the config so that dependent builds can find Abseil.


wohe:
This is a follow up of #1570.

FYI @bochen87 @ojura

bochen87:
> This is a follow up of #1570.
> 
> FYI @bochen87 @ojura

The build for cartographer_ros seems to be broken since #1570  - does this PR fix it?

bochen87:
ok, great! :-)

PR #1713, opened 2020-07-06
Remove Debian Jessie from CI.

wohe:
It has reached end-of-life with the
end of LTS on June 30, 2020.


wohe:
See https://wiki.debian.org/LTS.

PR #1714, opened 2020-07-08
Small improvements to 2D fixed frame pose support.

wohe:
This is a follow up to #1580.


wohe:
FYI @jihoonl

jihoonl:
Got it. Thanks for taking care of pending PRs.

wohe:
Thanks for the review. I don't think there is a reviewer with more context, unfortunately.

PR #1723, opened 2020-07-22
Update package.xml for ROS Noetic.

wohe:


wohe:
FYI @MichaelGrupp 

This is to address the issue mentioned at cartographer-project/cartographer_ros#1494
Verified to work with ROS Kinetic and Melodic.
See also #1705.

PR #1724, opened 2020-07-23
Only set C++11 explicitly for old versions of GCC.

wohe:
Starting with GCC 6.1, C++14 became the default.


wohe:
This is to help cartographer-project/cartographer_ros#1494.
FYI @MichaelGrupp

PR #1731, opened 2020-07-29
Extrapolator Result for motion compensation

schwoere:
Signed-off-by: mschworer <mschworer@lyft.com>

This makes use of the `ExtrapolationResult` returned by the pose extrapolator. `ExtrapolationResult` contains poses corresponding to the timestamps of range data measurements to be used for motion compensation. When using the `PoseExtrapolator` implementation there is no functional difference. This PR prepares the integration of `ImubasedPoseExtrapolator`, which requires using `ExtrapolationResult` returned by `ExtrapolatePosesWithGravity ` to run efficiently.

### DIFF
--- a/cartographer/mapping/internal/3d/local_trajectory_builder_3d.cc
+++ b/cartographer/mapping/internal/3d/local_trajectory_builder_3d.cc
@@ -211,12 +211,23 @@ LocalTrajectoryBuilder3D::AddRangeData(
   last_sensor_time_ = current_sensor_time;
 
   const common::Time current_time = hit_times.back();
+  const auto voxel_filter_start = std::chrono::steady_clock::now();
   const sensor::RangeData filtered_range_data = {
       extrapolation_result.current_pose.translation().cast<float>(),
       sensor::VoxelFilter(options_.voxel_filter_size())
           .Filter(accumulated_range_data.returns),
       sensor::VoxelFilter(options_.voxel_filter_size())
           .Filter(accumulated_range_data.misses)};
+  const auto voxel_filter_stop = std::chrono::steady_clock::now();
+  const auto voxel_filter_duration = voxel_filter_stop - voxel_filter_start;
+
+  if (sensor_duration.has_value()) {
+    const double voxel_filter_fraction =
+        common::ToSeconds(voxel_filter_duration) /
+        common::ToSeconds(sensor_duration.value());
+    kLocalSlamVoxelFilterFraction->Set(voxel_filter_fraction);
+  }
+
   return AddAccumulatedRangeData(
       current_time,
       sensor::TransformRangeData(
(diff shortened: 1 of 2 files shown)


cartographer/mapping/internal/3d/local_trajectory_builder_3d.cc
```diff
@@ -134,111 +133,123 @@ LocalTrajectoryBuilder3D::AddRangeData(
...
-  std::vector<transform::Rigid3f> hits_poses;
-  hits_poses.reserve(hits.size());
-  bool warned = false;
-
-  for (const auto& hit : hits) {
-    common::Time time_point =
-        current_sensor_time + common::FromSeconds(hit.point_time.time);
-    if (time_point < extrapolator_->GetLastExtrapolatedTime()) {
-      if (!warned) {
-        LOG(ERROR)
```
schwoere, line 157: @wohe shall I keep this warning?

cartographer/mapping/internal/3d/local_trajectory_builder_3d.cc
```diff
@@ -134,111 +133,123 @@ LocalTrajectoryBuilder3D::AddRangeData(
...
 }
 
 std::unique_ptr<LocalTrajectoryBuilder3D::MatchingResult>
 LocalTrajectoryBuilder3D::AddAccumulatedRangeData(
     const common::Time time,
     const sensor::RangeData& filtered_range_data_in_tracking,
-    const absl::optional<common::Duration>& sensor_duration) {
+    const absl::optional<common::Duration>& sensor_duration,
+    const transform::Rigid3d& pose_prediction,
+    const Eigen::Vector3d& current_velocity,
```
wohe, line 247: Why is `current_velocity` given as an argument? It seems to be unused.

cartographer/mapping/internal/3d/local_trajectory_builder_3d.cc
```diff
@@ -134,111 +133,123 @@ LocalTrajectoryBuilder3D::AddRangeData(
...
-        LOG(ERROR)
-            << "Timestamp of individual range data point jumps backwards from "
-            << extrapolator_->GetLastExtrapolatedTime() << " to " << time_point;
-        warned = true;
-      }
-      time_point = extrapolator_->GetLastExtrapolatedTime();
-    }
-    hits_poses.push_back(
-        extrapolator_->ExtrapolatePose(time_point).cast<float>());
+  // !!!!
```
wohe, line 154: ?

cartographer/mapping/internal/3d/local_trajectory_builder_3d.cc
wohe: Seems like a good idea. Also the handling of time like in #936. Or is there some reason this cannot happen anymore?

schwoere: Ok I implemented something equivalent.

cartographer/mapping/internal/3d/local_trajectory_builder_3d.cc
schwoere: Good catch, that was unrelated. Removed it.

cartographer/mapping/internal/3d/local_trajectory_builder_3d.cc
schwoere: removed it.

PR #1314, opened 2018-07-20
Migrate pbstream files with old probability grid format.

MichaelGrupp:
Extends the range of old v0.3 pbstream files that can be migrated.
Adds support for files that contain 2D submaps with the old, non-generalized grid format.

Just specify the `--migrate_grid_format` flag when calling
`cartographer_pbstream migrate`.

PR #1733, opened 2020-08-03
Change pbstream migrate to support new submap format.

wohe:
This changes the migration support to:
1. Remove migration from old files from before version 1.0.
   If this is still needed by someone, they can use the tool
   as it was before this change.
2. Adds migration of maps without submap histograms. These
   were so far migrated on-the-fly, but #1710 aims to remove
   this.

See also #1709.

This was tested by converting a serialized 3D map and verifying
it can still be used for localization. The file changed a little
bit, but it seems to be good enough.

### DIFF
--- a/cartographer/io/internal/pbstream_migrate.cc
+++ b/cartographer/io/internal/pbstream_migrate.cc
@@ -21,19 +21,18 @@
 #include "gflags/gflags.h"
 #include "glog/logging.h"
 
-DEFINE_bool(migrate_grid_format, false,
-            "Set if the submap data of the input pbstream uses the old "
-            "probability grid format.");
+DEFINE_bool(include_unfinished_submaps, true,
+            "Whether to write to include unfinished submaps in the output.");
 
 namespace cartographer {
 namespace io {
 
 int pbstream_migrate(int argc, char** argv) {
   std::stringstream ss;
-  ss << "\n\nTool for migrating files that use the serialization output of "
-        "Cartographer 0.3, to the new serialization format, which includes a "
-        "header (Version 1). You may need to specify the '--migrate_grid_format"
-        " flag if the input file contains submaps with the legacy grid format."
+  ss << "\n\nTool for migrating files that use submaps without histograms "
+        "to the new submap format, which includes a histogram. You can "
+        "set --include_unfinished_submaps to false if you want to exclude "
+        "unfinished submaps in the output."
      << "\nUsage: " << argv[0] << " " << argv[1]
      << " <input_filename> <output_filename> [flags]";
   google::SetUsageMessage(ss.str());
@@ -44,10 +43,10 @@ int pbstream_migrate(int argc, char** argv) {
   }
   cartographer::io::ProtoStreamReader input(argv[2]);
   cartographer::io::ProtoStreamWriter output(argv[3]);
-  LOG(INFO) << "Migrating old serialization format in \"" << argv[2]
-            << "\" to new serialization format in \"" << argv[3] << "\"";
-  cartographer::io::MigrateStreamFormatToVersion1(&input, &output,
-                                                  FLAGS_migrate_grid_format);
+  LOG(INFO) << "Migrating serialization format 1 in \"" << argv[2]
+            << "\" to serialization format 2 in \"" << argv[3] << "\"";
+  cartographer::io::MigrateStreamVersion1ToVersion2(
+      &input, &output, FLAGS_include_unfinished_submaps);
   CHECK(output.Close()) << "Could not write migrated pbstream file to: "
                         << argv[3];
 
--- a/cartographer/io/internal/pbstream_migrate.cc
+++ b/cartographer/io/internal/pbstream_migrate.cc
@@ -22,7 +22,7 @@
 #include "glog/logging.h"
 
 DEFINE_bool(include_unfinished_submaps, true,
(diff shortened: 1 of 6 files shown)


wohe:
FYI @MichaelGrupp @schwoere

wohe:
The code is adapted from `MapBuilder::LoadState()` at https://github.com/cartographer-project/cartographer/blob/master/cartographer/mapping/map_builder.cc.

cartographer/io/serialization_format_migration.cc
```diff
@@ -18,284 +18,200 @@
...
 #include "cartographer/mapping/3d/submap_3d.h"
+#include "cartographer/mapping/id.h"
+#include "cartographer/mapping/internal/3d/pose_graph_3d.h"
 #include "cartographer/mapping/internal/3d/scan_matching/rotational_scan_matcher.h"
+#include "cartographer/mapping/map_builder.h"
+#include "cartographer/mapping/map_builder_interface.h"
+#include "cartographer/mapping/pose_graph.h"
 #include "cartographer/mapping/probability_values.h"
-#include "cartographer/mapping/proto/internal/legacy_serialized_data.pb.h"
-#include "cartographer/mapping/proto/internal/legacy_submap.pb.h"
```
MichaelGrupp, line 26: You can remove the protos in the `mapping/proto/internal` folder. I added them in #1314 for the sole purpose of this migration tool.

MichaelGrupp:
There's also a documentation page: https://github.com/cartographer-project/cartographer/blob/master/docs/source/pbstream_migration.rst

cartographer/io/serialization_format_migration.cc
wohe: Thanks for the suggestion. I'll do this in a follow up PR.

wohe:
@MichaelGrupp I updated the documentation to reflect the change in the tool. Even though it is now documented, it probably is best to point people towards the tool when loading a submap fails in #1710.

MichaelGrupp:
lgtm

cartographer/io/internal/pbstream_migrate.cc
```diff
@@ -21,19 +21,18 @@
 #include "gflags/gflags.h"
 #include "glog/logging.h"
 
-DEFINE_bool(migrate_grid_format, false,
-            "Set if the submap data of the input pbstream uses the old "
-            "probability grid format.");
+DEFINE_bool(include_unfinished_submaps, true,
+            "Whether to write to include unfinished submaps in the output.");
```
feuerste, line 25: Nit: remove either `to write` or `to include`?

cartographer/io/serialization_format_migration.cc
```diff
@@ -18,284 +18,200 @@
...
-  return header;
-}
+  mapping::MapById<mapping::SubmapId, mapping::proto::Submap>
+      submap_id_to_submap;
+  mapping::MapById<mapping::NodeId, mapping::proto::Node> node_id_to_node;
+  SerializedData proto;
+  while (deserializer.ReadNextSerializedData(&proto)) {
+    switch (proto.data_case()) {
+      case SerializedData::kPoseGraph:
+        LOG(ERROR) << "Found multiple serialized `PoseGraph`. Serialized "
```
feuerste, line 131: Could we even be more defensive and `LOG(FATAL)`?

cartographer/io/serialization_format_migration.cc
```diff
@@ -18,284 +18,200 @@
...
+                      "stream likely corrupt!.";
+        break;
+      case SerializedData::kAllTrajectoryBuilderOptions:
+        LOG(ERROR) << "Found multiple serialized "
```
feuerste, line 135: Same as above...

cartographer/io/serialization_format_migration.cc
```diff
@@ -18,284 +18,200 @@
...
-      SerializedData::kPoseGraphFieldNumber,
-      SerializedData::kAllTrajectoryBuilderOptionsFieldNumber,
-      SerializedData::kSubmapFieldNumber,
-      SerializedData::kNodeFieldNumber,
-      SerializedData::kTrajectoryDataFieldNumber,
-      SerializedData::kImuDataFieldNumber,
-      SerializedData::kOdometryDataFieldNumber,
-      SerializedData::kFixedFramePoseDataFieldNumber,
-      SerializedData::kLandmarkDataFieldNumber};
+  // TODO(schwoere): Remove backwards compatibility once the pbstream format
```
feuerste, line 196: Nit: Will @schwoere actually ever work on this code or would it make sense to assign it to yourself?

cartographer/io/internal/pbstream_migrate.cc
wohe: Done.

cartographer/io/serialization_format_migration.cc
wohe: Done.

cartographer/io/serialization_format_migration.cc
wohe: Style is to name the person with most context which is Martin in this case, not to assign them. See https://google.github.io/styleguide/cppguide.html#TODO_Comments. This also has the benefit that we do not have to speculate about the future. We could remove his name from the TODO but I do not see a benefit.

PR #1736, opened 2020-08-20
Remove legacy compatibility proto files.

MichaelGrupp:
Not needed anymore after the recent removal of that feature, see #1709.


cartographer/mapping/map_builder.cc
```diff
@@ -28,7 +28,6 @@
 #include "cartographer/mapping/internal/3d/pose_graph_3d.h"
 #include "cartographer/mapping/internal/collated_trajectory_builder.h"
 #include "cartographer/mapping/internal/global_trajectory_builder.h"
-#include "cartographer/mapping/proto/internal/legacy_serialized_data.pb.h"
```
MichaelGrupp, line 31: This is a leftover of #1165 and was only needed temporarily here in MapBuilder (removed in https://github.com/cartographer-project/cartographer/commit/73d18e5fc54bfa4e4e61fd7feb615b46834aa584)

PR #1743, opened 2020-08-28
Updated the ROS on Windows feed URL

seanyen:
Hi! I am from `ROS on Windows` team and we are currently migrating the feed URL. [Link](https://discourse.ros.org/t/reminder-https-aka-ms-ros-public-the-new-chocolatey-feed-url-for-ros-on-windows/16098).

This change is to update the URL and let me know if any additional questions! Thanks!

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -19,7 +19,7 @@ jobs:
   timeoutInMinutes: 360
   steps:
   - script: |
-      choco sources add -n=roswin -s https://roswin.azurewebsites.net/api/v2/ --priority 1
+      choco sources add -n=roswin -s https://aka.ms/ros/public --priority 1
       rem Azure VM runs out of space on C:, so use D: for ros and rosdeps
       mkdir D:\opt && mklink /J C:\opt D:\opt
       choco upgrade %ROS_METAPACKAGE% -y


MichaelGrupp:
Hi, we require a DCO. Can you please sign off your commit (see "Details" at the DCO check). Thanks!

seanyen:
@MichaelGrupp Done!

MichaelGrupp (approved):

MichaelGrupp:
@wohe I guess we can merge this. Going further: @ojura Do you know where/if this CI is running? I could imagine that it's not running or needs at least some maintenance. Do you want to take a look?

wohe (approved):

ojura:
Hi, we had the CI set up at larics/cartographer and larics/cartographer_ros. I just pushed to see if still works.

https://github.com/larics/cartographer/runs/1050502975

ojura:
@MichaelGrupp fyi libcartographer pipeline looks like it still works, at least 'till the build step. CMake configure fails because of missing Abseil. This line no longer in azure-pipelines.yml no longer installs all deps, it's missing Abseil:

```
"C:\opt\ros\melodic\x64\env.bat" rosdep install --from-paths src --ignore-src -r -y
```

Abseil is missing from `packages.xml`. Is there a key which can be specified for Abseil as a dependency in packages.xml? @seanyen is Abseil available as a choco ros package? Or we need to install it manually as part of build prep?

MichaelGrupp:
> Is there a key which can be specified for Abseil as a dependency in packages.xml?

AFAIK: not an officially working one. Debian has now packaged [libabsl-dev](https://packages.debian.org/bullseye/libabsl-dev), which is nice, but it is currently in testing and not in Ubuntu either. But this would be irrelevant for Windows anyway. There is also a `ros-melodic-abseil-cpp` package floating around, but it's not really maintained and I think that handling such generic system dependencies via ROS is a bad habit anyway.

So it looks like we still have to rely on a manual installation of abseil unfortunately. (to be honest, I'm starting to wonder if maintaining the Abseil dependency is really still worth it given that besides a handful of string and hash map utils, we mostly use `absl::make_unique` that is [obsolete](https://github.com/abseil/abseil-cpp/blob/master/absl/memory/memory.h#L100) with C++14+ anyway....)

wohe:
@MichaelGrupp The reason to use Abseil is not the small number of utility functions, but the fact that the mutex it provides is much better performing (lower latency!) for an application like SLAM due it providing a suitable amount of fairness which the `std::mutex` does not. It is absolutely critical to good performance.

Issue #1737, opened 2020-08-21
Cached precomputation grids of map are problematic for RAM efficiency in pure localization mode


MichaelGrupp:
_[This is a ticket to explain the motivation behind 2 upcoming PRs.]_

Construction of branch-and-bound submap scan matchers used for constraint search is dispatched dynamically, i.e. only when needed for individual submaps. They are kept afterwards. This is efficient because the precomputed grids at different resolutions only need to be computed once.

However, in localization mode this means that we construct and collect more and more of those precomputed grids - proportional to the explored area and bounded by the max. number of submaps in the frozen map. (driving over the same submaps multiple times doesn’t add new ones).

What I did to tackle this:
* metric for counting these matchers #1738 
* garbage collection of matchers that aren't needed anymore #1745 

The garbage collector has been saving us a large chunk of memory during 24/7 operation in large maps since we deployed it at Magazino last year.

Here's a comparison of two simulation runs we did, where the robot explores the whole area of a large map in 2D localization mode.
![image](https://user-images.githubusercontent.com/8985495/90880616-07ea5f80-e3a9-11ea-99b0-b9878238dbdd.png)
(time axis don't match, but roughly corresponds to explored area in both cases)

PR #1745, opened 2020-08-28
Delete submap scan matchers that are unconstrained after trimming.

MichaelGrupp:
This improves the memory issue in pure localization mode as
mentioned in issue #1737.

### DIFF
--- a/cartographer/mapping/internal/2d/pose_graph_2d.cc
+++ b/cartographer/mapping/internal/2d/pose_graph_2d.cc
@@ -1240,12 +1240,31 @@ void PoseGraph2D::TrimmingHandle::TrimSubmap(const SubmapId& submap_id) {
   // Remove all 'data_.constraints' related to 'nodes_to_remove'.
   {
     std::vector<Constraint> constraints;
+    std::set<SubmapId> unconstrained_submap_ids;
     for (const Constraint& constraint : parent_->data_.constraints) {
       if (nodes_to_remove.count(constraint.node_id) == 0) {
         constraints.push_back(constraint);
+      } else {
+        // A constraint to another submap will be removed, mark it as affected.
+        unconstrained_submap_ids.insert(constraint.submap_id);
       }
     }
     parent_->data_.constraints = std::move(constraints);
+    // Go through the changed constraints and finalize the list of submap IDs
+    // that lost constraints and now have zero inter-submap constraints.
+    for (const Constraint& constraint : parent_->data_.constraints) {
+      if (constraint.tag == Constraint::Tag::INTRA_SUBMAP) {
+        continue;
+      } else if (unconstrained_submap_ids.count(constraint.submap_id)) {
+        // False positive - this submap is still inter-submap-constrained with
+        // some node that is not from the submap that will be trimmed.
+        unconstrained_submap_ids.erase(constraint.submap_id);
+      }
+    }
+    // Delete submap scan matchers of now unconstrained submaps to save memory.
+    for (const SubmapId& submap_id : unconstrained_submap_ids) {
+      parent_->constraint_builder_.DeleteScanMatcher(submap_id);
+    }
   }
 
   // Mark the submap with 'submap_id' as trimmed and remove its data.
--- a/cartographer/mapping/internal/2d/pose_graph_2d.cc
+++ b/cartographer/mapping/internal/2d/pose_graph_2d.cc
@@ -1264,6 +1264,9 @@ void PoseGraph2D::TrimmingHandle::TrimSubmap(const SubmapId& submap_id) {
       }
     }
     // Delete scan matchers of the submaps that lost all constraints.
+    // TODO(wohe): An improvement to this implementation would be to add the
+    // caching logic at the constraint builder which could keep around only
+    // recently used scan matchers.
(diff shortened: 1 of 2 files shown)


MichaelGrupp:
This could be also ported to the 3D equivalent, but I haven't done that yet because I don't use 3D localization and thus haven't tested it..

cartographer/mapping/internal/2d/pose_graph_2d.cc
```diff
@@ -1240,12 +1240,31 @@ void PoseGraph2D::TrimmingHandle::TrimSubmap(const SubmapId& submap_id) {
   // Remove all 'data_.constraints' related to 'nodes_to_remove'.
   {
     std::vector<Constraint> constraints;
+    std::set<SubmapId> unconstrained_submap_ids;
```
wohe, line 1243: Took me a while to understand this. Can we add a comment? This is the set of other submaps to which the `nodes_to_remove` had inter-submap constraints before and which lost at least some of their constraints, right? Maybe renaming is sufficient. My attempt: `other_submap_ids_losing_some_constraints`. What do you think?

cartographer/mapping/internal/2d/pose_graph_2d.cc
```diff
@@ -1240,12 +1240,31 @@ void PoseGraph2D::TrimmingHandle::TrimSubmap(const SubmapId& submap_id) {
...
       if (nodes_to_remove.count(constraint.node_id) == 0) {
         constraints.push_back(constraint);
+      } else {
+        // A constraint to another submap will be removed, mark it as affected.
+        unconstrained_submap_ids.insert(constraint.submap_id);
       }
     }
     parent_->data_.constraints = std::move(constraints);
+    // Go through the changed constraints and finalize the list of submap IDs
+    // that lost constraints and now have zero inter-submap constraints.
```
wohe, line 1254: And here you reduce this list to the submaps which lost all of them. Can we clarify the comment?

wohe (commented):
I'm generally in favor of such a change, but I wonder whether it could be improved. My main concern is that it does not handle corner cases: if a submap scan matcher is constructed but never succeeds in computing a constraint, it will retained forever even with this change, right?

Did you consider the alternative implementation where this is handled at the constraint builder, e.g. by keeping the set of n closest submaps. This would have the additional benefit of starting matcher construction before it is needed further reducing latency.

cartographer/mapping/internal/2d/pose_graph_2d.cc
MichaelGrupp: Changed, I removed "some" as it doesn't add much value IMO.

cartographer/mapping/internal/2d/pose_graph_2d.cc
MichaelGrupp: I simplified all comments, I hope it's clearer now.

MichaelGrupp:
>  if a submap scan matcher is constructed but never succeeds in computing a constraint, it will retained forever even with this change, right?

Yes, I only use the inter-submaps constraints of the localization trajectory like reference counters. As commented above, this could theoretically have the same worst case memory accumulation if no constraints are found at all. But in practice it has a noticeable impact from our experience.

> Did you consider the alternative implementation where this is handled at the constraint builder, e.g. by keeping the set of n closest submaps.

I thought about sth similar, but didn't come up with a good idea. It would potentially require a distance check for each trimmed trajectory node, no? :thinking:

> This would have the additional benefit of starting matcher construction before it is needed further reducing latency.

I don't completely understand how this is related if we could construct before it's needed. Right now the matchers are built in `MaybeAddConstraint` for node-submap combos that are within `max_constraint_distance` ([src](https://github.com/cartographer-project/cartographer/blob/01cb6b1d6f2e67d0678d7ff2fe974d7ab956c49c/cartographer/mapping/internal/constraints/constraint_builder_2d.cc#L76)). I don't think it's a good idea to construct all (?) matchers in a neighborhood in a advance, it would effectively bypass the `constraint_sampling_ratio` that is also crucial for reducing latency. But maybe you mean something else.

cartographer/mapping/internal/2d/pose_graph_2d.cc
```diff
@@ -1238,14 +1238,35 @@ void PoseGraph2D::TrimmingHandle::TrimSubmap(const SubmapId& submap_id) {
...
+        continue;
+      } else if (other_submap_ids_losing_constraints.count(
+                     constraint.submap_id)) {
+        // This submap still has inter-submap constraints - ignore it.
+        other_submap_ids_losing_constraints.erase(constraint.submap_id);
+      }
+    }
+    // Delete scan matchers of the submaps that lost all constraints.
+    for (const SubmapId& submap_id : other_submap_ids_losing_constraints) {
+      parent_->constraint_builder_.DeleteScanMatcher(submap_id);
```
wohe, line 1268: Could we add a `TODO` here to revisit in the future and merge this as it is for now? Something like:
```
TODO(wohe): An improvement to this implementation would be to add the caching logic
at the constraint builder which could keep around only recently used scan matchers.
```
What do you think?

MichaelGrupp: fine for me

wohe (approved):

PR #1750, opened 2020-09-24
Depend on libceres-dev in package.xml for ROS.

wohe:
This changes cartographer_ros to no longer build
Ceres and instead uses libceres-dev for which a
suitable version is provided by ROS for Kinetic.

Related to #1705.

### DIFF
--- a/package.xml
+++ b/package.xml
@@ -41,9 +41,9 @@
   <build_depend>python3-sphinx</build_depend>
 
   <depend>boost</depend>
-  <depend>ceres-solver</depend>
   <depend>eigen</depend>
   <depend>libcairo2-dev</depend>
+  <depend>libceres-dev</depend>
   <depend>libgflags-dev</depend>
   <depend>libgoogle-glog-dev</depend>
   <depend>lua5.2-dev</depend>


wohe:
@MichaelGrupp @bochen87 This will break cartographer_ros CI as explained at #1705. I have a fix ready for cartographer_ros which we have to apply once this got merged.

wohe:
FYI @clalancette

MichaelGrupp (approved):

MichaelGrupp:
Since this should be resolved by rosdep then, you can remove the explicit `libceres-dev` installation calls that I recently added in https://github.com/cartographer-project/cartographer_ros/commit/9cde1b4f5c3c4683df67c86868b11a8e4d31f1e4

2catycm:
Hi， I am trying to install cartographer in Ubuntu20.04 with noetic, and https://google-cartographer-ros.readthedocs.io/en/latest/compilation.html fails at the last step:catkin_make_isolated --install --use-ninja
with message
CMake Error at CMakeLists.txt:39 (find_package):
  By not providing "FindCeres.cmake" in CMAKE_MODULE_PATH this project has
  asked CMake to find a package configuration file provided by "Ceres", but
  CMake did not find one.

  Could not find a package configuration file provided by "Ceres" with any of
  the following names:

    CeresConfig.cmake
    ceres-config.cmake

  Add the installation prefix of "Ceres" to CMAKE_PREFIX_PATH or set
  "Ceres_DIR" to a directory containing one of the above files.  If "Ceres"
  provides a separate development package or SDK, be sure it has been
  installed.

Does that means that I should install ceres manually by apt?

PR #1753, opened 2020-09-30
Random voxel filtering.

wohe:
This leads to slightly slower performance but better quality.
Also replaces the VoxelFilter class by free standing functions.

### DIFF
--- a/cartographer/mapping/internal/2d/local_trajectory_builder_2d.cc
+++ b/cartographer/mapping/internal/2d/local_trajectory_builder_2d.cc
@@ -58,8 +58,8 @@ LocalTrajectoryBuilder2D::TransformToGravityAlignedFrameAndFilter(
                             options_.min_z(), options_.max_z());
   return sensor::RangeData{
       cropped.origin,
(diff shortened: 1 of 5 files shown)


bochen87:
When you say "slightly", did you do some tests? can you express in CPU usage?

wohe:
@bochen87 Yes, of course. Main point here is better quality. Exact performance numbers depend on the sensors used and the environment. For the 3D backpack example voxel filtering takes a few milliseconds and performance regression is a fraction of that.

bochen87:
Ok, cool!

wohe:
Thanks for the review. Some numbers from the 3D demo:
CPU time changes from 1037 s to 1068 s (+3%)
712 loop closures are found instead of 681 (+5%)
translational residuals are 4.5 cm mean instead of 4.9 cm mean (-8%)
rotational residuals are 7.6 mrad means instead of 7.8 mrad (-3%)

bochen87:
thanks for also posting some numbers, very insightful! :+1:

PR #1754, opened 2020-10-06
Don't add IMU data to 2D optimization problem.

MichaelGrupp:
It was constantly added regardless of motion filtering. Since it's not
used it can be just ignored. The method still needs to be there to
satisfy the interface.

### DIFF
--- a/cartographer/mapping/internal/optimization/optimization_problem_2d.cc
+++ b/cartographer/mapping/internal/optimization/optimization_problem_2d.cc
@@ -208,7 +208,7 @@ void OptimizationProblem2D::InsertTrajectoryNode(const NodeId& node_id,
 }
 
 void OptimizationProblem2D::TrimTrajectoryNode(const NodeId& node_id) {
-  imu_data_.Trim(node_data_, node_id);
+  empty_imu_data_.Trim(node_data_, node_id);
   odometry_data_.Trim(node_data_, node_id);
   fixed_frame_pose_data_.Trim(node_data_, node_id);
   node_data_.Trim(node_id);
--- a/cartographer/mapping/internal/optimization/optimization_problem_2d.cc
+++ b/cartographer/mapping/internal/optimization/optimization_problem_2d.cc
@@ -175,7 +175,9 @@ OptimizationProblem2D::~OptimizationProblem2D() {}
 
 void OptimizationProblem2D::AddImuData(const int trajectory_id,
                                        const sensor::ImuData& imu_data) {
-  imu_data_.Append(trajectory_id, imu_data);
+  // TODO(MichaelGrupp): IMU data is currently not used in 2D optimization
+  // and just accumulates memory, so we ignore it for now.
+  // imu_data_.Append(trajectory_id, imu_data);
 }
 
 void OptimizationProblem2D::AddOdometryData(
--- a/cartographer/mapping/internal/optimization/optimization_problem_2d.cc
+++ b/cartographer/mapping/internal/optimization/optimization_problem_2d.cc
@@ -175,9 +175,8 @@ OptimizationProblem2D::~OptimizationProblem2D() {}
 
 void OptimizationProblem2D::AddImuData(const int trajectory_id,
                                        const sensor::ImuData& imu_data) {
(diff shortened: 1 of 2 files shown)


cartographer/mapping/internal/optimization/optimization_problem_2d.cc
```diff
@@ -175,7 +175,9 @@ OptimizationProblem2D::~OptimizationProblem2D() {}
 
 void OptimizationProblem2D::AddImuData(const int trajectory_id,
                                        const sensor::ImuData& imu_data) {
-  imu_data_.Append(trajectory_id, imu_data);
+  // TODO(MichaelGrupp): IMU data is currently not used in 2D optimization
```
wohe, line 178: A `TODO` should be actionable. How about just leaving a comment that this satisfies the interface but we do not want to keep IMU data?

We should then also remove `imu_data_` if possible.

cartographer/mapping/internal/optimization/optimization_problem_2d.cc
```diff
@@ -175,7 +175,9 @@ OptimizationProblem2D::~OptimizationProblem2D() {}
...
+  // and just accumulates memory, so we ignore it for now.
+  // imu_data_.Append(trajectory_id, imu_data);
```
wohe, line 180: Remove? No need to keep it. This is why we have version control.

MichaelGrupp:
Regarding removal of `imu_data_`: we need to have the getter [`imu_data()`](https://github.com/cartographer-project/cartographer/blob/9675d63926282f505f2e8333dd919429e49862c9/cartographer/mapping/internal/optimization/optimization_problem_interface.h#L80). I would keep it as a member, even though it will be empty now. Otherwise we need to return a new, empty object there to make it compile.

wohe:
@MichaelGrupp Good point. Rename to `empty_imu_data_` for readability?

cartographer/mapping/internal/optimization/optimization_problem_2d.cc
```diff
@@ -207,7 +208,7 @@ void OptimizationProblem2D::InsertTrajectoryNode(const NodeId& node_id,
 }
 
 void OptimizationProblem2D::TrimTrajectoryNode(const NodeId& node_id) {
-  imu_data_.Trim(node_data_, node_id);
+  empty_imu_data_.Trim(node_data_, node_id);
```
MichaelGrupp, line 211: We can keep this, it does nothing if it's empty. It's probably also good in case we ever use IMU again, so we don't forget about this part.

cartographer/mapping/internal/optimization/optimization_problem_2d.cc
```diff
@@ -175,7 +175,8 @@ OptimizationProblem2D::~OptimizationProblem2D() {}
 
 void OptimizationProblem2D::AddImuData(const int trajectory_id,
                                        const sensor::ImuData& imu_data) {
-  imu_data_.Append(trajectory_id, imu_data);
+  // IMU data is currently not used in 2D optimization, so we ignore this part
```
wohe, line 178: nit: better remove words like "currently" and "for now" as it is implied: predicting the future is futile unless you already have a plan to add it back in which case this plan should be described in a TODO. Also more concise:
```
// IMU data is not used in 2D optimization.
```

cartographer/mapping/internal/optimization/optimization_problem_2d.h
```diff
@@ -128,7 +128,8 @@ class OptimizationProblem2D
   MapById<NodeId, NodeSpec2D> node_data_;
   MapById<SubmapId, SubmapSpec2D> submap_data_;
   std::map<std::string, transform::Rigid3d> landmark_data_;
-  sensor::MapByTime<sensor::ImuData> imu_data_;
+  // IMU data is not used in 2D optimization currently and will be empty.
```
wohe, line 131: Please remove. This is the whole point of naming it `empty_imu_data_`. See https://google.github.io/styleguide/cppguide.html#Implementation_Comment_Donts

wohe (approved):

PR #1755, opened 2020-10-06
Use adaptive IMU weights in the PGO.

wohe:
This weights IMU based on the time between nodes in the
pose graph optimization.

When moving slowly or stopping, IMU weights are reduced.
This improves quality in these cases. The parameters are
changed to approximately get the same behavior while
moving as before for the examples.


configuration_files/pose_graph.lua
```diff
@@ -63,8 +63,8 @@ POSE_GRAPH = {
   matcher_rotation_weight = 1.6e3,
   optimization_problem = {
     huber_scale = 1e1,
-    acceleration_weight = 1e3,
-    rotation_weight = 3e5,
+    acceleration_weight = 1.1e2,
+    rotation_weight = 1.6e4,
```
MichaelGrupp, line 67: These params are also in the Sphinx docs. I think you should document the new behavior there (https://github.com/cartographer-project/cartographer/blob/master/docs/source/configuration.rst).

cartographer/mapping/internal/optimization/optimization_problem_3d.cc
```diff
@@ -421,8 +421,9 @@ void OptimizationProblem3D::Solve(
               result_center_to_center.delta_velocity;
           problem.AddResidualBlock(
               AccelerationCostFunction3D::CreateAutoDiffCostFunction(
-                  options_.acceleration_weight(), delta_velocity,
-                  common::ToSeconds(first_duration),
+                  options_.acceleration_weight() /
+                      common::ToSeconds(first_duration + second_duration),
```
MichaelGrupp, line 425: I assume it's guaranteed that this can't be zero by some CHECKs before we reach this?

wohe: I don't know, but this is not a new assumption that nodes do not have the same timestamp, i.e. it would already not work. If you are concerned that we do not CHECK early enough on bad data, I think that should be a separate PR.

configuration_files/pose_graph.lua
wohe: What about the documentation do you think is no longer accurate?

MichaelGrupp: > Scaling parameter for the IMU [...] term.

This is not wrong, but there could be a few words that there is an additional time-dependent factor affecting the final weight of the term.

wohe: I disagree. If we wanted to describe the implementation sure, but briefly mentioning the time dependency in the parameter description would be misleading for the following reason:
This time dependency in the implementation is necessary since we do not have equal time steps because of the motion filter which made it hard/impossible to choose a good value. This change has the additional benefit of decoupling time step size from the choice of this parameter even in the constant time step case, i.e. now users should have to think less about the interaction between these weights and the interval between consecutive nodes.

MichaelGrupp (approved):

PR #1756, opened 2020-10-07
Constrain gravity to be positive.

wohe:
This makes sure gravity never flips the sign and becomes negative.

### DIFF
--- a/cartographer/mapping/internal/optimization/optimization_problem_3d.cc
+++ b/cartographer/mapping/internal/optimization/optimization_problem_3d.cc
@@ -441,6 +441,9 @@ void OptimizationProblem3D::Solve(
             C_nodes.at(second_node_id).rotation(),
             trajectory_data.imu_calibration.data());
       }
+
+      // Force gravity constant to be positive.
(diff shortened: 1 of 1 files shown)


MichaelGrupp (approved):

soorajanilkumar:
Hi @wohe ,
This commit introduces a bug when running cartographer with my config. 

When calling the `/finish_trajectory` while running 3d localization online, cartographer crashes with the error as shown in the image below.
finish_traj_crash

I am not sure, but I think it might be due to my config where I set a high value for motion filter max time.
> TRAJECTORY_BUILDER_3D.motion_filter.max_time_seconds = 1e9

I set this to overcome the fact that the IMU drifts a lot when still, and I don't want the gravity_constant to be effected by the drift error. (If left alone, the submap/robot pose rotates bit by bit along y axis due to the drift. Corrects when some movement is introduced).

Any thoughts on the matter or a possible solution, would be much appreciated until I get a better IMU :sweat_smile: 

Atleast, will it be OK if I revert this commit in my local branch (temporarily) ?

Thank you in advance
Sooraj

PR #1757, opened 2020-10-07
Optional motion-filter for adding odometry data to the pose graph.

MichaelGrupp:
This introduces an option to control the amount of odometry data that
enters the pose graph by motion-filtering it. This is very useful to
bound memory when standing still for longer periods of time.

If the new option is not configured, all odometry data goes unfiltered
into the pose graph as usual.


wohe:
@MichaelGrupp Can you explain a bit about the context when you see large memory consumption from odometry data? Is this for pure localization or SLAM? It would seem that odometry data is very small compared to the other data we keep.

MichaelGrupp:
The motivation for this was localization. Concrete example: we run a robot with pure localization mode that can be "idle" for some time, e.g. during a charging phase, but we want to keep Cartographer running. In this case, our motion filter is configured to not add new nodes to the pose graph unless we move again, which works fine. But we noticed that we were still accumulating odometry data that has no additional value without movement. Filtering odometry data that is close to zero prevents that and I didn't observe any side effects on the optimization since the values with information (when moving) still enter the pose graph.

bochen87:
> The motivation for this was localization. Concrete example: we run a robot with pure localization mode that can be "idle" for some time, e.g. during a charging phase, but we want to keep Cartographer running. In this case, our motion filter is configured to not add new nodes to the pose graph unless we move again, which works fine. But we noticed that we were still accumulating odometry data that has no additional value without movement. Filtering odometry data that is close to zero prevents that and I didn't observe any side effects on the optimization since the values with information (when moving) still enter the pose graph.

It makes sense. I think most robots have certain idle time where you would still want to keep localization alive.

MichaelGrupp:
@wohe can you have a look again if you have time?

wohe (approved):
We can definitely merge this as a stop gap solution for now.

That said, I think we should add a `TODO` for an improved solution. Generally I think we should be careful only to add options when they provide a significant benefit. Otherwise we have the disadvantage that to benefit from the improvements you need to know about, understand and tune your options.

Here I think we could do without since odometry data is only kept in the pose graph and we really only need the interpolated result per trajectory node. I filed #1768 for this which you could refer to in a `TODO`. What do you think?

cartographer/mapping/internal/global_trajectory_builder.cc
```diff
@@ -97,6 +101,14 @@ class GlobalTrajectoryBuilder : public mapping::TrajectoryBuilderInterface {
     if (local_trajectory_builder_) {
       local_trajectory_builder_->AddOdometryData(odometry_data);
     }
+    // TODO(MichaelGrupp): Instead of having an optional filter on this level,
+    // odometry could be marginalized between nodes in the pose graph.
+    // Related issue: cartographer-project/cartographer/#1768
```
MichaelGrupp, line 106: Added a note here. I agree that this could be implicitly handled by making the pose graph smarter.

wohe (approved):
Could you merge master?

MichaelGrupp:
Done.

PR #1760, opened 2020-10-13
Add a cost function to take intensities into account.

wohe:
Adds (optional) intensities to sensor::PointCloud and
uses it in a cost function. Not yet in use.

### DIFF
--- a/cartographer/mapping/internal/3d/scan_matching/intensity_cost_function_3d.cc
+++ b/cartographer/mapping/internal/3d/scan_matching/intensity_cost_function_3d.cc
@@ -1,0 +1,25 @@
+#include "cartographer/mapping/internal/3d/scan_matching/intensity_cost_function_3d.h"
+
+namespace cartographer {
+namespace mapping {
+namespace scan_matching {
+
+// This method is defined here instead of the header file as it was observed
+// that defining it in the header file has a negative impact on the runtime
+// performance.
+ceres::CostFunction* IntensityCostFunction3D::CreateAutoDiffCostFunction(
+    const double scaling_factor, const float intensity_threshold,
+    const sensor::PointCloud& point_cloud,
+    const IntensityHybridGrid& hybrid_grid) {
+  CHECK(!point_cloud.intensities().empty());
+  return new ceres::AutoDiffCostFunction<
+      IntensityCostFunction3D, ceres::DYNAMIC /* residuals */,
+      3 /* translation variables */, 4 /* rotation variables */>(
+      new IntensityCostFunction3D(scaling_factor, intensity_threshold,
+                                  point_cloud, hybrid_grid),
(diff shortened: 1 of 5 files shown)


cartographer/mapping/internal/3d/scan_matching/intensity_cost_function_3d.h
```diff
@@ -0,0 +1,93 @@
...
+#include "cartographer/sensor/point_cloud.h"
+#include "cartographer/transform/rigid_transform.h"
+#include "cartographer/transform/transform.h"
+#include "ceres/ceres.h"
+
+namespace cartographer {
+namespace mapping {
+namespace scan_matching {
+
+class IntensityCostFunction3D {
```
MichaelGrupp, line 32: Do you want to add a small summary as we have in the other cost functions? E.g. https://github.com/cartographer-project/cartographer/blob/ee98a92845f7ded3c7e509d32f245f3860d23001/cartographer/mapping/internal/3d/scan_matching/occupied_space_cost_function_3d.h#L31

cartographer/mapping/internal/3d/scan_matching/intensity_cost_function_3d.cc
```diff
@@ -0,0 +1,25 @@
+#include "cartographer/mapping/internal/3d/scan_matching/intensity_cost_function_3d.h"
+
+namespace cartographer {
+namespace mapping {
+namespace scan_matching {
+
+// This method is defined here instead of the header file as it was observed
+// that defining it in the header file has a negative impact on the runtime
+// performance.
```
MichaelGrupp, line 9: Interesting, do you know if this is something that only applies to this function or if it also affects the other cost functions?

cartographer/sensor/point_cloud.h
```diff
@@ -55,6 +57,9 @@ struct PointCloud {
  private:
   // For 2D points, the third entry is 0.f.
   std::vector<PointType> points_;
+  // Intensities are optional. If not empty they must have the same size as
```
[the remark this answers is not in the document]
wohe: Done.

cartographer/mapping/internal/3d/scan_matching/intensity_cost_function_3d.h
wohe: Done.

cartographer/mapping/internal/3d/scan_matching/intensity_cost_function_3d.cc
wohe: No idea. But a lot of attention was paid to make sure that this does not have a measurable effect on running time performance when intensities are not used.

MichaelGrupp (approved):

PR #1761, opened 2020-10-14
Add intensity cost function to the Ceres scan matcher.

wohe:
Not yet used.


MichaelGrupp (approved):
LGTM. One small nit-picking regarding the commit message:

```
Add intensity cost function to the Ceres scan matcher.

Not yet used.
```

makes more sense when reading the Git log than `Use intensities [...] ... Not yet used.` :smile:

PR #1762, opened 2020-10-14
Intensities from Lidar topic into the generated PCD

ikhann:
✸ Add intensity field from the Lidar topic into the generated PCD file.

If you have topic messages with intensities from the lidar measurements, it will be added when you generate a .pcd file. 
For some reason, this functionality was implemented only when generating a .ply file.


MichaelGrupp:
Hi, thanks for the pull request. You need to `sign-off` your commit with Git, otherwise we can't merge.

It takes 1 minute. The DCO check shows you what you need to do: https://github.com/cartographer-project/cartographer/pull/1762/checks?check_run_id=1253819084

cartographer/io/pcd_writing_points_processor.cc
```diff
@@ -32,20 +32,25 @@ namespace {
...
-void WriteBinaryPcdHeader(const bool has_color, const int64 num_points,
-                          FileWriter* const file_writer) {
+void WriteBinaryPcdHeader(const bool has_color, const bool has_intensities,
+                          const int64 num_points, FileWriter* const file_writer) {
   std::string color_header_field = !has_color ? "" : " rgb";
   std::string color_header_type = !has_color ? "" : " U";
   std::string color_header_size = !has_color ? "" : " 4";
   std::string color_header_count = !has_color ? "" : " 1";
 
+  std::string intensity_header_field = !has_intensities ? "" : " intensity";
```
wohe, line 42: Is this style of an intensity field commonly used in PCD files? Is it documented somewhere? https://pointclouds.org/documentation/tutorials/pcd_file_format.html does not mention it.

cartographer/io/pcd_writing_points_processor.cc
```diff
@@ -32,20 +32,25 @@ namespace {
...
+  std::string intensity_header_count = !has_intensities ? "" : " 1";
+
   std::ostringstream stream;
   stream << "# generated by Cartographer\n"
          << "VERSION .7\n"
-         << "FIELDS x y z" << color_header_field << "\n"
-         << "SIZE 4 4 4" << color_header_size << "\n"
-         << "TYPE F F F" << color_header_type << "\n"
-         << "COUNT 1 1 1" << color_header_count << "\n"
+         << "FIELDS x y z" << intensity_header_field << color_header_field <<"\n"
```
wohe, line 50: Any reason why the ordering between color and intensity is swapped here compared to above for arguments? Some consistency would be nice.

cartographer/io/pcd_writing_points_processor.cc
```diff
@@ -32,20 +32,25 @@ namespace {
...
+void WriteBinaryPcdHeader(const bool has_color, const bool has_intensities,
+                          const int64 num_points, FileWriter* const file_writer) {
   std::string color_header_field = !has_color ? "" : " rgb";
   std::string color_header_type = !has_color ? "" : " U";
   std::string color_header_size = !has_color ? "" : " 4";
   std::string color_header_count = !has_color ? "" : " 1";
 
+  std::string intensity_header_field = !has_intensities ? "" : " intensity";
+  std::string intensity_header_type = !has_intensities ? "" : " F";
+  std::string intensity_header_size = !has_intensities ? "" : " 4";
```
wohe, line 44: nit: these could all be made `const std::string...`. Same above for the color header strings.

cartographer/io/pcd_writing_points_processor.cc
```diff
@@ -65,6 +70,12 @@ void WriteBinaryPcdPointCoordinate(const Eigen::Vector3f& point,
   CHECK(file_writer->Write(buffer, 12));
 }
 
+void WriteBinaryPcdIntensity(const float intensity,
+                             FileWriter* const file_writer) {
+  CHECK(file_writer->Write(reinterpret_cast<const char*>(&intensity),
```
wohe, line 75: This is undefined behavior. See https://en.cppreference.com/w/cpp/language/reinterpret_cast. Use `memcpy` as above?
nit: You probably want to `#include <cstring>` for this and use `std::memcpy`. Similar for the existing code.

cartographer/io/pcd_writing_points_processor.cc
ikhann: Yes, not all possible fields are listed on the site. I usually define points of type pcl::PointXYZI, and these fields are also available in pcd.

For example, you can find all supported fields for the point cloud library file format here https://libpointmatcher.readthedocs.io/en/latest/ImportExport/

cartographer/io/pcd_writing_points_processor.cc
ikhann: The order is not important.

cartographer/io/pcd_writing_points_processor.cc
ikhann: Yeah, you are right.

cartographer/io/pcd_writing_points_processor.cc
ikhann: Thanks. These comments were implemented in the commit below.

cartographer/io/pcd_writing_points_processor.cc
wohe: Thanks.

cartographer/io/pcd_writing_points_processor.cc
wohe: Did you see my nit? It is good practice to include what you use (IWYU) which here means: `#include <cstring>`. Also, in C++ `memcpy` is in the `std` namespace.

cartographer/io/pcd_writing_points_processor.cc
wohe: You mean not important for correctness, but this is setting the bar too low. Please have a look at https://google.github.io/styleguide/cppguide.html#Goals. Inconsistencies should be there for a reason, not consistency. If you aim for readable code everyone who later reads the code, including yourself, has an easier time.

wohe (approved):

cartographer/io/pcd_writing_points_processor.cc
```diff
@@ -65,6 +70,13 @@ void WriteBinaryPcdPointCoordinate(const Eigen::Vector3f& point,
   CHECK(file_writer->Write(buffer, 12));
 }
 
+void WriteBinaryPcdIntensity(const float intensity,
+                                 FileWriter* const file_writer) {
```
wohe, line 74: nit: you should run clang-format.

cartographer/io/pcd_writing_points_processor.cc
```diff
@@ -125,6 +139,10 @@ void PcdWritingPointsProcessor::Process(std::unique_ptr<PointsBatch> batch) {
       WriteBinaryPcdPointColor(ToUint8Color(batch->colors[i]),
                                file_writer_.get());
     }
+    if (has_intensities_) {
```
wohe, line 142: Isn't this wrong? Which would be another argument for consistency. In case of both colors and intensities you declare that intensities come first, but here you right them in the wrong order. Or do I misread this?

wohe (changes requested):

bochen87:
@ikhann friendly ping

PR #1763, opened 2020-10-15
Take intensities into account in PointCloud transforms and filters.

wohe:
Adds a new method to sensor::PointCloud, copy_if, which copies all
points satisfying a provided condition, together with the associated
intensities (if they exist).

Adapts transforms/filters to use this new method.

### DIFF
--- a/cartographer/sensor/point_cloud.h
+++ b/cartographer/sensor/point_cloud.h
@@ -54,6 +54,35 @@ class PointCloud {
 
   void push_back(PointType value);
 
+  // Creates a PointCloud consisting of all the points for which `predicate`
+  // returns true, together with the corresponding intensities.
+  template <class UnaryPredicate>
+  PointCloud copy_if(UnaryPredicate predicate) const {
+    std::vector<PointType> points;
+    std::vector<float> intensities;
+
+    // Note: benchmarks show that it is better to have this conditional outside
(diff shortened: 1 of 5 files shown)


cartographer/sensor/point_cloud.h
```diff
@@ -54,6 +54,35 @@ class PointCloud {
...
 
+  // Creates a PointCloud consisting of all the points for which `predicate`
+  // returns true, together with the corresponding intensities.
+  template <class UnaryPredicate>
+  PointCloud copy_if(UnaryPredicate predicate) const {
+    std::vector<PointType> points;
+    std::vector<float> intensities;
+
+    // Note: benchmarks show that it is better to have this conditional outside
+    // the loop.
```
MichaelGrupp, line 65: Makes sense. Out of curiosity I was thinking if sth like this would also work, but with your explicit two loops we are on the safe side.

```cpp
      const bool with_intensities = !intensities_.empty();
      for (size_t index = 0; index < size(); ++index) {
        const PointType& point = points_[index];
        if (predicate(point)) {
          points.push_back(point);
          if (with_intensities) {
            intensities.push_back(intensities_[index]);
          }
        }
      }
```

MichaelGrupp (approved):

PR #1764, opened 2020-10-16
Keep intensities in the voxel filter.

wohe:

### DIFF
--- a/cartographer/sensor/internal/voxel_filter_test.cc
+++ b/cartographer/sensor/internal/voxel_filter_test.cc
@@ -25,6 +25,7 @@ namespace sensor {
 namespace {
 
 using ::testing::Contains;
+using ::testing::IsEmpty;
 
 TEST(VoxelFilterTest, ReturnsOnePointInEachVoxel) {
   const PointCloud point_cloud({{{0.f, 0.f, 0.f}},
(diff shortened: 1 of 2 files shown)


cartographer/sensor/internal/voxel_filter_test.cc
```diff
@@ -33,18 +34,37 @@ TEST(VoxelFilterTest, ReturnsOnePointInEachVoxel) {
...
+    const float value = 0.1f * i;
+    points.push_back({{-100.f, 0.3f, value}});
+    intensities.push_back(value);
+  }
+  const PointCloud point_cloud(points, intensities);
+  const PointCloud result = VoxelFilter(point_cloud, 0.3f);
+
+  ASSERT_EQ(result.intensities().size(), result.points().size());
+  for (size_t i = 0; i < result.size(); ++i) {
+    ASSERT_NEAR(result[i].position.z(), result.intensities()[i], 1e-6);
```
MichaelGrupp, line 56: This confused me in the beginning.
So you applied a trick by setting the Z and intensity values for each point to an unique number for each entry of the point cloud, such that you can assert here that the pairs are still correct after the voxel filter - am I right?
Could be worth to mention this in a small comment.

wohe: Added a comment. PTAL.

MichaelGrupp (approved):

cartographer/sensor/internal/voxel_filter_test.cc
MichaelGrupp: Thank you

PR #1765, opened 2020-10-19
Use intensities in `Submap3D` and `RangeDataInserter3D`.

wohe:
`Submap3D` now also stores a pointer to `IntensityHybridGrid`.
Adapted `RangeDataInserter3D` to also insert intensities.

### DIFF
--- a/cartographer/mapping/3d/submap_3d.cc
+++ b/cartographer/mapping/3d/submap_3d.cc
@@ -336,10 +336,10 @@ void ActiveSubmaps3D::AddSubmap(
     // This will crop the finished Submap before inserting a new Submap to
     // reduce peak memory usage a bit.
     CHECK(submaps_.front()->insertion_finished());
-    // We do `ForgetIntensityHybridGrid` in order to reduce memory usage.
-    // As we use active submaps, in particular intensity hybrid grids associated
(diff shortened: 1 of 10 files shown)


cartographer/mapping/3d/submap_3d.cc
```diff
@@ -332,6 +336,11 @@ void ActiveSubmaps3D::AddSubmap(
     // This will crop the finished Submap before inserting a new Submap to
     // reduce peak memory usage a bit.
     CHECK(submaps_.front()->insertion_finished());
+    // We do `ForgetIntensityHybridGrid` in order to reduce memory usage.
+    // As we use active submaps, in particular intensity hybrid grids associated
+    // with them, for scan matching, thus we call `ForgetIntensityHybridGrid`
+    // only when we are about to remove the submap from active submaps.
```
MichaelGrupp, line 342: This sentence is a bit bumpy to read. How about this:

> As we use active submaps for scan matching, in particular intensity hybrid grids associated with them, we call `ForgetIntensityHybridGrid` only when we are about to remove the submap from active submaps.

wohe: Rephrased. What do you think?

MichaelGrupp (approved):

PR #1767, opened 2020-10-28
Clean trimming of frozen trajectory data.

MichaelGrupp:
The deletion logic needs to take care of deleting all data that is
"exclusively" connected to the submaps that are to be removed. This is
achieved by looking at the data that is connected via constraints in
the graph.

Deleting a frozen trajectory (one without optimization constraints)
doesn't work that way and would leave dangling nodes in the graph.

This modifies the logic to use the `node_ids` field of the submap
instead of the constraints.

### DIFF
--- a/cartographer/mapping/internal/2d/pose_graph_2d.cc
+++ b/cartographer/mapping/internal/2d/pose_graph_2d.cc
@@ -1210,28 +1210,24 @@ void PoseGraph2D::TrimmingHandle::TrimSubmap(const SubmapId& submap_id) {
   CHECK(parent_->data_.submap_data.at(submap_id).state ==
         SubmapState::kFinished);
 
-  std::set<NodeId> nodes_to_remove;
-  // We need to use node_ids instead of constraints here to be also compatible
-  // with frozen trajectories that don't have intra-constraints.
-  nodes_to_remove.insert(
-      parent_->data_.submap_data.at(submap_id).node_ids.begin(),
-      parent_->data_.submap_data.at(submap_id).node_ids.end());
-
   // Compile all nodes that are still INTRA_SUBMAP constrained to other submaps
   // once the submap with 'submap_id' is gone.
+  // We need to use node_ids instead of constraints here to be also compatible
+  // with frozen trajectories that don't have intra-constraints.
   std::set<NodeId> nodes_to_retain;
-  for (const auto& other_submap : parent_->data_.submap_data) {
-    if (other_submap.id == submap_id) {
-      continue;
+  for (const auto& submap_data : parent_->data_.submap_data) {
+    if (submap_data.id != submap_id) {
+      nodes_to_retain.insert(submap_data.data.node_ids.begin(),
+                             submap_data.data.node_ids.end());
     }
-    std::set_intersection(
-        nodes_to_remove.begin(), nodes_to_remove.end(),
-        other_submap.data.node_ids.begin(), other_submap.data.node_ids.end(),
-        std::inserter(nodes_to_retain, nodes_to_retain.begin()));
   }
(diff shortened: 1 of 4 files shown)


cartographer/mapping/internal/3d/pose_graph_3d_test.cc
```diff
@@ -323,6 +323,59 @@ TEST_F(PoseGraph3DTest, EvenSubmapTrimmer) {
   }
 }
 
+TEST_F(PoseGraph3DTest, EvenSubmapTrimmerOnFrozenTrajectory) {
```
MichaelGrupp, line 326: Note: this test-case is based on the existing `EvenSubmapTrimmer` test case above, with the difference that no constraints are added and the graph is built like we do when loading frozen state.

There's no equivalent test case in `pose_graph_2d_test.cc` yet, so I did it only for 3D. If we want to extend the 2D test cases, this would be a larger change for separate PRs IMO.

MichaelGrupp:
Related to the linked RFC PR and the associated PR in cartographer_ros.

cartographer/mapping/internal/2d/pose_graph_2d.cc
```diff
@@ -1237,6 +1237,26 @@ void PoseGraph2D::TrimmingHandle::TrimSubmap(const SubmapId& submap_id) {
     }
     parent_->data_.constraints = std::move(constraints);
   }
+
+  // A frozen trajectory needs special care because there we can't identify the
+  // nodes_to_remove based on optimization constraints (it doesn't have them).
+  if (parent_->IsTrajectoryFrozen(submap_id.trajectory_id)) {
+    // Remove node_ids that are associated with the frozen submap to be trimmed.
+    nodes_to_remove.insert(
+        parent_->data_.submap_data.at(submap_id).node_ids.begin(),
```
wohe, line 1246: Do these `node_ids` always exist? Would it then make sense to get rid of the old style of handling deletion of nodes and replace it by your improved strategy completely? Something along these lines:
1. Compute `nodes_to_retain` by going through all submaps and adding all `node_ids` for submaps not equal to the given `submap_id` to remove.
2. Compute `nodes_to_remove` as all `node_ids` of `submap_id` not to be retained.
3. Remove the constraints similar to before, but without the need to build up the `nodes_to_remove` as part of it. We could add a `CHECK` to validate that all retained constraints use nodes not in `nodes_to_remove`.

What do you think?

MichaelGrupp: Regarding your 3. point: There's already a loop that only keeps constraints for which `nodes_to_remove.count(constraint.node_id) == 0` holds. With this in place we don't need to check afterwards, no?

wohe (approved):
Looks good. Only one nit: Could we update the comment in `map_builder.cc`? Now setting the nodes contained in a submap is not only needed for 3D, but also 2D: https://github.com/cartographer-project/cartographer/blob/master/cartographer/mapping/map_builder.cc#L362

Maybe simply change it to:
`// Information about which nodes belong to which submap is required even without constraints.`

cuicuihaohaoxuexi:
Hi there, I've tried this DeleteTrajectory service to try to remove a loaded frozen trajectory. It works well and the frozen trajectory is deleted. However, the code crashes after the frozen trajectory is deleted. The error is shown as below:

terminate called after throwing an instance of 'std::out_of_range'
  what():  map::at

I find it happens in constraint_builder_2d.cc. The line of "(*callback)(result);" I used the latest master for cartographer which has all the updates above. Does anyone have any idea why it happens and how to resolve it? Thanks very much in advance.

shamengjun:
@cuicuihaohaoxuexi I'm also confused with this problem, my error is the same as you!

PR #1773, opened 2020-11-03
Add missing virtual destructors, fix #1771

### LINKED ISSUE #1771: conda-forge: Test failures on OS X
Hi cartographer team!

We are packaging cartographer on conda-forge and we have a test failure on OS X.

We already have working builds for Windows and Linux :tada:

Do you have any idea where this ILLEGAL error could come from? Is it an illegal instruction?

https://github.com/conda-forge/cartographer-feedstock/pull/7
https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=235112&view=results

Tobias-Fischer:
Signed-off-by: Tobias Fischer <info@tobiasfischer.info>

See #1771

Successfully fixed the test failures on osx on conda-forge & compiler does not complain anymore (previous warning example: `warning: delete called on 'cartographer::mapping::Grid2D' that is abstract but has non-virtual destructor [-Wdelete-abstract-non-virtual-dtor]`.

### DIFF
--- a/cartographer/mapping/2d/grid_2d.h
+++ b/cartographer/mapping/2d/grid_2d.h
@@ -42,6 +42,7 @@ class Grid2D : public GridInterface {
          ValueConversionTables* conversion_tables);
   explicit Grid2D(const proto::Grid2D& proto,
                   ValueConversionTables* conversion_tables);
+  virtual ~Grid2D() {}
 
   // Returns the limits of this Grid2D.
   const MapLimits& limits() const { return limits_; }
--- a/cartographer/mapping/2d/probability_grid.h
+++ b/cartographer/mapping/2d/probability_grid.h
@@ -34,6 +34,7 @@ class ProbabilityGrid : public Grid2D {
                            ValueConversionTables* conversion_tables);
   explicit ProbabilityGrid(const proto::Grid2D& proto,
                            ValueConversionTables* conversion_tables);
+  virtual ~ProbabilityGrid() {}
 
   // Sets the probability of the cell at 'cell_index' to the given
   // 'probability'. Only allowed if the cell was unknown before.
--- a/cartographer/mapping/2d/probability_grid_range_data_inserter_2d.h
+++ b/cartographer/mapping/2d/probability_grid_range_data_inserter_2d.h
@@ -40,6 +40,7 @@ class ProbabilityGridRangeDataInserter2D : public RangeDataInserterInterface {
  public:
   explicit ProbabilityGridRangeDataInserter2D(
       const proto::ProbabilityGridRangeDataInserterOptions2D& options);
+  virtual ~ProbabilityGridRangeDataInserter2D() {}
 
   ProbabilityGridRangeDataInserter2D(
       const ProbabilityGridRangeDataInserter2D&) = delete;
--- a/cartographer/mapping/range_data_inserter_interface.h
+++ b/cartographer/mapping/range_data_inserter_interface.h
@@ -37,6 +37,8 @@ class RangeDataInserterInterface {
   // Inserts 'range_data' into 'grid'.
   virtual void Insert(const sensor::RangeData& range_data,
                       GridInterface* grid) const = 0;
+
+  virtual ~RangeDataInserterInterface() {}
 };
 
 }  // namespace mapping


cartographer/mapping/2d/probability_grid_range_data_inserter_2d.h
```diff
@@ -40,6 +40,7 @@ class ProbabilityGridRangeDataInserter2D : public RangeDataInserterInterface {
  public:
   explicit ProbabilityGridRangeDataInserter2D(
       const proto::ProbabilityGridRangeDataInserterOptions2D& options);
+  virtual ~ProbabilityGridRangeDataInserter2D() {}
```
johuber, line 43: same as above

cartographer/mapping/2d/grid_2d.h
```diff
@@ -42,6 +42,7 @@ class Grid2D : public GridInterface {
          ValueConversionTables* conversion_tables);
   explicit Grid2D(const proto::Grid2D& proto,
                   ValueConversionTables* conversion_tables);
+  virtual ~Grid2D() {}
```
johuber, line 45: this is redundant, since the parent destructor is virtual already. If you want to make it explicit here you may instead use:
`~Grid2D() override = default` (I always like this since it kind of provides a sanity check that the parents destructor is virtual)

cartographer/mapping/2d/probability_grid.h
```diff
@@ -34,6 +34,7 @@ class ProbabilityGrid : public Grid2D {
                            ValueConversionTables* conversion_tables);
   explicit ProbabilityGrid(const proto::Grid2D& proto,
                            ValueConversionTables* conversion_tables);
+  virtual ~ProbabilityGrid() {}
```
johuber, line 37: same as above

cartographer/mapping/range_data_inserter_interface.h
```diff
@@ -37,6 +37,8 @@ class RangeDataInserterInterface {
   // Inserts 'range_data' into 'grid'.
   virtual void Insert(const sensor::RangeData& range_data,
                       GridInterface* grid) const = 0;
+
+  virtual ~RangeDataInserterInterface() {}
```
johuber, line 41: this is the only important one, good catch! I personally like syntax:
`virtual ~RangeDataInserterInterface() = default` better though

johuber (changes requested):
good catch, thank you! please see my my comments and note that inherited virtual functions are always virtual again, hence this does not need to be re-declared. In this case use `final` or `override` depending on the situation

cartographer/mapping/range_data_inserter_interface.h
johuber: actually just realized you are re-declaring this, so should remove!

johuber:
I'm curious to see if with the requested changes this compiler warning will be rsolve.. because everything was already done correctly before, the only thing this changes is more explicit syntax.

cartographer/mapping/2d/grid_2d.h
wohe: +1. Please remove `virtual`, but add `override`, see https://google.github.io/styleguide/cppguide.html#Inheritance. Similar below.

wohe (changes requested):
I do not understand this PR. To me it looks like the compiler error is a false positive since both `GridInterface` and `RangeDataInserterInterface` already define a virtual destructor. Can you explain what you think is wrong with the code?

If this is indeed to mitigate a compiler issue: did you file an issue about this compiler issue?

johuber:
> I do not understand this PR. To me it looks like the compiler error is a false positive since both `GridInterface` and `RangeDataInserterInterface` already define a virtual destructor. Can you explain what you think is wrong with the code?
> 
> If this is indeed to mitigate a compiler issue: did you file an issue about this compiler issue?

I fully agree! marking them override cannot harm though and maybe suffices to get rid of the compiler warning

Tobias-Fischer:
Hi there,

Thank you for your comments!

I do not have any knowledge of cartographer - this PR originated (as described above, where I also link to the issue https://github.com/cartographer-project/cartographer/issues/1771 which contains the compiler warnings and more information) from failing test cases on OSX when packaging cartographer for conda-forge.

After applying the changes in this PR, both the compiler warnings (not "compiler error", I nowhere talk of compiler errors), as well as the test failures, disappear. So I am not sure how you think that "everything was done correctly before" and "the compiler error is a false positive". Things obviously weren't all done correctly before, nor was it a false positive, as there were test failures (albeit only on OSX with clang, but nonetheless) that are now resolved.

You are more than welcome to make the requested changes in this PR if you want to improve cartographer by fixing these test failures, edits from the maintainers are allowed. If you still think I am fixing false positives, I am equally happy to close this PR and continue to use a patch to fix the problems in conda-forge.

wolfv:
Hey @johuber @wohe 

here are the build & test logs for cartographer on osx-64, without the fixes applied: https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=235112&view=logs&jobId=58ac6aab-c4bc-5de2-2894-98e408cc8ec9&j=58ac6aab-c4bc-5de2-2894-98e408cc8ec9&t=933f325c-924e-533d-4d95-e93b5843ce8b  

With the fixes applied all builds are green: https://github.com/conda-forge/cartographer-feedstock/pull/7

We now provide up-to-date cartographer builds on conda-forge for Windows, OS X and Linux :)

Btw, if either of you is interested in helping to maintain cartographer on conda-forge, please let me know and we can add you as maintainer. This usually doesn't involve a lot of effort, just now and then updating the version if a new release appears (there are bots for this, too).

wolfv:
Hm, I think the issue is that the 1.0.0 which is quite old by now didn't have the virtual destructor in the base class. E.g. this PR https://github.com/cartographer-project/cartographer/pull/1453 hasn't made it in the 1.0.0 release as far as I can see. How about a new release? Looks like there is quite a bit of development since 1.0 :)

So I guess we could also fix our build in conda-forge by adding the empty virtual destructor in the GridInterface base class.
Either way, what we have works for the moment.

Tobias-Fischer:
Thanks for investigating this @wolfv! Seems like this issue does not occur on the master branch anymore - apologies for the confusion caused. Closing here

PR #1776, opened 2020-11-05
Clean up use of MapBuilderInterface.

wohe:
Moves options next to the interface like we do for other interfaces.
Adds a factory function to remove the need for direct use of MapBuilder.


MichaelGrupp (approved):
Looks fine to me. Do you also want to do the follow-up in cartographer_ros? E.g. there's currently a factory function defined in the offline node code, which should be removed then.

wohe:
@MichaelGrupp Yes, this is my intention.

PR #1781, opened 2020-11-18
Removed unused code.

wohe:


MichaelGrupp:
Why do you think the mock classes are unused? They are included in the [handler_test.h](https://github.com/cartographer-project/cartographer/blob/5b44305ea30e6af0d1b1edb50e758881f3330c07/cartographer/cloud/internal/testing/handler_test.h), so I would expect that they are required by the different handler tests that include that header, e.g. [here](https://github.com/cartographer-project/cartographer/blob/5b44305ea30e6af0d1b1edb50e758881f3330c07/cartographer/cloud/internal/handlers/add_trajectory_handler_test.cc)

wohe:
@MichaelGrupp You are right. I did not find where they are used because the includes did not follow best practices. Reverted.

MichaelGrupp (approved):

PR #1880, opened 2022-05-04
Use grid.GetProbability rather than grid.GetCorrespondenceCost

ysklab:
I believe that the program is trying to get the probability value of some cell, so just use GetProbability() will be fine.

Signed-off-by: ysklab <ruincity@qq.com>

Want to contribute? Great! Make sure you've read and understood
[CONTRIBUTING.md](https://github.com/cartographer-project/cartographer/blob/master/CONTRIBUTING.md).


ysklab:
And even the unit test file uses GetProbability() to verify this module's function, too.

wohe:
Looking into this PR, I noticed 2 things:

1. You change a call to `GetProbability()` for a `Grid2D` which does not have this function, so your code will not compile. I assume you did not try to compile it before sending this PR? It is always a good idea to make sure the code of a PR actually works. The unit test uses a `ProbabilityGrid` which is a more specific class. As I see it, the implementation is more general and the unit test is verifying a concrete case of using probability grids.
2. Looking into the history of the code you want to change is always a good idea if you are unsure why the code has its current form. For the code in question this leads you to https://github.com/cartographer-project/rfcs/blob/master/text/0019-probability-grid-and-submap2d-restructuring.md.

Issue #1879, opened 2022-03-30
ceres::LocalParameterization has been deprecated in the Ceres Solver 2.1.0, migrate to using ceres::Manifold


sandwichmaker:
``ceres::LocalParameterization`` has been deprecated in the latest version of Ceres Solver (v 2.1.0) and will be removed going forward.  Cartographer should migrate to using the new api (maybe guarded by ceres version number).

PR #1882, opened 2022-05-12
Use Manifold instead of LocalParameterization.

zjwoody:
We need to migrate the Cartographer code to use Manifolds instead of LocalParameterization to allow Ceres code to remove the LocalParameterization code.

The PR converts internal use of LocalParameterization to Manifold.  Removes headers that reference local_parameterization or implementation classes.

Note that internal variable and method names with type Manifold were renamed Manifold/manifold, but the outer abstraction class, proto name and so on were left as Parameterization/parameterization.

Signed-off-by: Jie Zheng <zhengj@google.com>

### DIFF
--- a/cartographer/mapping/internal/optimization/optimization_problem_3d.cc
+++ b/cartographer/mapping/internal/optimization/optimization_problem_3d.cc
@@ -160,8 +160,8 @@ void AddLandmarkCostFunctions(
                                          *prev_node_pose, *next_node_pose);
         C_landmarks->emplace(
             landmark_id,
-            CeresPose(starting_point, nullptr /* translation_parametrization */,
-                      absl::make_unique<ceres::QuaternionParameterization>(),
+            CeresPose(starting_point, nullptr /* translation_manifold */,
+                      absl::make_unique<ceres::QuaternionManifold>(),
                       problem));
         // Set landmark constant if it is frozen.
         if (landmark_node.second.frozen) {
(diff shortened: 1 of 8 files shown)


wohe:
@zjwoody Sorry, it took slightly longer than expected. Now CI should work again. Can you rebase your PR? I also had a quick look at it already, and my understanding is that it will not compile as is:
1. We support Ubuntu and Debian releases until LTS ends, so for several years we will have to be able to use a Ceres version which does not yet support Manifold.
2. Thus, we would need to guard using a version number as suggested in #1879.
3. Just FYI, this should not delay this PR in any way. If I understand correctly then both Ubuntu 22.04 and Debian Bullseye do not support Manifold, so CI would not be able to test the Manifold related code. But it seems the following releases when they happen will pick it up. Probably 2024 if Ubuntu releases as usual, and whenever Debian does.

zjwoody:
Hi Wolfgang, thanks for the information!
I rebased the PR and it is not able to compile due to the reason you mentioned, so it can't be merged.
I have some stupid questions since I'm not clear about what to do next:
1. For the guard of ceres version number, do you mean limit the ceres version used in Cartographer to the old version(e.g. < 2.1.0) before it switched to the manifold? 
2. If 1 is true and we guard the ceres version, what should we do with this pr?
Thanks!

bufeng-12:
您好！我已收到你的邮件！谢谢！          ——曾君

cartographer/mapping/internal/optimization/optimization_problem_3d.cc
dennuguyen, line 364: Should the argument be `new ceres::QuaternionManifold()`?

PR #1884, opened 2022-06-03
Move to GitHub Actions for CI.

wohe:
This moves away from travis-ci.org which no longer works.


wohe:
You can see it running here: https://github.com/wohe/cartographer/actions

zjwoody:
> @bochen87 @zjwoody @MichaelGrupp I looked into moving from travis-ci.org to travis-ci.com as Travis suggests, but after doing this work, it only led to an error: "Owner cartographer-project is not on a new pricing." (see https://app.travis-ci.com/github/cartographer-project/cartographer/requests). So I decided to instead move to GitHub Actions which seems to work. Thoughts?

SGTM. Thanks for the work!
I searched about the "not on a new pricing" issue: https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing
It affects the user who build on mac or have more than 5 concurrent builds. IIRC cartographer doesn't support mac, is it because cartographer has more than 5 concurrent builds? Is the concurrent builds caused by the too many pending prs?

MichaelGrupp (approved):

PR #1888, opened 2022-06-29
Add Ubuntu 22.04 to the install scripts, CI, docs.

wohe:

### DIFF
--- a/.github/workflows/ci-jammy.yaml
+++ b/.github/workflows/ci-jammy.yaml
@@ -1,0 +1,15 @@
+name: Ubuntu 22.04 CI
+
+on:
+    push:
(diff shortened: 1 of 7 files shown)


wohe:
@MichaelGrupp @bochen87 This adds Ubuntu 22.04. Debian Bullseye is still missing.

The failure of Ubuntu 18.04 is because of gRPC which we only use there and I will fix this in a follow up PR (most likely splitting the gRPC version into a separate Dockerfile so we can test Bionic similar to the others).

Once this is fixed we can again make passing CI a precondition for merging. What do you think?

bochen87 (approved):
lgtm!

PR #1897, opened 2022-07-21
Add Debian Bullseye to the install scripts, CI, docs.

wohe:

### DIFF
--- a/Dockerfile.bullseye
+++ b/Dockerfile.bullseye
@@ -1,4 +1,4 @@
-# Copyright 2020 The Cartographer Authors
+# Copyright 2022 The Cartographer Authors
 #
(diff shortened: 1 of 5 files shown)


Dockerfile.bullseye
```diff
@@ -0,0 +1,30 @@
+# Copyright 2020 The Cartographer Authors
```
MichaelGrupp, line 1: Nit: 2022?

MichaelGrupp (approved):
LGTM besides tiny comment nit.

Dockerfile.bullseye
wohe: Fixed.

PR #1919, opened 2023-02-28
Update code to work with recent abseil changes

kkufieta:


kkufieta:
Gentle ping - I would love to get these changes merged so we can stop using the `ABSL_LEGACY_THREAD_ANNOTATIONS` compatibility macro before it is going to be removed in the future.

PR #1945, opened 2024-01-05
Fix Build with Bazel 4 and gcc 9

limdor:


limdor:
Seeing there is no interest, I will just close the PR

bufeng-12:
您好！我已收到你的邮件！谢谢！          ——曾君

PR #1968, opened 2025-06-01
Remove unused landmark observations

manhtoanbkhn:
Remove landmark observations that are no longer participate in the optimization problem, reducing the amount of computation, avoiding memory bloat or even crashes.


bufeng-12:
您好！我已收到你的邮件！谢谢！          ——曾君

